PR #8, opened 2018-12-07
Create README.md

rayluo:
What is in this PR?

* The structure, especially the second half of this README is derived from [ADAL Python's README](https://github.com/AzureAD/azure-activedirectory-library-for-python#microsoft-azure-active-directory-authentication-library-adal-for-python). I did change all "adal" wording into "msal", and I changed those generic AAD reference docs from v1 to v2.
* Since the installation steps look simple enough, currently I just put them inline inside the README. (In ADAL, it is a separated page in wiki.) I suspect, as long as the app developers upgrade their `pip` (which we now emphasized in this README), we might not necessarily need those lengthy installation variations at all.
* Again, since we have not touched wiki yet, I basically documented what we went through during the latest sync-up meeting, inside the README. In the future we might just add more and more links to each new separated sample repos.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ Acquiring tokens with MSAL Python need to follow this 3-step pattern.
 1. MSAL proposes a clean separation between
    [public client applications, and confidential client applications](https://tools.ietf.org/html/rfc6749#section-2.1).
    So you will first create either a `PublicClientApplication` or a `ConfidentialClientApplication` instance,
-   and ideally reuse it during the lifecycle of your app. For example:
+   and ideally reuse it during the lifecycle of your app. The following example shows a `PublicClientApplication`:
 
    ```python
    from msal import PublicClientApplication
@@ -56,7 +56,8 @@ Acquiring tokens with MSAL Python need to follow this 3-step pattern.
    It will automatically handle the token refresh for you.
 
    ```python
-   # We now check the cache to see if we have some end users already signed in before.
+   # We now check the cache to see
+   # whether we already have some accounts that the end user already used to sign in before.
    accounts = app.get_accounts()
    if accounts:
        # If so, you could then somehow display these accounts and let end user choose
@@ -71,13 +72,12 @@ Acquiring tokens with MSAL Python need to follow this 3-step pattern.
 
 3. Either there is no suitable token in the cache, or you chose to skip the previous step,
    now it is time to actually send a request to AAD to obtain a token.
-   There are different methods based on your client type. Here we demonstrate the username password flow.
+   There are different methods based on your client type and scenario. Here we demonstrate a placeholder flow.
 
    ```python
    if not result:
        # So no suitable token exists in cache. Let's get a new one from AAD.
-       result = app.acquire_token_by_username_password(
-           "johndoe@contoso.com", "fakepassword", scopes=["user.read"])
+       result = app.acquire_token_by_one_of_the_actual_method(..., scopes=["user.read"])
    if "access_token" in result:
        print(result["access_token"])  # Yay!
    else:
--- a/README.md
+++ b/README.md
@@ -1,12 +1,27 @@
-# Microsoft Authentication Library (MSAL) for Python
+# Microsoft Authentication Library (MSAL) for Python Preview
 
-The MSAL library for Python gives your app the ability to begin using the [Microsoft Cloud](https://cloud.microsoft.com)
-by supporting [Microsoft Azure Active Directory](https://azure.microsoft.com/en-us/services/active-directory/)
-and [Microsoft Accounts](https://account.microsoft.com) in a converged experience using industry standard OAuth2 and OpenID Connect.
+The MSAL library for Python enables your app to access the
+[Microsoft Cloud](https://cloud.microsoft.com)
+by supporting authentication of users with
+[Microsoft Azure Active Directory accounts](https://azure.microsoft.com/en-us/services/active-directory/)
+and [Microsoft Accounts](https://account.microsoft.com) using industry standard OAuth2 and OpenID Connect.
 Soon MSAL Python will also support [Azure AD B2C](https://azure.microsoft.com/services/active-directory-b2c/).
 
 More and more detail about MSAL Python functionality and usage will be documented in the
 [Wiki](https://github.com/AzureAD/microsoft-authentication-library-for-python/wiki).
+
+## Important Note about the MSAL Preview
+
+This library is suitable for use in a production environment.
+We provide the same production level support for this library as we do our current production libraries.
+During the preview we may make changes to the API, internal cache format, and other mechanisms of this library,
+which you will be required to take along with bug fixes or feature improvements.
+This may impact your application.
+For instance, a change to the cache format may impact your users, such as requiring them to sign in again.
+An API change may require you to update your code.
+When we provide the General Availability release
+we will require you to update to the General Availability version within six months,
+as applications written using a preview version of library may no longer work.
 
 ## Installation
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -0,0 +1,116 @@
...
+Here is the latest Q&A on Stack Overflow for MSAL:
+[http://stackoverflow.com/questions/tagged/msal](http://stackoverflow.com/questions/tagged/msal)
+
+## Security Reporting
+
+If you find a security issue with our libraries or services please report it to [secure@microsoft.com](mailto:secure@microsoft.com) with as much detail as possible. Your submission may be eligible for a bounty through the [Microsoft Bounty](http://aka.ms/bugbounty) program. Please do not post security issues to GitHub Issues or any other public site. We will contact you shortly upon receiving the information. We encourage you to get notifications of when security incidents occur by visiting [this page](https://technet.microsoft.com/en-us/security/dd252948) and subscribing to Security Advisory Alerts.
+
+## Contributing
+
+All code is licensed under the MIT license and we triage actively on GitHub. We enthusiastically welcome contributions and feedback. Please read the [contributing guide](./contributing.md) before starting.
```
navyasric: A contributing.md file needs to be added to this repo similar to https://github.com/AzureAD/azure-activedirectory-library-for-python with modifications to match MSAL.

README.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Samples and Documentation
+
+The generic documents on
+[Auth Scenarios](https://docs.microsoft.com/en-us/azure/active-directory/develop/authentication-scenarios)
+and
+[Auth protocols](https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-v2-protocols)
+are recommended reading.
+
+The API reference of MSAL Python is coming soon.
```
navyasric: Can we also add a link to the samples folder where the initial samples are being built for the flows? Somthing like "You can find the samples for the flows supported by MSAL here"

I see you have linked it above but having it under this Samples section would be good

README.md
```diff
@@ -0,0 +1,116 @@
+# Microsoft Authentication Library (MSAL) for Python
+
+The MSAL library for Python gives your app the ability to begin using the [Microsoft Cloud](https://cloud.microsoft.com)
```
navyasric: Can we reword as: "The MSAL library for Python enables your app to access the Microsoft Cloud by supporting authentication of users with Microsoft Azure Active Directory accounts and Microsoft Accounts using industry standard OAuth2 and OpenID Connect."

README.md
```diff
@@ -0,0 +1,116 @@
...
+1. If you haven't already, [install and/or upgrade the pip](https://pip.pypa.io/en/stable/installing/)
+   of your Python environment to a recent version. We tested with pip 18.1.
+2. For now, you can install from our latest dev branch, by `pip install https://github.com/AzureAD/microsoft-authentication-library-for-python.git@dev`
+
+## Usage
+
+Before using MSAL Python (or any MSAL SDKs, for that matter), you will have to
+[register your application with the AAD 2.0 endpoint](https://docs.microsoft.com/en-us/azure/active-directory/develop/quickstart-v2-register-an-app).
+
+Acquiring tokens with MSAL Python is somewhat different than ADAL Python. You will need to follow this 3-step pattern.
```
navyasric: I would not mention ADAL Python and its concepts in Readme since people visiting the MSAL repo may not necessarily be migrating from ADAL( as in there may be new greenfield app developers). I suggest instead we have a section in Wiki comparing or stating differences between ADAL and MSAL for the people who would be migrating.

README.md
```diff
@@ -0,0 +1,116 @@
...
+
+1. Contrary to ADAL (which proposes the notion of `AuthenticationContext`, which is a connection to Azure AD),
```
navyasric: Same comment as above about ADAL.

README.md
```diff
@@ -0,0 +1,116 @@
...
+   and ideally reuse it during the lifecycle of your app. For example:
+
+   ```python
+   from msal import PublicClientApplication
+   app = PublicClientApplication("your_client_id", authority="...")
+   ```
+   
+   Later, each time you would want an access token, you start by:
+   ```python
+   result = None
```
navyasric: Can we also include the acquireToken call here to demonstrate the pattern? Even pseudo code like "# call one of the acquireToken methods to get access token" is fine.

I see that you have explained acquireToken pattern below. Maybe in that case move this line also to the step 2 saying start as follows.

README.md
```diff
@@ -0,0 +1,116 @@
...
+           print(a["username"])
+       # Assuming the end user chose this one
+       chosen = accounts[0]
+       # Now let's try to find a token in cache for this account
+       result = app.acquire_token_silent(config["scope"], account=chosen)
+   ```
+
+3. Either there is no suitable token in the cache, or you chose to skip the previous step,
+   now it is time to actually send a request to AAD to obtain a token.
+   There are different methods based on your client type. Here we demonstrate the username password flow.
```
navyasric: Can we demonstrate auth code or client creds flow? Since we don't want to encourage username password flow, I think it is better to not have it as the Readme example.

README.md
```diff
@@ -0,0 +1,116 @@
...
+   from msal import PublicClientApplication
+   app = PublicClientApplication("your_client_id", authority="...")
+   ```
+   
+   Later, each time you would want an access token, you start by:
+   ```python
+   result = None
+   ```
+
+2. The API model in MSAL provides you explicit control on how to utilize token cache.
```
navyasric: At the end of this step as an optional, can we show or explain how the developer can explicitly control the token cache instead of using MSAL cache if they choose to?

README.md
```diff
@@ -0,0 +1,116 @@
+# Microsoft Authentication Library (MSAL) for Python
+
```
navyasric: As with all our MSAL SDKs, we should add **Preview** to the title and this note about Preview: https://github.com/AzureAD/microsoft-authentication-library-for-js#important-note-about-the-msal-preview.

In addition, we should probably mention that we have added certain flows and others are yet to come.

rayluo: OK, I'm adding that preview keyword into the title and adding that preview paragraph.

For the sake of getting this README published quicker (because a short readme is still better than no readme at all), I would suggest we do not mention flows in this version (besides, neither of MSAL .Net or MSAL javascript mentions any flows in their readme). But I totally agree that we would revisit this in (near?) future to give a concise flows coverage.

README.md
rayluo: I agree this line might seem not very intuitive, but there are 3 steps in total, and this line is just the end of first step, and serves as a starting point of the next 2 steps.
OK I am changing that line into `result = None  # It is just an initial value. Please follow instructions below.`
Will that work?

README.md
rayluo: I think the content in readme should be a minimalist on how to obtain a token. Anything beyond that, such as "How to customize cache", is definitely something needed to be documented - in its own wiki page. :-)

README.md
rayluo: I understand and I did consider this rationale while writing this first draft. The trade-off here is, this (short?) README can only cover that much. So I think we would like to quickly mention an overall usage pattern of MSAL, i.e. the step 1-2-3 thing. In that sense, the Username Password flow is conceptually easiest to NOT get in the way. On the contrary, the auth code flow happens to be too complicated when it incorporates web service framework/logic.

On a side note, I/we do plan to work on an auth code sample, which is the only missing part so far. Later we will have dedicated links to those samples, and we will put auth code sample link here in the top place, and also somehow document that we do not recommend U/P flow, perhaps in its sample too.

henrik-me: Details should be in the wiki not in the readme, imo.

README.md
henrik-me: I believe those details belong in the wiki not in the readme. You can point to a wiki page in the readme which talks about supported flows.

README.md
navyasric: I agree that the details should be in the Wiki and I'm not suggesting covering all the flows here.
The Readme is the first page people see when they look at the repo and I think it should show an acquireToken pattern which is more widely recommended(than the U/P flow) even if a bit more complicated instead of choosing by ease of the pattern. And since we don't yet have auth code flow, I suggest the client credential flow with a client secret which is also simple.

PR #11, opened 2019-01-14
client assertion: fix type error 

yugangw-msft:
On Python 3.6, the jwt encode would return a byte array

### DIFF
--- a/msal/oauth2cli/oauth2.py
+++ b/msal/oauth2cli/oauth2.py
@@ -70,7 +70,7 @@ class BaseClient(object):
             if client_assertion_type is None:  # RFC7521 defines only 2 profiles
                 TYPE_JWT = "urn:ietf:params:oauth:client-assertion-type:jwt-bearer"
                 TYPE_SAML2 = "urn:ietf:params:oauth:client-assertion-type:saml2-bearer"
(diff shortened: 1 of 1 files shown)


msal/oauth2cli/oauth2.py
```diff
@@ -70,7 +70,7 @@ def __init__(
             if client_assertion_type is None:  # RFC7521 defines only 2 profiles
                 TYPE_JWT = "urn:ietf:params:oauth:client-assertion-type:jwt-bearer"
                 TYPE_SAML2 = "urn:ietf:params:oauth:client-assertion-type:saml2-bearer"
-                client_assertion_type = TYPE_JWT if "." in client_assertion else TYPE_SAML2
+                client_assertion_type = TYPE_JWT if "." in client_assertion.decode() else TYPE_SAML2
```
rayluo: @yugangw-msft Thanks for pointing this out! Now I'm actually thinking of do it in the other way around: `... if b"." in client_assertion else ...`. If you don't mind, you also change it this way and give it a quick run in your local environment, and tell me the result.

Meanwhile I will probably organize such fix in a separate PR, because I'll have other places need such fix, as well as relevant docstrings etc.. And it is not fair to ask you to do all that in YOUR pr. :-)

Issue #15, opened 2019-01-15
trace: `verify` flag is not wired up to turn off cert validation for capturing traces


yugangw-msft:
This is important to capture traces through proxy tools like fiddler which uses self-signed certificate.
The [verify](https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/dev/msal/application.py#L81) appears designed for such purpose, but it is not wired up in oauth2.py, nor is wired up in other places like `authority.py`

rayluo:
OK I'll take a look. By the way, you are using similar steps like [this](https://github.com/AzureAD/azure-activedirectory-library-for-python/wiki/Client-credentials) to generate your self-signed cert, aren't you?

yugangw-msft:
do this, which might be simpler
  > install azure-cli
  > run `az ad sp create-for-rbac -n http://you-great-sp --create-cert --skip-assignment`

PR #17, opened 2019-01-25
Full support for verify and proxies

rayluo:
This PR resolves issue #15.

How to test?

* Use MSAL Python 0.1.0. Start fiddler. Add a `verify=False` flag in any of the samples. Run it. You won't be able to complete a sample run due to the library complaining (this is essentially a Man-in-the-middle attack).

* You can test this PR by:

      pip install git+https://github.com/AzureAD/microsoft-authentication-library-for-python.git@full-support-for-verify-and-proxies

* Rerun the previous test. This time you will successfully obtain a token (despite some warning message emitted from underlying library).

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -97,7 +97,7 @@ class ClientApplication(object):
         self.timeout = timeout
         self.authority = Authority(
                 authority or "https://login.microsoftonline.com/common/",
-                validate_authority, verify=self.verify, proxies=self.proxies)
+                validate_authority, verify=verify, proxies=proxies, timeout=timeout)
             # Here the self.authority is not the same type as authority in input
         self.token_cache = token_cache or TokenCache()
         self.client = self._build_client(client_credential, self.authority)
@@ -167,7 +167,8 @@ class ClientApplication(object):
             sending them on the wire.)
         """
         the_authority = Authority(
-            authority, verify=self.verify, proxies=self.proxies,
+            authority,
+            verify=self.verify, proxies=self.proxies, timeout=self.timeout,
             ) if authority else self.authority
         client = Client(
             {"authorization_endpoint": the_authority.authorization_endpoint},
@@ -275,7 +276,8 @@ class ClientApplication(object):
         """
         assert isinstance(scopes, list), "Invalid parameter type"
         the_authority = Authority(
-            authority, verify=self.verify, proxies=self.proxies,
+            authority,
+            verify=self.verify, proxies=self.proxies, timeout=self.timeout,
             ) if authority else self.authority
 
         if not force_refresh:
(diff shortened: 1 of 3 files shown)


msal/application.py
```diff
@@ -383,17 +387,20 @@ def acquire_token_by_username_password(
...
     def _acquire_token_by_username_password_federated(
             self, user_realm_result, username, password, scopes=None, **kwargs):
+        verify = kwargs.pop("verify", self.verify)
+        proxies = kwargs.pop("proxies", self.proxies)
         wstrust_endpoint = {}
         if user_realm_result.get("federation_metadata_url"):
             wstrust_endpoint = mex_send_request(
-                user_realm_result["federation_metadata_url"])
+                user_realm_result["federation_metadata_url"],
+                verify=self.verify, proxies=self.proxies)
```
yugangw-msft: I believe the right one should be `verify=verify, proxies=proxies`

msal/application.py
```diff
@@ -383,17 +387,20 @@ def acquire_token_by_username_password(
 
     def _acquire_token_by_username_password_federated(
             self, user_realm_result, username, password, scopes=None, **kwargs):
+        verify = kwargs.pop("verify", self.verify)
+        proxies = kwargs.pop("proxies", self.proxies)
```
yugangw-msft: Guess we must have reasons for taking related settings through arguments, but I do feel it is a bit redundant. So wonder why?

rayluo: Excellent question! The use of `**kwargs` a.k.a. Keyworded Variables is not a requirement in terms of authentication api design, that's why they are [purposely undocumented](https://msal-python.readthedocs.io/en/refine-docs/). The behavior remains here, only as an influence by the underlying requests, which can accept [verify](http://docs.python-requests.org/en/master/user/advanced/#ssl-cert-verification), [proxies](http://docs.python-requests.org/en/master/user/advanced/#proxies), and some more [other parameters](http://docs.python-requests.org/en/master/user/quickstart/#timeouts), all through ad-hoc arguments. And we get these [syntactic sugar](https://en.wikipedia.org/wiki/Syntactic_sugar), almost for free, by wiring them properly.

All that being said, I would personally have no objection if many of you prefer to remove all those `**kwargs` from MSAL Python APIs. Just let us know.

yugangw-msft: I suggest don't remove **kwargs, and also don''t extract `verify` and `proxies` out of the kwargs. My main point is we should only have one place to set these 2 fields which is through the constructor

rayluo: Understood. But:

1. I think our current approach, i.e. only documenting `verify` and `proxies` in constructor, is conceptually close enough to "we only have one place to set these 2 fields which is through the constructor".
2. As long as we keep the **kwargs in this method signature, it is possible that the caller still provides `verify` and `proxies`. Then what shall we do in this undocumented scenario? The current implementation tries to provide an arguably intuitive behavior matching the caller's intention. Rather than, accepting the `verify` and `proxies` yet silently ignore them, even though it would technically be the "correct" behavior for any undocumented/unimplemented parameter.
3. If, however, we really don't like the undocumented behavior in the previous paragraph, we could choose go one step further by removing the **kwargs in this method signature. That way it would become impossible to provide `verify` and `proxies` through this method.

yugangw-msft: Okay, got it. For context, I was a bit confused by why we only inspect the kwargs here, while `acquire_token_silent` above doesn't have the inspections

rayluo: Yes, that seemingly discrepancy was necessary, only for this method, because the `mex_send_request(...)` and `wstrust_send_request(...)` happen to require an explicit `verify` and `proxies`. All in all, personally I think the current approach is good enough.
Guess we can conclude this conversation thread, then?

PR #19, opened 2019-02-15
Adjusting token cache format

rayluo:
* Change scope's internal representation from a list to a string
* Remove client_info from RefreshToken because it was optional and now obsolete
* Add test case to ensure the token cache representation complies with the Unified Schema
* The existing token cache implementation is already extendable, we only need to add a test case to assert that.

You can install this branch for your testing:

    pip install git+https://github.com/AzureAD/microsoft-authentication-library-for-python.git@target-in-string

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -292,12 +292,14 @@ class ClientApplication(object):
                     })
             now = time.time()
             for entry in matches:
-                if entry["expires_on"] - now < 5*60:
+                expires_in = int(entry["expires_on"]) - now
+                if expires_in < 5*60:
                     continue  # Removal is not necessary, it will be overwritten
+                logger.debug("Cache hit an AT")
                 return {  # Mimic a real response
                     "access_token": entry["secret"],
                     "token_type": "Bearer",
-                    "expires_in": entry["expires_on"] - now,
+                    "expires_in": int(expires_in),  # OAuth2 specs defines it as int
                     }
 
         matches = self.token_cache.find(
@@ -311,6 +313,7 @@ class ClientApplication(object):
                 })
         client = self._build_client(self.client_credential, the_authority)
         for entry in matches:
+            logger.debug("Cache hit an RT")
             response = client.obtain_token_by_refresh_token(
                 entry, rt_getter=lambda token_item: token_item["secret"],
                 scope=decorate_scope(scopes, self.client_id))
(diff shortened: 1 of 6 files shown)


msal/application.py
```diff
@@ -292,12 +292,14 @@ def acquire_token_silent(
...
-                if entry["expires_on"] - now < 5*60:
+                expires_in = int(entry["expires_on"]) - now
+                if expires_in < 5*60:
                     continue  # Removal is not necessary, it will be overwritten
+                logger.debug("Cache hit an AT")
                 return {  # Mimic a real response
                     "access_token": entry["secret"],
                     "token_type": "Bearer",
-                    "expires_in": entry["expires_on"] - now,
+                    "expires_in": int(expires_in),  # OAuth2 specs defines it as int
```
henrik-me: >int [](start = 34, length = 3)

it seems like expires_in is already an int? Did -now change it to something else? Should that then have been converted to an int?

henrik-me:
>             logger.debug(

should this be info or error instead of debug? (not sure how the python logger works though)

---
Refers to: msal/application.py:322 in 9110eca. [](commit_id = 9110eca2ce0dc850437b19c8acddac98364a2c9c, deletion_comment = False)

msal/token_cache.py
```diff
@@ -38,13 +38,19 @@ def __init__(self):
     def find(self, credential_type, target=None, query=None):
         target = target or []
         assert isinstance(target, list), "Invalid parameter type"
+        target_set = set(target)
```
henrik-me: is target scopes? if yes, do you have a place where you remove dupes?

msal/application.py
rayluo: Good question! It is true that the `expires_in` in the original server-side response is an integer. But here in this code path we hit a token in the cache, and then re-calculate a new `expires_in` based on its original value and the [current time(), which happens to be a float in Python](https://docs.python.org/3/library/time.html#time.time). So we do a Forceful Unobvious Conversion to Keep the Integer Time, at the exact place where it is needed.

msal/token_cache.py
rayluo: Yes `target` is exactly `scope` in Unified Schema's dialect.

The `set(...)` method creates a native set object in Python. It automatically deduplicates the scopes.

PR #25, opened 2019-03-25
Reading Authority Aliases

abhidnya13:
This PR covers the first two parts of Authority Migration mentioned as [here](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=%2FAuthorityMigration%2Fsummary.md&version=GBauthority-migration).

UPDATE: If you prefer, you can install this branch for your smoke testing:

    pip install git+https://github.com/AzureAD/microsoft-authentication-library-for-python.git@authority_migration


msal/application.py
```diff
@@ -240,8 +256,13 @@ def get_accounts(self, username=None):
         # but does NOT correlate them with saved RTs. It probably won't matter,
         # because in MSAL universe, there are always Accounts and RTs together.
         accounts = self.token_cache.find(
-            self.token_cache.CredentialType.ACCOUNT,
-            query={"environment": self.authority.instance})
+            self.token_cache.CredentialType.ACCOUNT, query={"environment": self.authority.instance})
```
rayluo: Nit pick: As controversial as it might be, Python coding style does recommend ["limit all lines to a maximum of 79 characters."](https://www.python.org/dev/peps/pep-0008/#maximum-line-length). Although in our code base we don't strictly enforce such rule and you can occasionally see 80+ characters per line, we don't usually combine existing short lines into one extra long line in a new PR. Please refer to the "reduce noise" topic below. :-)

tests/test_application.py
```diff
@@ -239,3 +239,44 @@ def tester(url, data=None, **kwargs):
...
+                          'realm': 'common',
+                          'local_account_id': None,
+                          'username': None,
+                          'authority_type': 'MSSTS'}, accounts[0])
+
+    def test_acquire_token_silent(self):
+        app = PublicClientApplication(
+            "my_client_id", authority="https://login.microsoftonline.com/common",
+            token_cache=self.cache)
+        a = app.acquire_token_silent(["s1", "s2"], self.account)
```
rayluo: Ask me a funny story about this variable naming. :)

abhidnya13: Hey Ray, is there a reason why its different than the one we add in token_cache? The acquire token silent will fail if its a different client_id causing the test case to fail ?

rayluo: Nope, client_id should remain the same here.

In fact, in some earlier version there were some hardcoded client_id here and there, thus vulnerable to fragile test behaviors due to accidentally changing some of the client_id. So in the current version, we define one single copy of truth as `self.client_id` and reuse them everywhere. That's the whole point of defining some test settings as class-wide members. (On the other hand, the current version also demote some previous `self.xyz` settings to local variables if they are only used inside `setUp()`.)

PR #26, opened 2019-03-27
Cache lookup change

abhidnya13:
- This PR corrects the cache look up logic which was case sensitive for 'target' field in the cache. 
- It ensures that in scopes comparison when finding token cache User.Read and user.read is not differentiated.
- This change was found when testing #25 as scopes comparison would not return an Access Token and would then use Refresh token to acquire a new Access token.

### DIFF
--- a/msal/token_cache.py
+++ b/msal/token_cache.py
@@ -37,18 +37,20 @@ class TokenCache(object):
 
     def find(self, credential_type, target=None, query=None):
         target = target or []
+        target = [item.lower() for item in target]
         assert isinstance(target, list), "Invalid parameter type"
         target_set = set(target)
         with self._lock:
             # Since the target inside token cache key is (per schema) unsorted,
             # there is no point to attempt an O(1) key-value search here.
             # So we always do an O(n) in-memory search.
-            return [entry
-                for entry in self._cache.get(credential_type, {}).values()
-                if is_subdict_of(query or {}, entry)
-                and (target_set <= set(entry.get("target", "").split())
-		    if target else True)
-                ]
+            cache_entries = []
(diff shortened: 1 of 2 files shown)


msal/token_cache.py
```diff
@@ -37,18 +37,20 @@ def __init__(self):
...
-                for entry in self._cache.get(credential_type, {}).values()
-                if is_subdict_of(query or {}, entry)
-                and (target_set <= set(entry.get("target", "").split())
-		    if target else True)
-                ]
+            cache_entries = []
+            for entry in self._cache.get(credential_type, {}).values():
+                target_cmp = entry.get("target", "").lower()
+                if is_subdict_of(query or {}, entry) and (target_set <= set(target_cmp.split())
+                if target else True):
```
henrik-me: >if target else True): [](start = 16, length = 21)

why is this needed? Will it ever evaluate to false?

abhidnya13: This is the part of second expression after the and operator. So we evaluate the expression only if target is a non empty list. If target is an empty list we want it to be True and skip evaluating the expression. Target can be an empty list in case of finding a refresh token. So if the first part is True in case of Refresh Token , the cache entry is valid and should be appended.

henrik-me: hmm, looks strange to me. will add this to my training list

---
In reply to: [269679161](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/26#discussion_r269679161) [](ancestors = 269679161)

rayluo:
Scope in OAuth2 is defined as case-sensitive. Better not relax this check from our implementation.

PR #27, opened 2019-03-29
Family of Client Id (FoCI) support

rayluo:
This PR contains:

* Refactor the cache test code, to provide `build_response(...)` helpers. They are helpful for the rest of this PR, and probably for all the recent and future PRs in unit testing.
* The main FoCI behaviors are implemented in 2 files:
  - application.py, [its main logic](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/27/files#diff-f7ae931bcdab4b14f3a256baf7d6922aR312) closely represents the upper half of [this internal diagram](https://identitydivision.visualstudio.com/77f0161c-4567-4cec-a336-29ca6d5459d9/_apis/git/repositories/dad996e9-1c8b-4403-8efd-462caaa962e1/Items?path=%2FFamilyOfClientId%2FFociClient.png&versionDescriptor%5BversionOptions%5D=0&versionDescriptor%5BversionType%5D=2&versionDescriptor%5Bversion%5D=60ec576c6bd64aa838d4462de7d8afa1a3715b6b&download=false&resolveLfs=true&%24format=octetStream&api-version=5.0-preview.1). (We do not need to explicitly implement the bottom half, because those behaviors already exist in our code base.) Let me know if you have any question.
  - test_application.py, which is basically equivalent to [this test file in MSAL .Net FOCI PR](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/957/files#diff-3ffdf9723b84f964b3c6ca20e10669aa)
* As a pre-existing behavior, the get_accounts() API does not filter account by client_id. This PR does not change that.
* HIstorically we do not yet have a remove_account() API (there is actually an open github issue/feature request for that). We will tackle that task at a later time, after some internal discussion/clarification on that topic settles down. For now, this PR contains no remove_account().


henrik-me:
remove_account() : You can still add this using the behavior known for now. Some details may change later but that is ok. tracking it as a separate item is ok, but it should be part of the API even if the internal behavior may change. This is still in preview.

msal/application.py
```diff
@@ -305,26 +305,71 @@ def acquire_token_silent(
                     "token_type": "Bearer",
                     "expires_in": int(expires_in),  # OAuth2 specs defines it as int
                     }
+        return self._acquire_token_silent_by_finding_rt_belongs_to_me_or_my_family(
+                the_authority, decorate_scope(scopes, self.client_id), account,
+                **kwargs)
 
+    def _acquire_token_silent_by_finding_rt_belongs_to_me_or_my_family(
+            self, authority, scopes, account, **kwargs):
+        query = {
```
abhidnya13: Should we have to add authority migration to look for aliases in this query as well ?

rayluo: In our current plan, the authority aliases would work at one layer higher, i.e. in the `acquire_token_silent(...)` level.

We will see how it works and continue this conversation at your PR there. Meanwhile, thanks for your review. I'm going to merge this now.

PR #30, opened 2019-04-03
Add an early link to other ADAL or MSAL SDKs

rayluo:
Add an early link to the recently-revised [other ADAL or MSAL SDKs page](https://github.com/AzureAD/microsoft-authentication-library-for-python/wiki/Microsoft-Authentication-Client-Libraries), as inspired by the same thing [Kusto Python library did](https://github.com/Azure/azure-kusto-python/blob/master/README.md#looking-for-sdks-for-other-languagesplatforms). (They are another first party internal customer who takes dependency of ADAL Python, by the way.)

Also reorganize the sample links a little bit.

Consider this PR is part of the April effort of refining our docs before the Build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,8 +7,8 @@ by supporting authentication of users with
 and [Microsoft Accounts](https://account.microsoft.com) using industry standard OAuth2 and OpenID Connect.
 Soon MSAL Python will also support [Azure AD B2C](https://azure.microsoft.com/services/active-directory-b2c/).
 
-More and more detail about MSAL Python functionality and usage will be documented in the
-[Wiki](https://github.com/AzureAD/microsoft-authentication-library-for-python/wiki).
+Not sure whether this is the SDK you are looking for? There are other Microsoft Identity SDKs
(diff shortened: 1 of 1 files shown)


rayluo:
Good question @navyasric !

That "[separated page](https://github.com/AzureAD/microsoft-authentication-library-for-python/wiki/Microsoft-Authentication-Client-Libraries)" was acturally [created months ago](https://github.com/AzureAD/microsoft-authentication-library-for-python/wiki/Microsoft-Authentication-Client-Libraries/_history), initially as an effort to answer an FAQ in email. The platform information (in particular, Linux) in that page is not found elsewhere, at least not organized in such a language-and-platform matrix.

At that time I did suggest, and others agreed, to eventually also host that page into some more generic document websites. But before that happens, it doesn't harm to host it here and we benefit from it. :-)

jmprieur:
I Agree with Navya that we should host it in docs.ms.

rayluo:
Like I said earlier, I'm also more than happy to see that matrix show up in docs.ms. And when that eventually happens, all we need to do is to update the link here. Or do you mean we shall NOT even mention that page in our readme here?

PR #35, opened 2019-04-11
Adding client assertion type

abhidnya13:
- Client assertion type was not set in configuration of client for certificate credential flow. 
- This PR adds the client assertion type.


rayluo:
Hmm, is this change necessary to make it work? I'll take a close look tomorrow.

PR #37, opened 2019-04-13
Make msal namespace extensible

marstr:
Developing [marstr/microsoft-authentication-extensions-for-python](https://github.com/marstr/microsoft-authentication-extensions-for-python), I came to realize that I was unable to extend the `msal` namespace. The result was a weird situation where the extensions library would need to be installed first. This change will allow these two distributions to be installed side-by-side in the same way the .NET packages can be.

### DIFF
--- a/msal/__init__.py
+++ b/msal/__init__.py
@@ -24,6 +24,7 @@
 # THE SOFTWARE.
 #
 #------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


marstr:
Chatted offline with @rayluo, we agreed that this is really more of an aesthetic thing. For now, we'll just rename the work I've been doing to `msal_extensions`.

PR #40, opened 2019-04-23
Authorization code flow sample

abhidnya13:
- Python Flask Web Application acquiring token using authorization code flow


abhidnya13:
Hey Ray,

1. You are right about acquire token silent not working for the Web App. I have made that change and as you mentioned we can revisit storing it in the session storage or some way to make sure they don't have to authenticate every other hour.

2. The code is already structured very similar to the repo you mentioned because I referred it when I started developing [this](https://github.com/Azure-Samples/active-directory-python-webapp-graphapi) for our ADAL sample that uses Microsoft Graph. I think we should still have this sample here and not in the repo you mentioned mainly because that is a particular use case of using Graph. We would want to show the process of acquiring a token as the base case here for any resource just like how all our other samples do in this repo. For this,  we can add a task to create a sample in Azure Samples like we did for ADAL that demonstrates how we can display information specifically from Microsoft Graph.

sample/authorization-code-flow-sample/authorization_code_flow_sample.py
```diff
@@ -0,0 +1,80 @@
...
+
+    python sample.py parameters.json
+"""
+
+import sys  # For simplicity, we'll read config file from 1st CLI param sys.argv[1]
+import json
+import logging
+import uuid
+
+import flask as flask
```
rayluo: Just do `import flask` would be enough, since we are not renaming the imported module anyway. :)

sample/authorization-code-flow-sample/authorization_code_flow_sample.py
```diff
@@ -0,0 +1,80 @@
...
+        client_credential=config["client_secret"],
+        # token_cache=...  # Default cache is in memory only.
+        # You can learn how to use SerializableTokenCache from
+        # https://msal-python.rtfd.io/en/latest/#msal.SerializableTokenCache
+    )
+
+
+@app.route("/")
+def main():
+    login_url = 'http://localhost:5000/login'
```
rayluo: We can use `/login` here, so that we do not need to hard-code the port number here. Otherwise app dev won't immediately know they need to change this line, should they choose to run/deploy this sample on a different port.

BTW, since we are going to change this line anyway, we can probably just do one inline assignment at line 50 below. But this suggestion is totally optional. :-)

sample/authorization-code-flow-sample/authorization_code_flow_sample.py
```diff
@@ -0,0 +1,80 @@
...
+    auth_state = str(uuid.uuid4())
+    flask.session['state'] = auth_state
+    authorization_url = application.get_authorization_request_url(config['scope'], state=auth_state,
+                                                                  redirect_uri=config['redirect_uri'])
+    resp = flask.Response(status=307)
+    resp.headers['location'] = authorization_url
+    return resp
+
+
+# Our configured redirect uri is http://localhost:8000/getAToken.
```
rayluo: Although it is just a commented line, the port number inside does not match elsewhere in this sample. We can probably just remove this line, since it is kind of duplicate of line 9 anyway.

sample/authorization-code-flow-sample/authorization_code_flow_sample.py
henrik-me: should this be defined in the config file? or wouldn't the web service itself have this defined somewhere. (is that what /login means?)

---
In reply to: [277909842](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/40#discussion_r277909842) [](ancestors = 277909842)

henrik-me:
Does the wiki reference the list of samples including notes on how to run those and what they achieve? Perhaps the code should also reference that wiki page?

sample/authorization-code-flow-sample/authorization_code_flow_sample.py
rayluo: The `/login` literally means to use the exact string `/login` here (or directly assigned to [line 50](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/40/files#diff-3a67f168fce508486b1b71721472b542R50)). Those host and port can be omitted because the [HTTP Location header supports using relative url](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Location#Directives), which we want here, so that we do not need to hardcode a configurable port number.

On the other hand, the path `/login` will not be in config file because it is not something can be or need to be configured. It needs to match [this hard-coded line](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/40/files#diff-3a67f168fce508486b1b71721472b542R54), so it is a self-contained business logic of this sample app.

rayluo:
Q: Does the wiki reference the list of samples including notes on how to run those and what they achieve? Perhaps the code should also reference that wiki page?

A: Historically we do not have a wiki reference the list of samples. We have only one place referencing the list of built-in samples, and that is buried deep at THE END OF [this section inside the README](https://github.com/AzureAD/microsoft-authentication-library-for-python#usage-and-samples), because we were assuming new comers would start from that "Usage and Samples" section in README, and then naturally discover and follow that link. Now on a second thought, we should probably also have a wiki page (and [now we do](https://github.com/AzureAD/microsoft-authentication-library-for-python/wiki/Samples)!) referencing to the list of samples, just to capture audiences coming from a different "entrance".

But then I understand your comment was more about "how to configure/run such specific sample". That kind of information should be inside our self-contained samples. Recently we've been [adding more descriptions into our existing samples, right at each configuration setting where a new dev would ask questions](https://github.com/AzureAD/microsoft-authentication-library-for-python/blame/dev/sample/confidential_client_secret_sample.py#L8-L13). That approach would also address what David mentioned yesterday that new app devs would tend to ask "client secret? what is that? where do I get one?", that sorts of questions. In that sense, we may want to add [this url](https://docs.microsoft.com/en-us/azure/active-directory/develop/quickstart-configure-app-access-web-apis) into our "redirect_url" setting.

PR #41, opened 2019-05-10
Add remove_account() API

### LINKED ISSUE #12: token cache: support logout
The goal is to remove all tokens belong to a user name. In ADAL, with flat array, it is easy to do it through filtering by user_id; but now with the new universal token cache scheme, it is hard for client applications to do it.

rayluo:
Add remove_account() API, and it will signout the account from the app family. This PR will close #12 .

Most of the implementations are refactoring the underlying cache helpers. To review high level "what does it do" logic, please focus on only 50 lines starting from [here](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/41/files#diff-f7ae931bcdab4b14f3a256baf7d6922aR283) (half of them are inline documentation).

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -280,22 +280,47 @@ class ClientApplication(object):
                 return [alias for alias in group if alias != instance]
         return []
 
-    def remove_account(self, home_account):
-        """Remove all relevant RTs and ATs from token cache"""
-        owned_by_account = {
+    def remove_account(self, account):
+        """Sign me out and forget me from token cache"""
+        self._forget_me(account)
+
+    def _sign_out(self, home_account):
+        # Remove all relevant RTs and ATs from token cache
+        owned_by_home_account = {
             "environment": home_account["environment"],
             "home_account_id": home_account["home_account_id"],}  # realm-independent
-        for rt in self.token_cache.find(  # Remove RTs, and RTs are realm-independent
-                TokenCache.CredentialType.REFRESH_TOKEN, query=owned_by_account):
+        app_metadata = self._get_app_metadata(home_account["environment"])
+        # Remove RTs/FRTs, and they are realm-independent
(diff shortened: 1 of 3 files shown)


rayluo:
Thanks @SomkaPe for reviewing this PR promptly. I'm also documenting the summary of the code review conversations below for future reference.

> [5/10 3:19 PM] Petro Somka
>     is there any api review pr for it?
>     i mean is there some description of logic
> ​[5/10 3:22 PM] Ray Luo
>     API review PR for the remove_account() behavior is the internal PR 775 I sent out weeks ago, about the sign-out-of-app or sign-out-of-family. 
> Now this actual implementation do the sign out of family.
> ​[5/10 3:22 PM] Petro Somka
>     ok

> ​[5/10 3:28 PM] Petro Somka
>     why do you not remove idTokens in _sign_out() ?
> ​[5/10 3:32 PM] Ray Luo
>     I believe id_token are intrinsically more close to accounts, so I only do it in _forget_me().
> ​[5/10 3:33 PM] Petro Somka
>     _forget_me() and _sign_out() are private , and only removeAccount() is public which just call _forget_me()
> [5/10 3:34 PM] Ray Luo
>     Yes
>     In that sense, whether removing IDT happens in signout() or in forget_me(), does not really matter right now.
>    It would matter in the future if our entire org agree to officially provide public api for signout() and forget_me(). At that time we would need to precisely define and agree on their behaviors.

> 

> ​[5/10 3:31 PM] Petro Somka
>     one more question if you remove account realm - independently It also means that get Accounts api returns accounts realm independently as well , correct ?
> ​[5/10 3:33 PM] Ray Luo
>     Yes, I believe get_account should return all accounts regardless of their realms.
> [5/10 3:35  PM] Petro Somka
> but not returning same account twice if there are few account in the cache with diff realms 
> (1 liked)

PR #43, opened 2019-05-17
Rename vague has_state_changed to has_memory_state_changed (On Hold)

rayluo:
It caused some confusion when team members implementing token cache file state change feature in a subclass.

### DIFF
--- a/msal/token_cache.py
+++ b/msal/token_cache.py
@@ -259,7 +259,7 @@ class SerializableTokenCache(TokenCache):
     """This serialization can be a starting point to implement your own persistence.
 
     This class does NOT actually persist the cache on disk/db/etc..
-    Depending on your need,
+    If your app would not share token cache file with other apps during runtime,
     the following simple recipe for file-based persistence may be sufficient::
 
         import os, atexit, msal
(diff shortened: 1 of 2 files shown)


rayluo:
@marstr Thanks for your thoughtful input, and your sharing of that "Strong opinion, weakly held" approach!

I think your suggestion makes sense. On the other hand, for the sake of completeness, I can also explain why the current `has_memory_state_changed` flag was invented and why it did not reach its full potential.

The `has_memory_state_changed` was indeed designed for a very simplified scenario, which is to persist token cache in one way i.e. from-memory-to-disk, flush at the time of app termination, for one single app who does NOT share cache with other apps. As naive as it might seem, it is arguably the most common scenario, say, a 3rd-party script (rather than Azure CLI), being run manually or by crontab, on a trusted machine (either on end user's desktop or on a server under their control), so [the corresponding implementation](https://github.com/AzureAD/microsoft-authentication-library-for-python/compare/file-based-token-cache?expand=1) did not even have encryption.

That experimental implementation was not merged, due to potential security review rejection on its lacking of encryption. In the real world people are rolling their own similar implementation anyway, and they were even sending their implementation to us in hoping that we would adopt it. Currently we provide no out-of-box token persistence, but provide a [how-to recipe (which uses that `has_state_changed` flag)](https://msal-python.readthedocs.io/en/latest/#msal.SerializableTokenCache). Such a middle ground was never ideal in the first place, because it splits one otherwise integral behavior i.e. memory-to-disk-persistence into two incomplete pieces: (1) a vague `has_memory_state_changed` flag being exposed but would otherwise be an internal member, and (2) an workable (albeit naive) file cache implementation that has no plan to be shipped.

So, that is the sad life of `has_memory_state_changed`.

Lastly, the logic @marstr you described is for the other way around, from-disk-to-memory, which is suitable to the extension module you are currently working on. As long as we'll ship a feature-completed `TokenCache` sub-class, there shouldn't have a need to expose its internal flag. So, you are good.

We will probably keep this PR on-hold for now, and revisit it later when your extension is completed.

rayluo:
We don't see to have a strong need for this change right now. Closing.

PR #48, opened 2019-05-23
Updating authority URL regexp to have escaped backslash

### LINKED ISSUE #47: Invalid Escape Character in Authority Regex
If you take a look at the following regex:
https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/85bdb5e72443f5ab5ac223b443b8bda2c6a53387/msal/authority.py#L68

You'll see that it includes `\?` in one of the groups. I believe the intention was to disallow both backslash characters and question marks, but the result is that it is being interpreted as an invalid escape code. This is causing `pytest` to throw a warning in the downstream project [AzureAD/microsoft-authentication-extensions-for-python](https://github.com/AzureAD/microsoft-authentication-extensions-for-python).

marstr:
Fixes #47

### DIFF
--- a/msal/authority.py
+++ b/msal/authority.py
@@ -65,7 +65,7 @@ class Authority(object):
 
 def canonicalize(url):
     # Returns (canonicalized_url, netloc, tenant). Raises ValueError on errors.
-    match_object = re.match(r'https://([^/]+)/([^/?#\\]+)', url.lower())
+    match_object = re.match(r'https://([^/]+)/([^/?#]+)', url.lower())
     if not match_object:
         raise ValueError(
             "Your given address (%s) should consist of "
--- a/msal/authority.py
+++ b/msal/authority.py
@@ -65,7 +65,7 @@ class Authority(object):
 
 def canonicalize(url):
     # Returns (canonicalized_url, netloc, tenant). Raises ValueError on errors.
-    match_object = re.match("https://([^/]+)/([^/\?#]+)", url.lower())
+    match_object = re.match(r'https://([^/]+)/([^/?#\\]+)', url.lower())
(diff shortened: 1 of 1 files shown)


msal/authority.py
```diff
@@ -65,7 +65,7 @@ def user_realm_discovery(self, username):
 
 def canonicalize(url):
     # Returns (canonicalized_url, netloc, tenant). Raises ValueError on errors.
-    match_object = re.match("https://([^/]+)/([^/\?#]+)", url.lower())
+    match_object = re.match(r'https://([^/]+)/([^/?#\\]+)', url.lower())
```
rayluo: Thanks very much for the catch! In such case, I think we can drop the `\\` part because it was not the intention.

By the way, can you share a link to that pytest warning message? Would like to see it in action. :-)

marstr: No problem! Here's the message I got:

```
========================================================================================================================================================================= warnings summary =========================================================================================================================================================================
/home/marstr/.local/share/virtualenvs/microsoft-authentication-extensions-for-python/lib/python3.6/site-packages/msal/authority.py:68
  /home/marstr/.local/share/virtualenvs/microsoft-authentication-extensions-for-python/lib/python3.6/site-packages/msal/authority.py:68: DeprecationWarning: invalid escape sequence \?
    match_object = re.match("https://([^/]+)/([^/\?#]+)", url.lower())

-- Docs: https://docs.pytest.org/en/latest/warnings.html
========================================================================================================================================================= 1 passed, 1 skipped, 1 warnings in 0.52 seconds ==========================================================================================================================================================
```

marstr: Fixed in ce450a7

PR #61, opened 2019-06-17
Pass through verify flag and fix scopes default value

roederja2:
* Pass through verify flag
* change default value from None to [] as None leads to a crash further downstream

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -271,7 +271,7 @@ class ClientApplication(object):
         if not self.authority_groups:
             resp = requests.get(
                 "https://login.microsoftonline.com/common/discovery/instance?api-version=1.1&authorization_endpoint=https://login.microsoftonline.com/common/oauth2/authorize",
-                headers={'Accept': 'application/json'})
+                headers={'Accept': 'application/json'}, verify=self.verify, proxies=self.proxies, timeout=self.timeout)
             resp.raise_for_status()
             self.authority_groups = [
                 set(group['aliases']) for group in resp.json()['metadata']]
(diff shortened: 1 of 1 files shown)


msal/application.py
```diff
@@ -271,7 +271,7 @@ def _get_authority_aliases(self, instance):
         if not self.authority_groups:
             resp = requests.get(
                 "https://login.microsoftonline.com/common/discovery/instance?api-version=1.1&authorization_endpoint=https://login.microsoftonline.com/common/oauth2/authorize",
-                headers={'Accept': 'application/json'})
+                headers={'Accept': 'application/json'}, verify=self.verify)
```
rayluo: Thanks for the catch! And, since we are modifying this anyway, I would suggest to also wire up the `proxies` and `timeout`.

                headers={'Accept': 'application/json'},
                verify=self.verify, proxies=self.proxies, timeout=self.timeout)

msal/application.py
```diff
@@ -511,7 +511,7 @@ def acquire_token_by_device_flow(self, flow, **kwargs):
                 **kwargs)
 
     def acquire_token_by_username_password(
-            self, username, password, scopes=None, **kwargs):
+            self, username, password, scopes=[], **kwargs):
```
rayluo: I understand your point. However this fix is [not ideal](https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html). I would suggest to rollback this change in current PR, so that it would only fix that `verify` (and `proxies` and `timeout` issue) and get merged soon, and we will come up with a separated PR to address that scope thing.

roederja2: Yes I meant further down the stack inside MSAL. My application is supplying a scope I only noticed the crash when I was experimenting. As a rule I think libraries should not crash if valid (according to the interface) arguments are supplied.

But agreed I'll remove that fix and put in timeout and proxies as well in addition to verify.

rayluo: The "default value" topic is addressed in PR #64. Closing this conversation.

PR #68, opened 2019-06-19
ConfidentialClient with extra claims

### LINKED ISSUE #67: Confidential client provides a way to accept extra claims
It could be achieved by accepting new optional parameter in the `ConfidentialClientApplication(..., client_claims={...}, ...)`.

Should the need comes up, we can also explore the api to accept a previously-signed assertion.

rayluo:
This ~is a proof-of-concept in attempt to~ resolve #67.

~It has NOT been tested at this time.~ A unit test case is added.

Please also help proof-read [the documentation](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/68/files#diff-f7ae931bcdab4b14f3a256baf7d6922aR75), which will automatically be published into our [online reference doc](https://msal-python.readthedocs.io/en/latest/#msal.ClientApplication.__init__) once this PR is merged.

@trwalke You do not have to look into the actual implementation, but please help double check with the [API surface](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/68/files#diff-f7ae931bcdab4b14f3a256baf7d6922aR96) and then [the new claim(s) would be sent inside client assertion (rather than, say, as a user assertion like OBO)](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/68/files#diff-f7ae931bcdab4b14f3a256baf7d6922aR160) are aligned with our feature requirement.

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -56,7 +56,8 @@ class ClientApplication(object):
             self, client_id,
             client_credential=None, authority=None, validate_authority=True,
             token_cache=None,
-            verify=True, proxies=None, timeout=None):
+            verify=True, proxies=None, timeout=None,
+            client_claims=None):
         """Create an instance of application.
 
         :param client_id: Your app has a clinet_id after you register it on AAD.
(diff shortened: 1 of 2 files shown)


msal/application.py
```diff
@@ -140,7 +158,8 @@ def _build_client(self, client_credential, authority):
                 client_credential["private_key"], algorithm="RS256",
                 sha1_thumbprint=client_credential.get("thumbprint"), headers=headers)
             client_assertion = signer.sign_assertion(
-                audience=authority.token_endpoint, issuer=self.client_id)
+                audience=authority.token_endpoint, issuer=self.client_id,
+                additional_claims=self.client_claims or {})
```
abhidnya13: I think we can just pass the claims here because we are anyway doing "or empty dictionary" assignment inside the assertion function again. Any thoughts ?

rayluo: We discussed offline. Even though the underlying internal api HAPPENS TO also accepts a `None` as input, this higher level code snippet does not necessarily want to depend on that behavior when we don't have to.

PR #69, opened 2019-06-24
Exclude tests and add license file to pip package

roederja2:
Adding the XML tests files so the tests can be run from a conda build

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -66,7 +66,8 @@ setup(
         'License :: OSI Approved :: MIT License',
         'Operating System :: OS Independent',
     ],
-    packages=find_packages(),
+    packages=find_packages(exclude=["tests"]),
+    data_files=[('', ['LICENSE'])],
     install_requires=[
         'requests>=2.0.0,<3',
         'PyJWT[crypto]>=1.0.0,<2',


roederja2:
Actually thinking about it - we probably want to either exclude the tests module or move it inside the msal module as it currently gets installed as a top level module called "tests" which is probably not what we want.

setup.py
```diff
@@ -67,6 +67,8 @@
         'Operating System :: OS Independent',
     ],
     packages=find_packages(),
+    package_data={'tests': '*.xml'},
```
rayluo: Actually we intended to do the same as you later described "we probably want to either exclude the tests module". So the way to do it, is to modify the previous line to be `packages=find_packages(exclude=["tests"]),`. If you would kindly make that change and roll back this line, we would happily accept it!

setup.py
```diff
@@ -67,6 +67,8 @@
...
+    data_files=[('', ['LICENSE'])],
```
rayluo: I haven't personally used conda. Does msal in conda still ship with an `PKG-INFO` file which would then contain a line as `License: MIT`, just like what pip did? If so, I would suggest to also rollback this line too.

roederja2: Conda Forge is quite keen on actually having an actual license_file: https://conda-forge.org/docs/maintainer/adding_pkgs.html#packaging-the-license-manually

PR #71, opened 2019-06-28
Subject Name and Issuer Authentication

### LINKED ISSUE #60: SubjectName/Issuer (SNI) authentication
Background of SNI: How does it work?
-------------------------------------------

The following description is derived from Matt Bearup, a software developer from Microsoft.

1. User update AAD tenant to accept approved issuers.
   * The AAD Tenant only accepts certain issuers (AME) which have strict association of subject name and requestor
2. User associates an issuer/subject name with their service principal.
   * Once I as the requestor register contoso.com with AME CA, no one else can get it.
   * This is the foundation of SN/Issuer trust. Rather than trusting the thumbprint, you’re trusting that the CA vouches for me as the only one who can generate certs for that subject name.
3. User configures app to use SNIssuer auth instead of thumbprint-based.
   * We don’t need to deprecate thumbprint auth or switch back and forth. The client would simply use a different login option going forward, e.g. `az login --newoption /path/tocert`. MSAL needs to help enable this.
   * As the requester/user it’s still my responsibility to get the latest pfx from keyvault/dsms/etc. An expired cert would still be rejected by AAD.
   * What this saves us is the trouble of registering a new thumbprint with AAD, which would create a more complex autorotation scenario (use old cert to register new cert, then use new cert to deregister old cert).
   * Without SN/Issuer auth there is also a gap in authentication when key material is rotated. When KeyVault issues a new cert, its thumbprint is not in AAD. My app would have to recognize this, rotate thumbprints in AAD, and try again.
   * Note that the key material does get rotated, but that’s not what is being trusted. Instead we’re trusting that an approved CA issued a cert with an approved subject name and that we have the matching private key for that cert.
   * With SN/Issuer auth the *only* thing I have to worry about is getting the latest pfx for my app.

How does it need to be implemented in MSAL?
----------------------------------------------------
* What the “x5c” JWT header value would be? It appears only public certificate is sufficient. MSAL do NOT need to, and MUST NOT send private cerficate.
Sagar: Correct, the x5c is the public cert.

* With “x5c” available, why “x5t” is still required, since it can be generated from “x5c” through sha1 hash easily?
Sagar: This is due to the way JWT header is validated in eSTS for an incoming client assertion. eSTS parses the JWT header and extracts the x5t, does not generate it.

* The ["x5c" parameter means "X.509 Certificate Chain", which is represented as a JSON array of certificate value strings](https://tools.ietf.org/html/rfc7515#section-4.1.6). See [this example](https://tools.ietf.org/html/rfc7515#appendix-B) too. Historically AAD would also accept a single cert as a string. MSAL Python would NOT rely on such behavior, and would still follow this specs and send it as an array. During the implementation, please test it on the wire, and then also come up with a test case to ensure such behavior.

* As a reference, this was the [same feature we added in ADAL Python](https://github.com/AzureAD/azure-activedirectory-library-for-python/pull/173), but it was sending public cert as a string (rather than an array).

* We would also like to include the usability enhancement mentioned in this [ADAL Python issue](https://github.com/AzureAD/azure-activedirectory-library-for-python/issues/199#issuecomment-480088104), and then also back port such enhancement to ADAL Python.

abhidnya13:
- This PR closes #60 
- Adds support for Subject Name Issuer Authentication which is to support cert auto rolls.
- We have tested the end to end scenario by creating our own Test Environment. We will work with the lab team for creating test environment and then integrating in Test automation.
- Note: 
This type of authentication expects authority url to be mentioned with the actual tenant id at the moment. Using organizations will fail the request

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -110,22 +110,17 @@ class ClientApplication(object):
         client_assertion = None
         client_assertion_type = None
         default_body = {"client_info": 1}
+        headers = {}
         if isinstance(client_credential, dict):
             assert ("private_key" in client_credential
                     and "thumbprint" in client_credential)
             if 'public_certificate' in client_credential:
-                if isinstance(client_credential['public_certificate'], list):
-                    public_certificate = client_credential['public_certificate']
-                else:
-                    public_certificate = [client_credential['public_certificate']]
-                signer = JwtSigner(
-                    client_credential["private_key"], algorithm="RS256",
-                    sha1_thumbprint=client_credential.get("thumbprint"),
-                    headers={"x5c": public_certificate})
-            else:
-                signer = JwtSigner(
-                    client_credential["private_key"], algorithm="RS256",
-                    sha1_thumbprint=client_credential.get("thumbprint"))
+                headers["x5c"] = client_credential['public_certificate'] \
+                    if isinstance(client_credential['public_certificate'], list) \
+                    else [client_credential['public_certificate']]  # We send x5c as a list of strings
+            signer = JwtSigner(
+                client_credential["private_key"], algorithm="RS256",
+                sha1_thumbprint=client_credential.get("thumbprint"), headers=headers)
             client_assertion = signer.sign_assertion(
                 audience=authority.token_endpoint, issuer=self.client_id)
             client_assertion_type = Client.CLIENT_ASSERTION_TYPE_JWT
--- a/msal/application.py
+++ b/msal/application.py
@@ -59,7 +59,7 @@ class ClientApplication(object):
             verify=True, proxies=None, timeout=None):
         """Create an instance of application.
 
-        :param client_id: Your app has a clinet_id after you register it on AAD.
+        :param client_id: Your app has a client_id after you register it on AAD.
         :param client_credential:
             For :class:`PublicClientApplication`, you simply use `None` here.
             For :class:`ConfidentialClientApplication`,
@@ -69,8 +69,11 @@ class ClientApplication(object):
                 {
                     "private_key": "...-----BEGIN PRIVATE KEY-----...",
                     "thumbprint": "A1B2C3D4E5F6...",
+                    "public_certificate": "...-----BEGIN CERTIFICATE-----..." (Only to be sent when using Subject Name Issuer Authentication)
                 }
 
+            public_certificate (optional) can be a public key certificate or certificate chain which is sent through
+            'x5c' JWT header only for subject name and issuer based authentication to support cert auto rolls
         :param str authority:
             A URL that identifies a token authority. It should be of the format
             https://login.microsoftonline.com/your_tenant
@@ -115,9 +118,10 @@ class ClientApplication(object):
             assert ("private_key" in client_credential
                     and "thumbprint" in client_credential)
             if 'public_certificate' in client_credential:
-                headers["x5c"] = client_credential['public_certificate'] \
-                    if isinstance(client_credential['public_certificate'], list) \
-                    else [client_credential['public_certificate']]  # We send x5c as a list of strings
+                public_certificates = re.findall(
+                    r'\-+BEGIN CERTIFICATE.+\-+(?P<cert_value>[^-]+)\-+END CERTIFICATE.+\-+',
+                    client_credential['public_certificate'], re.I)  # We send x5c as list of strings
+                headers["x5c"] = [cert.strip() for cert in public_certificates]
             signer = JwtSigner(
                 client_credential["private_key"], algorithm="RS256",
                 sha1_thumbprint=client_credential.get("thumbprint"), headers=headers)
--- a/msal/application.py
+++ b/msal/application.py
@@ -109,15 +109,23 @@ class ClientApplication(object):
     def _build_client(self, client_credential, authority):
         client_assertion = None
         client_assertion_type = None
-        public_certificate = None
         default_body = {"client_info": 1}
         if isinstance(client_credential, dict):
             assert ("private_key" in client_credential
                     and "thumbprint" in client_credential)
-            signer = JwtSigner(
-                client_credential["private_key"], algorithm="RS256",
-                sha1_thumbprint=client_credential.get("thumbprint"),
-                public_certificate = client_credential['public_certificate'])
+            if 'public_certificate' in client_credential:
+                if isinstance(client_credential['public_certificate'], list):
+                    public_certificate = client_credential['public_certificate']
+                else:
+                    public_certificate = [client_credential['public_certificate']]
+                signer = JwtSigner(
+                    client_credential["private_key"], algorithm="RS256",
+                    sha1_thumbprint=client_credential.get("thumbprint"),
+                    headers={"x5c": public_certificate})
+            else:
+                signer = JwtSigner(
+                    client_credential["private_key"], algorithm="RS256",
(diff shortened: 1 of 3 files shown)


msal/application.py
```diff
@@ -69,8 +69,11 @@ def __init__(
                 {
                     "private_key": "...-----BEGIN PRIVATE KEY-----...",
                     "thumbprint": "A1B2C3D4E5F6...",
+                    "public_certificate": "...-----BEGIN CERTIFICATE-----..." (Only to be sent when using Subject Name Issuer Authentication)
                 }
 
+            public_certificate (optional) can be a public key certificate or certificate chain which is sent through
```
rayluo: Although our implementation here will actually support sending multiple certs, but since AAD works fine with only one cert, I incline to NOT mention the "or certificate chain" part at all, in order to discourage the app dev to actually configure a cert chain. The chain would still work, but the side effect is it would significantly increase the request payload size, which we better avoid.

By the way, if this paragraph can be limited to [roughly 80 chars per line](https://www.python.org/dev/peps/pep-0008/#maximum-line-length), that would be great! :-)

msal/application.py
```diff
@@ -69,8 +69,11 @@ def __init__(
                 {
                     "private_key": "...-----BEGIN PRIVATE KEY-----...",
                     "thumbprint": "A1B2C3D4E5F6...",
+                    "public_certificate": "...-----BEGIN CERTIFICATE-----..." (Only to be sent when using Subject Name Issuer Authentication)
```
rayluo: Given that we will have a dedicated paragraph to talk about this, I think we can replace that long "(Only to be sent ...)" part with a short message, something like "(Optional. See below.)".

rayluo:
Oh this is an extra requirement beyond the original task scope: we will probably want to investigate how to automatically [compute the thumbprint from the public cert](https://stackoverflow.com/questions/6682815/deriving-an-ssh-fingerprint-from-a-public-key-in-python). That would save app dev the trouble of updating thumbprint when they rotate their certs.

msal/application.py
```diff
@@ -106,16 +110,27 @@ def __init__(
         self.client = self._build_client(client_credential, self.authority)
         self.authority_groups = None
 
+    def _extract_x5c_value(self):
```
rayluo: Glad to see improvement here comparing to the previous implementation! Thanks! And now it seems we can go even further to simplify its tests, if we refactor to completely separate this helper from this class, just like how we implemented the existing `decorate_scope(...)`. Let's try something like this:

```python
def extract_certs(pem_content):
    # Parse raw pem content and returns a list of string
    # Usage: headers = {"x5c": extract_certs(open("my_cert.pem").read())}
    ...
    return ...
```

Now its test cases won't couple with creating an app instance, yet give us better test coverage on [the 3 scenarios](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/71#discussion_r298740352), with less dependency, and also happen to be more concise.

```python
def test_extract_a_tag_less_cert(self):
    pem = "my_cert"
    self.assertEqual(["my_cert"], extract_certs(pem))

def test_extract_a_tag_enclosed_cert(self):
    pem = """
-----BEGIN CERTIFICATE-----
my_cert
-----END CERTIFICATE-----
"""
    self.assertEqual(["my_cert"], extract_certs(pem))

def test_extract_multiple_tag_enclosed_certs(self):
    pem = """
-----BEGIN CERTIFICATE-----
my_cert1
-----END CERTIFICATE-----

-----BEGIN CERTIFICATE-----
my_cert2
-----END CERTIFICATE-----
"""
    self.assertEqual(["my_cert1", "my_cert2"], extract_certs(pem))
```

Now we can remove the 3 `def test_subject_name_issuer_authentication_*(...)` test cases.

tests/test_application.py
```diff
@@ -99,6 +99,76 @@ def test_client_certificate(self):
...
+
+    @unittest.skipUnless("public_certificate" in CONFIG, "Missing Public cert")
+    def test_subject_name_issuer_authentication(self):
+        assert ("private_key_file" in CONFIG
+                and "thumbprint" in CONFIG and "public_certificate" in CONFIG)
+        key_path = os.path.join(THIS_FOLDER, CONFIG['private_key_file'])
+        public_certificate = os.path.join(THIS_FOLDER, CONFIG['public_certificate'])
+        with open(key_path) as f:
+            pem = f.read()
+        with open(public_certificate) as f:
```
rayluo: Rather than defining a variable just 3 lines above and then override its content (thus its meaning) 1 line below, let's just do this, thus avoiding one of the [most difficult things in computer science](https://image.slidesharecdn.com/naming-3-170610132552/95/2nd-hardest-problem-in-computer-science-3-638.jpg?cb=1497101264) :-)

```python
with open(os.path.join(...)) as f:
    public_certificate = f.read()
```

PS: The `key_path` and `pem` variables did not have such naming conflict, but you can go ahead to also apply the same technique. But we do NOT need to retrofit this new technique to pre-existing test cases.

abhidnya13:
Thanks for the review Ray ! This is really good learning for me :) I have made the changes. Let me know if it looks good now. Feel free to suggest anything else that we can do to make it better before we merge it in. 
Lets look into how to automatically compute the thumbprint from the public cert. That will make it easier for the developers for sure.

msal/application.py
```diff
@@ -50,6 +50,17 @@ def decorate_scope(
     return list(decorated)
 
 
+def extract_certs(public_cert_content):
+    # Parses raw public certificate file contents and returns a list of strings
+    # Usage: headers = {"x5c": extract_certs(open("my_cert.pem").read())}
+    public_certificates = re.findall(
+        r'\-+BEGIN CERTIFICATE.+\-+(?P<cert_value>[^-]+)\-+END CERTIFICATE.+\-+',
```
rayluo: Based on the [PEM format definition](https://tools.ietf.org/html/rfc7468#section-2), it seems the two `.+` in existing regex are redundant. Also we can reduce the use of `\-+` regex magic, in favor of plain text.

Will this one be slightly more readable? `r"(?:-----BEGIN CERTIFICATE-----)(?P<cert_value>[^-]+)(?:-----END CERTIFICATE-----)"`
By the way, there are plenty of [online regex playground](https://regex101.com/) for quick prototyping.

msal/application.py
```diff
@@ -50,6 +50,17 @@ def decorate_scope(
...
+        public_cert_content, re.I)
+    if len(public_certificates):
```
rayluo: Given that the return value is already a list, possibly empty, the pythonic way to detect a non-empty list is just `if public_certificates:`.

msal/application.py
```diff
@@ -50,6 +50,17 @@ def decorate_scope(
...
+        return [cert.strip() for cert in public_certificates]
+    return [public_cert_content.strip()]
```
rayluo: Before this line, we should add an extra detection as precaution. PEN test team will probably suggest the same.

```python
# The public cert tags are not found in the input,
# let's make best effort to exclude a private key pem file.
if "PRIVATE KEY" in public_cert_content:
    raise ValueError("We expect your public key but detect a private key instead")
return [public_cert_content.strip()]  # Return raw content as-is
```

msal/application.py
abhidnya13: Yes ! I did use that to try out different things. I can make it more readable by doing that just a little differently. Will address in the next commit. Stay tuned :)

PR #74, opened 2019-07-10
Pack LICENSE in a different way

rayluo:
It turns out [the approach we used in PR #69](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/69#discussion_r296942855) would cause the LICENSE being copied into `/usr/local`, which would typically end up with insufficient permission error while some downstream build pipeline (such as, OSX) attempts to install msal without explicitly specifying `--user` option.

This PR uses a different to pack that LICENSE file and ship it inside this library's directory.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -67,7 +67,7 @@ setup(
         'Operating System :: OS Independent',
     ],
     packages=find_packages(exclude=["tests"]),
-    data_files=[('', ['LICENSE'])],
+    package_data={'': ['LICENSE']},  # Do not use data_files=[...]. See https://stackoverflow.com/questions/14211575
     install_requires=[
         'requests>=2.0.0,<3',
         'PyJWT[crypto]>=1.0.0,<2',
--- a/setup.py
+++ b/setup.py
@@ -67,7 +67,10 @@ setup(
         'Operating System :: OS Independent',
     ],
     packages=find_packages(exclude=["tests"]),
-    package_data={'': ['LICENSE']},  # Do not use data_files=[...]. See https://stackoverflow.com/questions/14211575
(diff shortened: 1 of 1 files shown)


roederja2:
I can't find the LICENSE file in the pypi package anymore after this change.

rayluo:
Hmm, @roederja2 , thanks for bringing this to our attention. We don't immediately have a solution, but we can be transparent on what we know so far.

Admittedly, when working on this very PR (#74) we assumed that [the canonical `package_data` would do its job](https://setuptools.readthedocs.io/en/latest/setuptools.html#including-data-files), therefore did not really check the actual outcome. When I tried it just now with `pip install msal==0.4.1` i.e. the version without either the recent `data_files` or `package_data` adjustment, and comparing it to `pip install msal==0.5.0` i.e. the version with `data_files`, and then the `pip install msal==0.5.1` i.e. the version with the latest `package_data`, they *ALL* install a "my_virtual_env/Lib/site-packages/msal-0.x.x.dist-info/LICENSE" into my environment. In that sense, this `package_data` param seems like an no-op.

On the other hand, the `pip install msal==0.5.0` i.e. the `data_files` version indeed generates one extra file at "my_virtual_env/LICENSE". That behavior unnecessarily caused downstream pipeline trouble of permission issue, because, if not using a virtual environment, that LICENSE would be attempted to be written into `/usr/local` on OSX. Even if some sort of virtual environment is being used, that behavior is not correct to begin with, because that "my_virtual_env/LICENSE" location does not belong to msal library, and it would possibly be overwritten by other libraries which happen to use same way to distribute their LICENSE, hence defeat the purpose.

We are open to hear a better solution on this. At least, next time we will know how to test it.

PR #81, opened 2019-07-29
Adding YAML front matter for samples

rwike77:


jmprieur:
@rwike77 : I don't think we should expose the full library in the Azure Samples browser.
We are building proper samples in Azure-Samples.

PR #84, opened 2019-08-11
Use urlsafe_b64decode instead of b64decode

kit494way:
- This PR fixes the error that occurs in decoding token.
- The ID token is encoded as base64url, not base64.
- In the previous implementation, [device_flow_sample.py](https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/dev/sample/device_flow_sample.py) will fails with error `binascii.Error: Incorrect padding`.


rayluo:
Hi Yasutaka san, thanks for the catch! We will look into the test failure caused by this PR on Python 2.7, before we can merge this in.

If you don't mind, can you also send your ID token to [my email (which you can find here)](https://github.com/rayluo) so that I can test the issue and the fix by myself? This is totally optional, and we will still appreciate your help so far even without that test sample.

PR #89, opened 2019-08-27
Test Automation

rayluo:
A new end-to-end test mechanism based on lab infrastructure. After several milestones, this new mechanism itself is now ready for a review.

~~TODO: Refactor some pre-existing end-to-end test cases from test_application.py into this new test_e2e.py file, so that test_application no longer incur any network call.~~ Done in commit 6ea7890. Ready for review now.

### DIFF
--- a/tests/test_e2e.py
+++ b/tests/test_e2e.py
@@ -1,6 +1,7 @@
 import logging
 import os
 import json
+import time
 
 import requests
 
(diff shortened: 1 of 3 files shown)


tests/test_e2e.py
```diff
@@ -0,0 +1,329 @@
...
+            self.id(), json.dumps(self.app.token_cache._cache, indent=4))
+        self.assertIn(
+            "access_token", result,
+            "{error}: {error_description}".format(
+                # Note: No interpolation here, cause error won't always present
+                error=result.get("error"),
+                error_description=result.get("error_description")))
+        self.assertCacheWorksForUser(result, self.config["scope"], username=None)
+
+    def test_client_secret(self):
```
abhidnya13: Can we add this one in the set of automation tests using lab api ?

rayluo: Good point! Sum up our offline conversation:

* Currently we implicitly tested the confidential client with secret scenario, by becoming a confidential client to call lab user api and lab password api
* There seems no "app api" from the lab available at this time.
* In some future effort (as a separated PR), we will eventually get there, to cover even more than confidential client with secret, such as confidential client with certificate, and Subject Name Issuer (SNI) scenario. But right now we won't let this PR to block our upcoming other feature PRs.

Issue #53, opened 2019-06-05
Support On-Behalf-Of flow


rayluo:
The following description is derived from a [description](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/issues/482#issuecomment-262798342), written by @psignoret:

---

To illustrate how this works, let's suppose you have the following:

1. A  client application that the end-user interacts with: *ClientApp*.
2. A backend API used by your client app(s): *BackendAPI* (which has no user interface whatsoever, but is configured to use Azure AD for authentication).
3. A third-party API that you have no control over, that requires user authentication via Azure AD: *DynamicsAPI* (which also has no user interface whatsoever).

With the on-behalf-of flow, the following would happen:

1. The user signs in to *ClientApp*, by using any of the flows that include the user using the normal browser-based login flows. If the user needs to do MFA, provide consent, change their password, or any other sign-in interruption, they can do that. At the end of the sign-in, *ClientApp* will have an access tokenin the name of the authenticated user, destined for *BackendAPI*.
2. *ClientApp* will present this access token to *BackendAPI* (e.g. in the `Authorization` HTTP header).
3. *BackendAPI* will validate this token (e.g. check that its signature is good, that the claims are valid, etc.).
4. Now, *BackendAPI* can't call *DynamicsAPI* and provide the same access token it was given (since it is destined for *BackendAPI*, and *DynamicsAPI* will reject it). Instead, *BackendAPI* needs to get a new access token, *on behalf of* the signed in user. To do this, *BackendAPI* makes an authenticated token request (meaning *BackendAPI* authenticates itself with a client secret or certificate) to Azure AD. This request includes:
    - An indication that the resulting access token should be for *DynamicsAPI*.
    - An indication that this is an "on-behalf-of" (OBO) request.
    - The original access token that *BackendAPI* received (remember, to get this token, *ClientApp* put the user through a sign-in, so it includes all the details about the user who signed in).
5. If everything checks out, Azure AD will respond with an access token for *DynamicsAPI*, on behalf of the signed in user. *BackendAPI* can now call *DynamicsAPI* with that token.

Step 4, above, will be very easily invoked with MSAL Python:

 - `app.acquire_token_on_behalf_of(user_assertion, scopes)`

[...]

PR #92, opened 2019-09-12
On-behalf-of (OBO) Implementation

rayluo:
What
-----
This PR will solve #53.

How was it implemented
--------------------------
While the actual implementation is technically a one-liner `return ...`, there are some design decisions documented as inline comments inside both the implementation and the test case. Review/feedbacks are welcome.

The configuration of test case is reused from [a similar test case in MSAL .Net](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/4.3.1/tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/OboTests.cs).

How the new API looks like
-----------------------------
This is the [updated API reference document](https://msal-python.readthedocs.io/en/obo/#msal.ConfidentialClientApplication.acquire_token_on_behalf_of), please also proofread.

How to install this not-yet-released feature for testing
-----------------------------------------------
If you prefer to, you can install this branch for your smoke testing:

    pip install git+https://github.com/AzureAD/microsoft-authentication-library-for-python.git@obo

What's Next
-------------
* We will provide an OBO sample at a later time, as a separated repo.
* We may provide a wiki page, similar to [the MSAL .Net one](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/on-behalf-of)
* Figure out how to add/modify [this docs page](https://docs.microsoft.com/en-us/azure/active-directory/develop/scenario-web-api-call-api-app-configuration) to include a code snippet for MSAL Python

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -646,10 +646,19 @@ class ConfidentialClientApplication(ClientApplication):  # server-side web app
         representing an end user.
         The current app can use such token (a.k.a. a user assertion) to request
         another token to access downstream service, on behalf of that user.
+        See `detail docs here <https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-on-behalf-of-flow>`_ .
 
         The current middle-tier app has no user interaction to obtain consent.
         See how to gain consent upfront for your middle-tier app from this article.
         https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-on-behalf-of-flow#gaining-consent-for-the-middle-tier-application
+
+        :param str user_assertion: The incoming token already received by this app
+        :param list[str] scopes: Scopes required by downstream API (a resource).
+
+        :return: A dict representing the json response from AAD:
+
+            - A successful response would contain "access_token" key,
+            - an error response would contain "error" and usually "error_description".
         """
         # The implementation is NOT based on Token Exchange
         # https://tools.ietf.org/html/draft-ietf-oauth-token-exchange-16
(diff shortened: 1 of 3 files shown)


msal/application.py
```diff
@@ -639,6 +639,38 @@ def acquire_token_for_client(self, scopes, **kwargs):
                 scope=scopes,  # This grant flow requires no scope decoration
                 **kwargs)
 
-    def acquire_token_on_behalf_of(self, user_assertion, scopes, authority=None):
-        raise NotImplementedError()
+    def acquire_token_on_behalf_of(self, user_assertion, scopes, **kwargs):
+        """Acquires token using on-behalf-of (OBO) flow.
+
+        The current app is a middle-tier service which already receives a token
```
jmprieur: which already receives a token => which was called with a token representing an end user?

msal/application.py
```diff
@@ -639,6 +639,38 @@ def acquire_token_for_client(self, scopes, **kwargs):
...
+        representing an end user.
+        The current app can use such token (a.k.a. a user assertion) to request
+        another token to access downstream service, on behalf of that user.
```
jmprieur: downstream service => a downstream web API ?

rayluo:
@jmprieur : Here we quote [your question in our Release PR](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/100#issuecomment-535363151) below, and then answer it here.

> Q: is the plan to work on an OBO sample?

Yes. We actually mentioned that as the first to-do item in the "What's Next" section in the [current feature PR's description](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/92#issue-317102125). Perhaps that to-do list itself can become a new workitem/issue in its own.

PR #103, opened 2019-10-01
Add correlation id, public api id, and current request header

sangonzal:
Quoted from [a conversation](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/103#discussion_r337651275):

Link to API ID doc: https://microsoft.sharepoint.com/:w:/t/aad/devex/EZGF46opilBKrgC59mUFttIBebPBY6MMGLO1eGz0uXQh6w?e=TSq76f

Note that these values are not in there yet. The way that this has worked in the past for .NET and Java is that we have picked values that fit within the ranges specified which have not been taken, then we'd update the doc.

We are working with some of the other teams to make sure that most up to date values are in there. For that, we have this spreadsheet: https://microsoft-my.sharepoint-df.com/:x:/p/sagonzal/EXSrr4vM1utAqQQfD6bMln4BYKwjrqh3cagiNJWPVNjLzw?e=G5FybL


msal/application.py
```diff
@@ -68,6 +69,16 @@ def extract_certs(public_cert_content):
...
 
+    ACQUIRE_TOKEN_SILENT_ID = "84"
+    ACQUIRE_TOKEN_BY_USERNAME_PASSWORD_ID = "301"
+    ACQUIRE_TOKEN_ON_BEHALF_OF_ID = "523"
+    INITIATE_DEVICE_FLOW = "621"
+    ACQUIRE_TOKEN_BY_DEVICE_FLOW_ID = "622"
+    ACQUIRE_TOKEN_FOR_CLIENT_ID = "730"
+    ACQUIRE_TOKEN_BY_AUTHORIZATION_CODE_ID = "832"
+    GET_ACCOUNTS_ID = "902"
+    REMOVE_ACCOUNT_ID = "903"
```
sangonzal: get_accounts and remove_account id not being used at the moment, but will be when we include a solution for last request header

msal/application.py
```diff
@@ -498,7 +512,7 @@ def _get_app_metadata(self, environment):
 
     def _acquire_token_silent_by_finding_specific_refresh_token(
             self, authority, scopes, query,
-            rt_remover=None, break_condition=lambda response: False, **kwargs):
+            rt_remover=None, break_condition=lambda response: False, force_refresh=False, **kwargs):
```
abhidnya13: You can use the **kwargs to send any additional parameters to this function instead of adding a new parameter

rayluo: Yes Abhi and I briefly talked on this one yesterday. Either the implicit `**kwargs` or the explicit `force_refresh` here would work. Now on closer inspection, I think it is OK to use an explicit `force_refresh` here, because the later inside this function, we do use that parameter. But then, some extra changes are needed:

1. You would need to pull in the latest dev branch which contains a bugfix relevant to this feature.
2. You would need to modify an "outer" level method `_acquire_token_silent_from_cache_and_possibly_refresh_it()` to wire up this `force_refresh`, only then it would be available here inside the innermost method.
3. Given that these intrinsic complexity of the acquire_token_silent(...) family, I would suggest let's see if we can add a test case for this function. You can find the [pattern to test in this pre-existing test case](https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/0.7.0/tests/test_application.py#L75-L81). That would have helped you to catch the above issue.
4. Lastly, per PEP 8 (even though we have not strictly enforce it, yet), I would suggest you move this new parameter into its own line, so that the current line remains within the recommend [max line length](https://www.python.org/dev/peps/pep-0008/#maximum-line-length).

msal/application.py
```diff
@@ -533,6 +551,12 @@ def _validate_ssh_cert_input_data(self, data):
                     "you must include a string parameter named 'key_id' "
                     "which identifies the key in the 'req_cnf' argument.")
 
+    def _get_new_correlation_id(self):
```
rayluo: @sangonzal Good job on noticing the underscore prefix naming convention for non-public methods!

Given that these 2 new helpers do not share any data with the `ClientApplication` class, I would suggest to put them as 2 highest level functions.

msal/application.py
```diff
@@ -597,13 +626,18 @@ def acquire_token_by_username_password(
             - an error response would contain "error" and usually "error_description".
         """
         scopes = decorate_scope(scopes, self.client_id)
+        headers = {'client-request-id': self._get_new_correlation_id(),
+                   'x-client-current-telemetry': self._build_current_telemetry_request_header(
+                       self.ACQUIRE_TOKEN_BY_USERNAME_PASSWORD_ID)}
         if not self.authority.is_adfs:
-            user_realm_result = self.authority.user_realm_discovery(username)
+            user_realm_result = self.authority.user_realm_discovery(username, headers)
```
rayluo: Question: Does our API IDs doc define the User Realm Discovery (URD) to use the same API ID as the AcquireTokenByUsernamePassword one? On the wire, they are 2 different requests. Unless the API ID doc specifically requires us to treat them as the same, I would suggest we can skip telemetry for this undefined URD request, otherwise it would cause our Username Password flow telemetry amount to be doubled. Can you share the API IDs doc link to help us confirm that?

> In the face of ambiguity, refuse the temptation to guess. -- Quoted from the [Zen](https://www.python.org/dev/peps/pep-0020/).

msal/authority.py
```diff
@@ -58,11 +58,11 @@ def __init__(self, authority_url, validate_authority=True,
         _, _, self.tenant = canonicalize(self.token_endpoint)  # Usually a GUID
         self.is_adfs = self.tenant.lower() == 'adfs'
 
-    def user_realm_discovery(self, username):
+    def user_realm_discovery(self, username, headers):
```
rayluo: If we ended up really need to add telemetry for this User Realm Discovery behaviour (see the comment above), then let's put the headers as an optional parameter with default value as `None`. Because this header thing is NOT part of this method's core business.

msal/application.py
sangonzal: Link to API ID doc: https://microsoft.sharepoint.com/:w:/t/aad/devex/EZGF46opilBKrgC59mUFttIBebPBY6MMGLO1eGz0uXQh6w?e=TSq76f

The document does not explicitly answer your question. My 2 cents:  API Id is an identifier for a public API request, so all operations that happen as part of the public API call should have the same public API Id . Having said that, we have updated the telemetry design to only include the telemetry headers to the token endpoint. In this case, we would get rid of  'x-client-current-telemetry' for this request, but IMO we should still include the correlation Id header ('client-request-id'), as when you search for correlation Id in server side logs, you would want to see all requests (including user realm and instance discovery)

msal/application.py
sangonzal: Link to API ID doc: https://microsoft.sharepoint.com/:w:/t/aad/devex/EZGF46opilBKrgC59mUFttIBebPBY6MMGLO1eGz0uXQh6w?e=TSq76f

Note that these values are not in there yet. The way that this has worked in the past for .NET and Java is that we have picked values that fit within the ranges specified which have not been taken, then we'd update the doc. 

We are working with some of the other teams to make sure that most up to date values are in there. For that, we have this spreadsheet: https://microsoft-my.sharepoint-df.com/:x:/p/sagonzal/EXSrr4vM1utAqQQfD6bMln4BYKwjrqh3cagiNJWPVNjLzw?e=G5FybL

msal/application.py
```diff
@@ -275,6 +286,9 @@ def acquire_token_by_authorization_code(
             data=dict(
                 kwargs.pop("data", {}),
                 scope=decorate_scope(scopes, self.client_id)),
+            headers={'client-request-id', self._get_new_correlation_id(),
```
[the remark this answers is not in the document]
sangonzal: Good catch. I have fixed it. I have not validated all flows (was only testing with username password as that was the easiest one to set up), but plan to do so today.

sangonzal:
I have tested all flows and have validated that data is showing up in Kusto

sangonzal:
@rayluo @abhidnya13  are there any plans to merge this in? As far as I can tell, all feedback has been addressed. 
cc: @henrik-me

msal/oauth2cli/oauth2.py
```diff
@@ -214,9 +213,9 @@ def initiate_device_flow(self, scope=None, timeout=None, **kwargs):
         if not self.configuration.get(DAE):
             raise ValueError("You need to provide device authorization endpoint")
         flow = self.session.post(self.configuration[DAE],
-            data={"client_id": self.client_id, "scope": self._stringify(scope or [])},
-            timeout=timeout or self.timeout,
-            **kwargs).json()
+                                 data={"client_id": self.client_id, "scope": self._stringify(scope or [])},
```
rayluo: There seems no substantial changes at all. I think you can simply revert changes to this entire file.

msal/application.py
```diff
@@ -605,13 +635,18 @@ def acquire_token_by_username_password(
             - an error response would contain "error" and usually "error_description".
         """
         scopes = decorate_scope(scopes, self.client_id)
+        headers = {'client-request-id': _get_new_correlation_id(),
+                   'x-client-current-telemetry': _build_current_telemetry_request_header(
+                       self.ACQUIRE_TOKEN_BY_USERNAME_PASSWORD_ID)}
         if not self.authority.is_adfs:
-            user_realm_result = self.authority.user_realm_discovery(username)
+            user_realm_result = self.authority.user_realm_discovery(username, headers['client-request-id'])
```
```suggestion
            user_realm_result = self.authority.user_realm_discovery(username, correlation_id=headers['client-request-id'])
```

msal/application.py
```diff
@@ -562,6 +588,7 @@ def initiate_device_flow(self, scopes=None, **kwargs):
             - an error response would contain some other readable key/value pairs.
         """
         return self.client.initiate_device_flow(
+            headers={'client-request-id': _get_new_correlation_id()},
```
rayluo: Don't you mean to do this instead?
```suggestion
            headers={
                'client-request-id': _get_new_correlation_id(),
                'x-client-current-telemetry': _build_current_telemetry_request_header(
                    self.INITIATE_DEVICE_FLOW),
                },
```

msal/application.py
```diff
@@ -519,6 +542,9 @@ def _acquire_token_silent_by_finding_specific_refresh_token(
                 entry, rt_getter=lambda token_item: token_item["secret"],
                 on_removing_rt=rt_remover or self.token_cache.remove_rt,
                 scope=scopes,
+                headers={'client-request-id': _get_new_correlation_id(),
+                         'x-client-current-telemetry': _build_current_telemetry_request_header(
+                            self.ACQUIRE_TOKEN_SILENT_ID, force_refresh)},
```
```suggestion
                            self.ACQUIRE_TOKEN_SILENT_ID, force_refresh=force_refresh)},
```

msal/application.py
```diff
@@ -283,6 +302,9 @@ def acquire_token_by_authorization_code(
             data=dict(
                 kwargs.pop("data", {}),
                 scope=decorate_scope(scopes, self.client_id)),
+            headers={'client-request-id': _get_new_correlation_id(),
+                     'x-client-current-telemetry': _build_current_telemetry_request_header(
+                self.ACQUIRE_TOKEN_BY_AUTHORIZATION_CODE_ID)},
```
```suggestion
                          self.ACQUIRE_TOKEN_BY_AUTHORIZATION_CODE_ID)},
```

msal/application.py
```diff
@@ -585,6 +612,9 @@ def acquire_token_by_device_flow(self, flow, **kwargs):
                     # 2018-10-4 Hack:
                     # during transition period,
                     # service seemingly need both device_code and code parameter.
+                headers={'client-request-id': _get_new_correlation_id(),
```
rayluo: The following topic is more about Device Flow.

Shall we consider to let the initial Device Flow call and all its subsequent polling to have same `client-request-id`? It is easy to implement too. We just store the first id as `_client_request_id` in the `flow` object, and then reuse it here. Something like this:
```suggestion
                headers={
                    'client-request-id': flow.get("_client_request_id", _get_new_correlation_id()),
```

msal/application.py
sangonzal: telemetry data is only sent to the /authorize or /token endpoint, so in this case we wouldn't include the header.

msal/oauth2cli/oauth2.py
sangonzal: No changes to logic, but fixes spacing and indentation that breaks PEP 8. I reverted changes, you can update that later.

msal/application.py
rayluo: In that case, we might as well completely remove that `INITIATE_DEVICE_FLOW` api id definition.

msal/application.py
sangonzal: I'm not sure how server generates device codes - I think it's a safer bet to use UUIDs to avoid collisions.

msal/application.py
sangonzal: You're right - no need for INITIATE_DEVICE_FLOW. Also added comment

msal/application.py
sangonzal: I'm not sure how server generates device code. I would rather use UUID as, as their is less chance of collision (and this is what we use for every other flow, don't see reason to change it here)

PR #104, opened 2019-10-04
B2c implementation

### LINKED ISSUE #52: Support Azure AD B2C
Support [Azure AD B2C](https://azure.microsoft.com/en-us/services/active-directory-b2c/)

rayluo:
There is an automated test case in this PR to cover username password flow (i.e. ROPC grant), and it passes.

Currently, this implementation is still considered in an early stage. We can use it as a prototype to [work on some sample as proof-of-concept](https://github.com/Azure-Samples/ms-identity-python-webapp/pull/4), in order to get the whole picture of end-to-end experience of the "new" API surface and calling patterns. The interesting part of this PR, is something NOT in this PR: You will see this PR did NOT touch the source file `application.py` which contains current public API definition. That means, we keep the high level API surface unchanged for B2C and non-B2C scenarios.

We can install this branch by:

    pip install git+https://github.com/AzureAD/microsoft-authentication-library-for-python.git@b2c

We can organize a code walk-through when we officially start the PR code review at a later time.

This PR will eventually resolve #52 .

### DIFF
--- a/msal/authority.py
+++ b/msal/authority.py
@@ -23,6 +23,8 @@ class Authority(object):
     Once constructed, it contains members named "*_endpoint" for this instance.
     TODO: It will also cache the previously-validated authority instances.
     """
+    _domains_without_user_realm_discovery = set([])
+
     def __init__(self, authority_url, validate_authority=True,
             verify=True, proxies=None, timeout=None,
             ):
@@ -67,17 +69,21 @@ class Authority(object):
         _, _, self.tenant = canonicalize(self.token_endpoint)  # Usually a GUID
         self.is_adfs = self.tenant.lower() == 'adfs'
 
-    def user_realm_discovery(self, username):
-        resp = requests.get(
-            "https://{netloc}/common/userrealm/{username}?api-version=1.0".format(
-                netloc=self.instance, username=username),
-            headers={'Accept':'application/json'},
-            verify=self.verify, proxies=self.proxies, timeout=self.timeout)
-        resp.raise_for_status()
-        return resp.json()
-        # It will typically contain "ver", "account_type",
+    def user_realm_discovery(self, username, response=None):
+        # It will typically return a dict containing "ver", "account_type",
         # "federation_protocol", "cloud_audience_urn",
         # "federation_metadata_url", "federation_active_auth_url", etc.
+        if self.instance not in self.__class__._domains_without_user_realm_discovery:
+            resp = response or requests.get(
+                "https://{netloc}/common/userrealm/{username}?api-version=1.0".format(
+                    netloc=self.instance, username=username),
+                headers={'Accept':'application/json'},
+                verify=self.verify, proxies=self.proxies, timeout=self.timeout)
+            if resp.status_code != 404:
+                resp.raise_for_status()
+                return resp.json()
+            self.__class__._domains_without_user_realm_discovery.add(self.instance)
+        return {}  # This can guide the caller to fall back normal ROPC flow
 
 def canonicalize(url):
     # Returns (canonicalized_url, netloc, tenant). Raises ValueError on errors.
--- a/msal/authority.py
+++ b/msal/authority.py
@@ -1,4 +1,7 @@
-import re
+try:
+    from urllib.parse import urlparse
+except ImportError:  # Fall back to Python 2
+    from urlparse import urlparse
 import logging
 
 import requests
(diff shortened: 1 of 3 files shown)


tests/test_e2e.py
```diff
@@ -417,3 +435,49 @@ def test_acquire_token_obo(self):
             result = cca.acquire_token_silent(downstream_scopes, account)
             self.assertEqual(cca_result["access_token"], result["access_token"])
 
+    def _build_b2c_authority(self, policy):
+        base = "https://msidlabb2c.b2clogin.com/msidlabb2c.onmicrosoft.com"
+        return base + "/" + policy  # We do not support base + "?p=" + policy
```
jmprieur: it's probably fine as supporting B2C in MSAL.Python it's a new feature.

tests/test_e2e.py
```diff
@@ -13,6 +13,21 @@
 logging.basicConfig(level=logging.INFO)
 
 
+def _get_app_and_auth_code(
+        client_id,
+        client_secret=None,
+        authority="https://login.microsoftonline.com/common",
+        port=44331,
+        scopes=["https://graph.windows.net/.default"],
```
jmprieur: We try to move everything from AAD Graph to Microsoft Graph.
Would it make sense to use https://graph.microsoft.com/.default instead?

msal/authority.py
```diff
@@ -37,18 +42,30 @@ def __init__(self, authority_url, validate_authority=True,
...
-        canonicalized, self.instance, tenant = canonicalize(authority_url)
-        tenant_discovery_endpoint = (
-            'https://{}/{}{}/.well-known/openid-configuration'.format(
-                self.instance,
-                tenant,
-                "" if tenant == "adfs" else "/v2.0" # the AAD v2 endpoint
-                ))
-        if (tenant != "adfs" and validate_authority
+        authority, self.instance, tenant = canonicalize(authority_url)
+        is_b2c = self.instance.endswith(".b2clogin.com")
```
jennyf19: only issue here is that we also have to support b2c in the sovereign clouds...so these also need to work:
 *.b2clogin.us, *.b2clogin.cn, and *.b2clogin.de

That's why MSAL.NET is stuck on the /tfp/ validation to determine if this is a b2c authority or not. if b2c has instance discovery, like ests, this would not be needed.

msal/authority.py
```diff
@@ -37,18 +42,30 @@ def __init__(self, authority_url, validate_authority=True,
...
-            tenant_discovery_endpoint = instance_discovery(
-                canonicalized + "/oauth2/v2.0/authorize",
+            payload = instance_discovery(
+                "https://{}{}/oauth2/v2.0/authorize".format(
+                    self.instance, authority.path),
                 verify=verify, proxies=proxies, timeout=timeout)
+            if payload.get("error") == "invalid_instance":
+                raise ValueError(
+                    "invalid_instance: "
+                    "The authority you provided, %s, is not whitelisted. "
```
jennyf19: nit: maybe use "on the allow list" instead...idk.

msal/authority.py
rayluo: Good point. I did not know B2C sovereign hosts before. Now I've added them by the latest commit, so it would also work for the sovereign clouds you just mentioned. Hello world, happy B2C! :-)

msal/authority.py
rayluo: Actually I thought about several options when I drafted that sentence. To me, "allow" does not sound right, because customer can of course choose a customized domain `Contoso.com` and we MSFT is in no position to disallow that. The precise cause of this error message is that the `Contoso.com` is not a so-called "well known" domain to MSFT Identity platform. And then again, it feels a bit weird to comment on other's name as "not well known", so I think "not whitelisted" sounds more neutral in tone. What do you think?

tests/test_e2e.py
rayluo: To be clear, that inline comment is inside an internal test case, so it is more like a remind-to-ourselves, but did not mean it as a limitation or deficiency. In fact, currently all the public materials we can find, uses the "base/policy" format, rather than "base?p=policy" format. Such as:

* The official B2C doc, like [this one](https://docs.microsoft.com/en-us/azure/active-directory-b2c/active-directory-b2c-reference-oauth-code#1-get-an-authorization-code),
* The [MSAL .Net wiki page](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/AAD-B2C-specifics#authority-for-a-b2c-tenant-and-policy)
* The [test cases in MSAL Java](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/v0.6.0-preview/src/test/java/com/microsoft/aad/msal4j/TestConfiguration.java#L32-L35)

That being said, I did heard from a previous internal presentation that the "base?p=policy" format would also work. We (the entire MSAL fleet) might actually want to switch to that format at a later time, because the current way - putting policy inside authority - has some intrinsic issues, but that is a different topic.

PR #105, opened 2019-10-10
Customizable response_type

rayluo:
Wire up and document the `get_authorization_request_url(..., response_type=...)` parameter.

It would be useful in scenario when a web app just want to acquire ID Token to sign in only.


jmprieur:
LGTM. Useful indeed. ASP.NET has a similar setting

PR #107, opened 2019-10-15
Enable Python 3.8 in test automation

rayluo:
[Python 3.8.0](https://www.python.org/downloads/release/python-380/) has been released yesterday. Now we include it in our test automation matrix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,9 @@ matrix:
     - python: 3.7
       dist: xenial
       sudo: true
+    - python: 3.8
+      dist: xenial
+      sudo: true
 
 install:
   - pip install -r requirements.txt


.travis.yml
```diff
@@ -11,6 +11,9 @@ matrix:
     - python: 3.7
```
henrik-me: >- python: 3.7 [](start = 4, length = 13)

remove ?

henrik-me:
>   - "3.6"

does new options need to be added here? If so are those needed in the include section?

---
Refers to: .travis.yml:6 in eb80b95. [](commit_id = eb80b955a59d2f434737e707a900ee2ebcd8dae6, deletion_comment = False)

.travis.yml
rayluo: No, we don't want to remove any of the previous versions. We want the test runs to cover all major Python versions, as seen [here](https://travis-ci.org/AzureAD/microsoft-authentication-library-for-python/builds/598305717?utm_source=github_status&utm_medium=notification).

PR #112, opened 2019-10-22
Updating samples to call MS graph

abhidnya13:


sample/confidential_client_certificate_sample.py
```diff
@@ -12,6 +12,8 @@
     "private_key_file": "filename.pem"
         // For information about generating thumbprint and private key file, refer:
         // https://github.com/AzureAD/microsoft-authentication-library-for-python/wiki/Client-Credentials#client-credentials-with-certificate
+
+    "endpoint": "https://graph.microsoft.com/v1.0/users"
```
jmprieur: for the scope, let's use `User.ReadBasic.All`

sample/confidential_client_certificate_sample.py
```diff
@@ -53,10 +56,11 @@
...
 if "access_token" in result:
-    print(result["access_token"])
-    print(result["token_type"])
-    print(result["expires_in"])  # You don't normally need to care about this.
-                                 # It will be good for at least 5 minutes.
+    # Calling graph using the access token
+    graph_data = requests.get(  # Use token to call downstream service
+        config["endpoint"],
+        headers={'Authorization': 'Bearer ' + result['access_token']}, ).json()
+    print("Users from graph: " + str(graph_data))
```
rayluo: Nit: Given that we make the `endpoint` configurable in settings, it would seem more appropriate we also use a more generic term here, perhaps something like `print("Graph API call result: %s" % graph_data)`?

sample/confidential_client_secret_sample.py
```diff
@@ -53,10 +56,11 @@
     result = app.acquire_token_for_client(scopes=config["scope"])
 
 if "access_token" in result:
-    print(result["access_token"])
-    print(result["token_type"])
-    print(result["expires_in"])  # You don't normally need to care about this.
-                                 # It will be good for at least 5 minutes.
+    # Calling graph using the access token
+    graph_data = requests.get(  # Use token to call downstream service
+        "https://graph.microsoft.com/v1.0/users",
```
rayluo: Did you mean to use `config["endpoint"]` here, as you did for the other sample?

sample/confidential_client_secret_sample.py
```diff
@@ -53,10 +56,11 @@
...
+        headers={'Authorization': 'Bearer ' + result['access_token']},).json()
+    print("Users from graph: " + str(graph_data))
```
rayluo: Same as above. Perhaps we want `print("Graph API call result: %s" % graph_data)`?

sample/confidential_client_certificate_sample.py
rayluo: A different topic here. Currently this line becomes the new last line in this json setting, it would cause a json loading error when customer (Saeed on stage?) do copy&paste and run it for the first time, because the [previous last setting (now it becomes the second-last setting) contains no trailing comma either](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/112/files#diff-3005f568326ab0f67d4ddfd9c3b0a7feR12).

Suggestion: it is better to add new setting in the middle of a json file, not at the end. :-)

Life Hack: In my own json setting files, I would purposely add an extra last setting as `"placeholder": null` BEFOREHAND, so I won't hit that issue at all. But we probably would not want to add that noise in such public-facing sample. So that is not applicable here.

rayluo: Thanks for the suggestion! @abhidnya13 and I backport some of the relevant descriptions from the stand-alone samples to here too!

PR #129, opened 2019-11-25
Adding support for exposing sub errors

abhidnya13:
This is a draft.

---

Experimenting different flavors of API surface.

1. If the app developer "subscribes" the CA error by a container in parameter, [this is a sample](https://github.com/AzureAD/microsoft-authentication-library-for-python/compare/return-error-in-param?expand=1#diff-350433198f3b1800ace53a8ba389538c) to demonstrate what the differences are, comparing to the pre-CA sample.

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -415,7 +415,7 @@ class ClientApplication(object):
         """
         assert isinstance(scopes, list), "Invalid parameter type"
         self._validate_ssh_cert_input_data(kwargs.get("data", {}))
-        result = None
+        response = None
         if authority:
             warnings.warn("We haven't decided how/if this method will accept authority parameter")
         # the_authority = Authority(
@@ -428,9 +428,10 @@ class ClientApplication(object):
             if result.get("access_token"):
                 return result
             if error_response:
-                if result.get("suberror") in set(["bad_token", "token_expired", "client_mismatch"]):
-                    result = None
-                return Error(result['error'], result['error_description'], result['suberror'])
+                classification = result.get("suberror")
+                if classification:
+                    if classification not in set(["bad_token", "token_expired", "client_mismatch"]):
+                        response = Error(result['error'], result['error_description'], result['suberror'])
         for alias in self._get_authority_aliases(self.authority.instance):
             the_authority = Authority(
                 "https://" + alias + "/" + self.authority.tenant,
@@ -442,11 +443,11 @@ class ClientApplication(object):
                 if result.get("access_token"):
                     return result
                 if error_response:
-                    if result.get("suberror") in set(["bad_token", "token_expired", "client_mismatch"]):
-                        result = None
-                    else:
-                        return Error(result['error_code'], result['error_description'], result['suberror'])
-        return result
+                    classification = result.get("suberror")
+                    if classification:
+                        if classification not in set(["bad_token", "token_expired", "client_mismatch"]):
+                            response = Error(result['error'], result['error_description'], result['suberror'])
+        return response
 
     def _acquire_token_silent_from_cache_and_possibly_refresh_it(
             self,
--- a/msal/application.py
+++ b/msal/application.py
@@ -15,6 +15,7 @@ from .mex import send_request as mex_send_request
 from .wstrust_request import send_request as wst_send_request
 from .wstrust_response import *
 from .token_cache import TokenCache
+from .exceptions import Error
 
 
 # The __init__.py will import this. Not the other way around.
@@ -392,6 +393,7 @@ class ClientApplication(object):
             account,  # type: Optional[Account]
             authority=None,  # See get_authorization_request_url()
             force_refresh=False,  # type: Optional[boolean]
+            error_response= False,  # type: Optional[boolean]
             **kwargs):
         """Acquire an access token for given account, without user interaction.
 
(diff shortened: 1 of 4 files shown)


msal/application.py
```diff
@@ -413,6 +413,7 @@ def acquire_token_silent(
         """
```
rayluo: We will also need to come up with parameter documentation. Something along these lines:

```
        :param error_response:
            A boolean flag for whether returning error response. 
            Default value is False, which will just return None instead of error.

        :return:
            - A dict containing "access_token" key, when cache lookup succeeds.
            - None when cache lookup does not yield anything.
            - A dict containing the error response,
               this will be available only when an error actually happens,
               and when the "error_response" parameter was True.
```

msal/application.py
```diff
@@ -428,9 +428,10 @@ def acquire_token_silent(
             if result.get("access_token"):
                 return result
             if error_response:
-                if result.get("suberror") in set(["bad_token", "token_expired", "client_mismatch"]):
-                    result = None
-                return Error(result['error'], result['error_description'], result['suberror'])
+                classification = result.get("suberror")
+                if classification:
+                    if classification not in set(["bad_token", "token_expired", "client_mismatch"]):
+                        response = Error(result['error'], result['error_description'], result['suberror'])
```
rayluo: I thought in last Tue's discussion, Henrik and us agreed on returning a dictionary as the return type?

sample/username_password_sample_with_classification.py
```diff
@@ -0,0 +1,103 @@
...
+                Launching interactive authentication flow will show a message 
+                explaining the condition.""")
+                sys.exit("Abort without bothering to try acquire_token_by_xyz()")
+            elif option == "consent_required":
+                print("""Condition can be resolved by giving consent 
+                during the interactive authentication flow.""")
+            elif option == "user_password_expired":
+                print("User password expired")
+            else:
+                print("Invoke default error handling routine")
```
rayluo: Although many of this sample is derived from [the other sample of our upcoming A-B test](https://github.com/AzureAD/microsoft-authentication-library-for-python/compare/return-error-in-param?expand=1#diff-350433198f3b1800ace53a8ba389538c), here this sample does not trigger a "default error handling routine" (which would be to call `acquire_token_by_xyz()`). This would not be a meaningful A-B test. Might give it another shot?

msal/application.py
```diff
@@ -431,7 +440,14 @@ def acquire_token_silent(
             result = self._acquire_token_silent_from_cache_and_possibly_refresh_it(
                 scopes, account, the_authority, force_refresh=force_refresh, **kwargs)
             if result:
-                return result
+                if result.get("access_token"):
+                    return result
+                if error_response:
```
rayluo: Can we refactor to avoid repeating the same logic here?

msal/application.py
```diff
@@ -15,6 +15,7 @@
 from .wstrust_request import send_request as wst_send_request
 from .wstrust_response import *
 from .token_cache import TokenCache
+from .exceptions import Error
```
rayluo: We do not need this anymore, do we?

msal/application.py
```diff
@@ -405,6 +406,7 @@ def acquire_token_silent(
             account,  # type: Optional[Account]
             authority=None,  # See get_authorization_request_url()
             force_refresh=False,  # type: Optional[boolean]
+            error_response= False,  # type: Optional[boolean]
```
```suggestion
            error_response=False,  # type: Optional[boolean]
```

sample/username_password_sample_with_classification.py
```diff
@@ -0,0 +1,96 @@
...
+    elif option == "consent_required":
+        print("""Condition can be resolved by giving consent 
+        during the interactive authentication flow.""")
+    elif option == "user_password_expired":
+        print("User password expired")
+    else:
+        print("Default error handling behaviour")
+        result = app.acquire_token_by_username_password(
+            config["username"], config["password"], scopes=config["scope"])
+        print(json.dumps(result, indent=4))
```
rayluo: Shall our sample also demonstrate how to use the returned token to make a graph api call?

tests/test_application.py
```diff
@@ -194,3 +194,37 @@ def test_acquire_token_silent(self):
...
+        })  # The add(...) helper populates correct home_account_id for future searching
+
+    def test_error_response(self):
+        app = ClientApplication(
+            self.client_id,
+            authority=self.authority_url, token_cache=self.cache)
+        def tester(url, data=None, **kwargs):
+            return Mock(status_code=200, json=Mock(return_value={
+                "error": "invalid_grant",
+                "error_description": "Was issued to another client",
```
rayluo: Perhaps we can remove this line? I have not tested it though.

```suggestion
```

abhidnya13: This cant be removed as error description is needed when logger is trying to write debug logs

sample/username_password_sample_with_classification.py
abhidnya13: I can certainly do that. It will however change the structure a little which is fine according to me but if we dont want too many deletions and new additions in the sample, may be we stick to this as calling a graph api is already explained above in the sample. Feel free to suggest another sample structure if you don't want too many additions and deletions.

tests/test_application.py
abhidnya13: Removing this will break the test as the logger is using error description value in the library

rayluo: Good to know. Then we should change our library instead, to remove our reliance on the error_description. I came up with [a PR](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/140) just now, please help review. After that, we can remove this line here.

rayluo:
Thanks for all the team effort from everyone here. All the learnings during the process are super useful and we apply our learning to #143

Issue #130, opened 2019-11-25
[Feature] Support for x-app-name and x-app-ver


henrik-me:
SDK Consistency item: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/865189
Additional details: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/865186

Ensure x-app-name and x-app-ver are sent to the same end-points as other x-client- params.
Let the developer specify the app name and the app version

PR #136, opened 2019-12-10
Adding app_name and app_version headers

abhidnya13:
~Resolves~ Addresses #130 <sub>Edited by Ray: Per the team-wide recent workflow, we have to avoid those magic keywords such as "resolves", in order to NOT close that issue until we actually release a new version containing this change.</sub>

For reference: The pattern MSAL Java uses is : 

```java
PublicClientApplication app = PublicClientApplication
                .builder("client-id")
                .correlationId("correlation-id")
                .applicationName("app-name")
                .applicationVersion("app-version")
                .build();
```

The API this PR will implemented will be:

```python
pca = PublicClientApplication(
    ...,
    app_name="my app",
    app_version="1.2.3",
    )
```

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -155,9 +155,9 @@ class ClientApplication(object):
             "x-client-cpu": "x64" if sys.maxsize > 2 ** 32 else "x86",
         }
         if app_name:
-            default_headers['app-name'] =  app_name
+            default_headers['x-app-name'] =  app_name
         if app_version:
-            default_headers['app-version'] = app_version
+            default_headers['x-app-ver'] = app_version
         default_body = {"client_info": 1}
         if isinstance(client_credential, dict):
             assert ("private_key" in client_credential
--- a/msal/application.py
+++ b/msal/application.py
@@ -131,6 +131,14 @@ class ClientApplication(object):
             It will be passed to the
             `timeout parameter in the underlying requests library
             <http://docs.python-requests.org/en/v2.9.1/user/advanced/#timeouts>`_
+        :param app_name: (optional)
+            You can provide your application name for telemetry purposes which will be
+            passed as an additional header value "x-app-name"
+            By default, the additional header value will not be passed
+        :param app_version: (optional)
+            You can provide your application version for telemetry purposes which will be
+            passed as an additional header value "x-app-ver"
+            By default, the additional header value will not be passed
         """
         self.client_id = client_id
         self.client_credential = client_credential
@@ -155,7 +163,7 @@ class ClientApplication(object):
             "x-client-cpu": "x64" if sys.maxsize > 2 ** 32 else "x86",
         }
         if app_name:
-            default_headers['x-app-name'] =  app_name
+            default_headers['x-app-name'] = app_name
         if app_version:
             default_headers['x-app-ver'] = app_version
         default_body = {"client_info": 1}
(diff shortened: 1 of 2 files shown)


msal/application.py
```diff
@@ -131,6 +131,14 @@ def __init__(
             It will be passed to the
             `timeout parameter in the underlying requests library
             <http://docs.python-requests.org/en/v2.9.1/user/advanced/#timeouts>`_
+        :param app_name: (optional)
+            You can provide your application name for telemetry purposes which will be
+            passed as an additional header value "x-app-name"
+            By default, the additional header value will not be passed
+        :param app_version: (optional)
```
rayluo: 1. Given that this feature requirement is an SDK consistency work item, shall we do a little research on a couple other MSAL's interface on this? We can paste their API in this PR's description as a comparison.

2. In particular, given that the underlying wire parameter happens to be named `x-app-ver`, is this parameter named `app_version` or `app_ver` in other MSALs? Want to double check before our merge and release, because we won't be able to easily change a parameter's name without a breaking change.

3. To a lesser extent, how is this feature documented in other MSALs' inline API doc? These 2 parameters need app developer's opt-in, and we currently advertise them as "for telemetry purposes", but app developers might not immediately realize that they don't actually have access to those telemetry data. Perhaps we can somehow reword this to say something like:

            You can provide your application version for Microsoft telemetry purposes.
            Default value is None, means it will not be passed to Microsoft.

msal/application.py
```diff
@@ -143,12 +151,21 @@ def __init__(
                 validate_authority, verify=verify, proxies=proxies, timeout=timeout)
             # Here the self.authority is not the same type as authority in input
         self.token_cache = token_cache or TokenCache()
-        self.client = self._build_client(client_credential, self.authority)
+        self.client = self._build_client(client_credential, self.authority, app_name, app_version)
```
rayluo: A good practice of calling optional parameter is to always use named notation.

```python
    self.client = self._build_client(
        client_credential, self.authority,
        app_name=app_name, app_version=app_version)
```

msal/application.py
```diff
@@ -143,12 +151,21 @@ def __init__(
...
         self.authority_groups = None
 
-    def _build_client(self, client_credential, authority):
+    def _build_client(self, client_credential, authority, app_name=None, app_version=None):
```
rayluo: There are 2 occurrences of calling `_build_client(...)` and we need to also change another one. (And then we might not choose to change these parameters at all.)

rayluo: There are 2 occurrences of calling `_build_client(...)` and we currently only changed one. We will also need to change another. (And then you might choose to not adding two new parameters here.)

msal/application.py
abhidnya13: Thanks @rayluo , All these are really good points to note.

1. I have added the MSAL Java Interface in the description. 
2. The Java implementation uses applicationName and applicationVersion. So I think its just more informative to use it as `version` instead of just `ver`.
3. The exact documentation for Java is this:
```
/**
         * Sets application version for telemetry purposes
         * 
         * @param val application version
         * @return instance of the Builder on which method was called
         */
```
But I think your point is valid and we can mention it explicitly that its for Microsoft telemetry purposes. But then if I understand it correct, they can actually use it in their own telemetry usecases too if they want to and are logging these requests on their side. Let me know your thoughts.

PR #143, opened 2020-01-15
New method acquire_token_silent_with_error()

rayluo:
In an effort to address #57 , we need to expose some of the underlying token refreshing errors to app developer. After weighing different options (such as extending the existing `acquire_token_silent()` method with different style of opt-in flag: boolean, container, callback, ..., you name it), we chose to keep the original `acquire_token_silent()`'s behavior unchanged, and introduce a new `acquire_token_silent_with_error()` method.

The online documentation for both the old `acquire_token_silent()` and new `acquire_token_silent_with_error()` has also been updated and staged [here](https://msal-python.readthedocs.io/en/acquire_token_silent_with_error/#msal.ClientApplication.acquire_token_silent). Please also proofread.

Besides, quoted below is the upcoming Wiki page on how to use the new API to handle CA error. Please also proofread.

PS: Since @abhidnya13 is OOF, @navyasric please help take a look.
//CC @henrik-me 

---

To capture possible Conditional Access (CA) error, you can modify your app in this way.

Before:

```python
result = app.acquire_token_silent(your_scopes, account=your_account)
if not result:
    logging.info("No suitable token exists in cache. Let's get a new one from AAD.")
    result = app.acquire_token_by_xyz_method(...)

if "access_token" in result:
    use_the_token(result["access_token"])
else:
    print(result.get("error"))
```

After:

```python
# Modification begins
result = app.acquire_token_silent_with_error(your_scopes, account=your_account)
successful = result and "error" not in result
if not successful:
    if result:  # CA error occurred. You can handle them specifically.
        print("There was a refresh token in cache, but it was rejected.")
        if result.get("classification") == "basic_action":
            print("""Condition can be resolved by user interaction
                during the interactive authentication flow.""")
        elif result.get("classification") == "additional_action":
            print("""Condition can be resolved by additional remedial interaction
                with the system, outside of the interactive authentication flow.""")
        elif result.get("classification") == "message_only":
            print("""Condition cannot be resolved at this time.
                Launching interactive authentication flow will NOT help,
                it would only show a message explaining the condition.""")
            sys.exit("Abort without bothering to try acquire_token_by_xyz()")
        else:
            print("Invoke default error handling routine")
# Modification ends

# The rest remains the same
    logging.info("No suitable token exists in cache. Let's get a new one from AAD.")
    result = app.acquire_token_by_xyz_method(...)

if "access_token" in result:
    use_the_token(result["access_token"])
else:
    print(result.get("error"))
```

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -431,9 +431,39 @@ class ClientApplication(object):
             **kwargs):
         """Acquire an access token for given account, without user interaction.
 
+        It behaves same as :func:`~acquire_token_silent_with_error`,
+        except that this method will combine the cache empty and refresh error
+        into one return value, `None`.
+        If your app does not need to care the exact token refresh error during
+        token cache look-up, then this method is easier to use.
+
+        Internally, this method calls :func:`~acquire_token_silent_with_error`.
+
+        :return:
+            - A dict containing no "error" key,
+              and typically contains an "access_token" key,
+              if cache lookup succeeded.
+            - None when cache lookup does not yield a token.
+        """
+        result = self.acquire_token_silent_with_error(
+            scopes, account, authority, force_refresh, **kwargs)
+        return result if result and "error" not in result else None
+
+    def acquire_token_silent_with_error(
+            self,
+            scopes,  # type: List[str]
+            account,  # type: Optional[Account]
+            authority=None,  # See get_authorization_request_url()
+            force_refresh=False,  # type: Optional[boolean]
+            **kwargs):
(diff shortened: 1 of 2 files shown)


rayluo:
By the way, pasting a quote from Azure SDK team's @chlowell here:

> [11:43 AM] Charles Lowell
    Thanks for the heads up. I don't think I have a use for it yet, and provided acquire_token_silent behaves the same, it won't break any of my code.
(1 liked)

PR #144, opened 2020-01-16
Add test for bug in TokenCache.Add

rlrossiter:
There is a bug in `TokenCache.Add()` that causes a collision for Access Tokens and SSH certificates. Because the key field for an access token and SSH certificates are generated identically, acquiring an existing SSH certificate with the same scope, authority, etc. as an access token that is already in the cache, the SSH certificate will override the value of the AT in the cache.

This is also a problem with `TokenCache.Find()` because the query parameters used to find an AT in the cache is a subset of the values that are set for an SSH certificate. So doing a lookup for an AT may return the result of an SSH token (which breaks adding it to a `Bearer` header), and vice-versa.

### DIFF
--- a/tests/test_token_cache.py
+++ b/tests/test_token_cache.py
@@ -222,6 +222,38 @@ class TokenCacheTestCase(unittest.TestCase):
             {}).get("key_id")
         self.assertEqual(my_key_id, cached_key_id, "AT should be bound to the key")
 
+    def test_add_access_token_and_ssh_certificate(self):
+        # test adding both an access token and SSH certificate into the
+        # cache with the same values
+
+        # add an access token
+        self.cache.add({
+            "client_id": "my_client_id",
+            "scope": ["s1"],
+            "token_endpoint": "https://login.example.com/contoso/v2/token",
+            "response": self.build_response(
+                uid="uid", utid="utid",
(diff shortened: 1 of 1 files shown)


rlrossiter:
Actually, it appears that doing `add()` with 2 different SSH certs will override each other as well (though the lookup will be OK because `key_id` is taken into consideration when querying).

rayluo:
Thanks @rlrossiter for even coming up with a test case and then use this PR to demonstrate his finding! A very efficient and creative method of issue reporting. :-)

We discussed off-line. The observation above is correct. However those are known limitations based on existing token cache schema, because the current schema does not key a token by its `key_id`. In practice, this does not affect the desired functionality, because:

1. The scope. As usual, the scope used by SSH cert feature (currently `https://pas.windows.net/CheckMyAccess/Linux/user_impersonation`) will not be shared by any other service. Therefore tokens for different services will naturally not collide.
2. The key_id. When acquiring ssh-cert token, if a new key_id is provided, MSAL will correctly bypass the existing ssh-cert bound to the previous key. Typically, Azure CLI already has a login RT that can be used to silently acquire a new SSH-cert, so end user experience remains uninterrupted.

We document the conclusion as above, and close this issue here. Thanks Ryan again for his effort!

PR #150, opened 2020-02-04
take openid as parameter

nwalsh1995:
Adds openid as a parameter to allow for async authorities later.

Idea for async authority: Async authority will have a `async` `classmethod` that returns the `openid_config` and then it will use the default initializer of this class to pass it in. When implementing the async version, logic will be shared between the sync/async for if `openid_config is None`

### DIFF
--- a/msal/authority.py
+++ b/msal/authority.py
@@ -34,7 +34,7 @@ class Authority(object):
     _domains_without_user_realm_discovery = set([])
 
     def __init__(self, authority_url, validate_authority=True,
-            verify=True, proxies=None, timeout=None,
+            verify=True, proxies=None, timeout=None, openid_config=None,
             ):
         """Creates an authority instance, and also validates it.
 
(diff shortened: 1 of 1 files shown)


msal/authority.py
```diff
@@ -34,7 +34,7 @@ class Authority(object):
     _domains_without_user_realm_discovery = set([])
 
     def __init__(self, authority_url, validate_authority=True,
-            verify=True, proxies=None, timeout=None,
+            verify=True, proxies=None, timeout=None, openid_config=None,
```
rayluo: This implementation achieves the goal defined in your PR description:

> Adds openid as a parameter to allow for async authorities later.

Look forward to see a prototype for this, either as new commits being appended to this PR, or in subsequent PR(s).

> Idea for async authority: Async authority will have a async classmethod that returns the openid_config and then it will use the default initializer of this class to pass it in. When implementing the async version, logic will be shared between the sync/async for if openid_config is None

nwalsh1995: We will implement the authority in subsequent PR. I believe async code can live in this repository as long as any Python 2 flow does not import any async code. Most likely we will create a separate module for it.

PR #151, opened 2020-02-04
create new methods for sending requests

nwalsh1995:
Moves the requests specific logic into isolated methods that can be overriden by subclasses.

Async support idea:
Async clients will override these methods along with the sending method to add `await`. Shared logic (building requests) can be shared amongst async/sync clients.


rayluo:
Thanks for your effort. This feature turned out to be much broader, which we recently addressed in #169.

PR #152, opened 2020-02-04
allow injection of authority/client and remove direct calls to transport

nwalsh1995:
Since authority/client by default will create something with a sync client, we need to be able to pass our own classes to these to support async/other transports.

Also moved coupled function calls with requests library into new methods that can be overridden by subclasses.


rayluo:
Thanks for your effort. This feature turned out to be much broader, which we recently addressed in #169.

PR #163, opened 2020-02-27
Switching to Lab App for OBO testing

abhidnya13:


tests/test_e2e.py
```diff
@@ -412,22 +412,21 @@ def test_adfs2019_onprem_acquire_token_by_auth_code(self):
         self.assertCacheWorksForUser(result, scopes, username=None)
 
     @unittest.skipUnless(
-        os.getenv("OBO_CLIENT_SECRET"),
-        "Need OBO_CLIENT_SECRET from https://buildautomation.vault.azure.net/secrets/IdentityDivisionDotNetOBOServiceSecret")
+        os.getenv("LAB_OBO_CLIENT_SECRET"), "OBO app secret not found")
     def test_acquire_token_obo(self):
         # Some hardcoded, pre-defined settings
-        obo_client_id = "23c64cd8-21e4-41dd-9756-ab9e2c23f58c"
```
rayluo: Per our earlier discussion, have we investigated why our previous test setup suddenly failed in the first place? Most of those previous settings were obtained from [MSAL .Net's OBO test case](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blame/4.8.2/tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/OboTests.cs), which seems to be still working?

tests/test_e2e.py
```diff
@@ -412,22 +412,21 @@ def test_adfs2019_onprem_acquire_token_by_auth_code(self):
         self.assertCacheWorksForUser(result, scopes, username=None)
 
     @unittest.skipUnless(
-        os.getenv("OBO_CLIENT_SECRET"),
-        "Need OBO_CLIENT_SECRET from https://buildautomation.vault.azure.net/secrets/IdentityDivisionDotNetOBOServiceSecret")
+        os.getenv("LAB_OBO_CLIENT_SECRET"), "OBO app secret not found")
```
rayluo: It is OK to switch to a different set of test setup, but then I would suggest to follow the previous implementation, to leave some hint for future troubleshooting (in case the test case would somehow fail again), and/or for auditing (i.e. to understand where those test settings come from).

tests/test_e2e.py
abhidnya13: MSAL .Nets OBO test case is an app set up in Travis's personal tenant. He had preconfigured the apps to work with the lab account we use. There was a change in the cloud accounts returned by lab api recently. The account we end up using to run these tests was not consented on Travis's app. While we can grant this consent for the first app using interactive auth, we will have to rely on Travis to add consent to this account for the second confidential client app in this test case. 
I confirmed with Travis that .NET test cases are failing too because of this change.
So, the best thing to do is move to using the accounts provided by lab so that they take care of this consent internally when they add new accounts that are returned by their API.
Aside, I am talking to Gladwin about how these API changes can be communicated so that we dont break the tests.

PR #165, opened 2020-02-28
Arlington automation

abhidnya13:
E2E Test Scenarios added:
- Acquire token by username password
- Acquire token by client_secret
- Acquire token by on behalf of flow
- Acquire token by device code flow

Besides, acquire token silent is tested in each of the flows individually (except acquire token by client secret)

### DIFF
--- a/tests/test_e2e.py
+++ b/tests/test_e2e.py
@@ -102,7 +102,7 @@ class E2eTestCase(unittest.TestCase):
             username=username if ".b2clogin.com" not in authority else None,
             )
 
-    def _test_device_code(
+    def _test_device_flow(
             self, client_id=None, authority=None, scope=None, **ignored):
         assert client_id and authority and scope
         self.app = msal.PublicClientApplication(
@@ -270,7 +270,7 @@ class DeviceFlowTestCase(E2eTestCase):  # A leaf class so it will be run only on
             cls.config = json.load(f)
 
     def test_device_flow(self):
-        self._test_device_code(**self.config)
+        self._test_device_flow(**self.config)
 
 
 def get_lab_app(
@@ -327,7 +327,7 @@ class LabBasedTestCase(E2eTestCase):
         cls.session.close()
 
     @classmethod
-    def get_lab_app_object(cls, **query): # https://msidlab.com/swagger/index.html
+    def get_lab_app_object(cls, **query):  # https://msidlab.com/swagger/index.html
         url = "https://msidlab.com/api/app"
         resp = cls.session.get(url, params=query)
         return resp.json()[0]
@@ -349,10 +349,10 @@ class LabBasedTestCase(E2eTestCase):
         _env = query.get("azureenvironment", "").lower()
         authority_base = {
             "azureusgovernment": "https://login.microsoftonline.us/"
-        }.get(_env, "https://login.microsoftonline.com/")
+            }.get(_env, "https://login.microsoftonline.com/")
         scope = {
             "azureusgovernment": ["https://graph.microsoft.us/.default"],
-        }.get(_env, ["https://graph.microsoft.com/.default"])
+            }.get(_env, ["https://graph.microsoft.com/.default"])
         return {  # Mapping lab API response to our simplified configuration format
             "authority": authority_base + result["tenantID"],
             "client_id": result["appId"],
@@ -366,7 +366,7 @@ class LabBasedTestCase(E2eTestCase):
             **ignored):
         assert client_id and authority and port and scope
         (self.app, ac, redirect_uri) = _get_app_and_auth_code(
-            client_id, authority, port, scope)
+            client_id, authority=authority, port=port, scopes=scope)
         result = self.app.acquire_token_by_authorization_code(
             ac, scope, redirect_uri=redirect_uri)
         logger.debug(
@@ -516,17 +516,20 @@ class WorldWideTestCase(LabBasedTestCase):
                 # This won't work https://msidlab.com/api/user?usertype=b2c
             password="***"  # From https://aka.ms/GetLabUserSecret?Secret=msidlabb2c
         """
-        config = {"authority": self._build_b2c_authority("B2C_1_SignInPolicy"),
-                  "client_id": "b876a048-55a5-4fc5-9403-f5d90cb1c852",
-                  "scope": ["https://msidlabb2c.onmicrosoft.com/msaapp/user_impersonation"], "port": 3843}
-        self._test_acquire_token_by_auth_code(**config)
+        self._test_acquire_token_by_auth_code(
+            authority=self._build_b2c_authority("B2C_1_SignInPolicy"),
+            client_id="b876a048-55a5-4fc5-9403-f5d90cb1c852", port=3843,
+            scope=["https://msidlabb2c.onmicrosoft.com/msaapp/user_impersonation"]
+            )
 
     def test_b2c_acquire_token_by_ropc(self):
-        config = {"authority": self._build_b2c_authority("B2C_1_ROPC_Auth"),
-                  "client_id": "e3b9ad76-9763-4827-b088-80c7a7888f79",
-                  "username": "b2clocal@msidlabb2c.onmicrosoft.com", "password": self.get_lab_user_secret("msidlabb2c"),
-                  "scope": ["https://msidlabb2c.onmicrosoft.com/msidlabb2capi/read"]}
-        self._test_username_password(**config)
+        self._test_username_password(
+            authority=self._build_b2c_authority("B2C_1_ROPC_Auth"),
+            client_id="e3b9ad76-9763-4827-b088-80c7a7888f79",
+            username="b2clocal@msidlabb2c.onmicrosoft.com",
+            password=self.get_lab_user_secret("msidlabb2c"),
+            scope=["https://msidlabb2c.onmicrosoft.com/msidlabb2capi/read"]
+            )
 
 
 class ArlingtonCloudTestCase(LabBasedTestCase):
@@ -556,10 +559,10 @@ class ArlingtonCloudTestCase(LabBasedTestCase):
 
         self._test_acquire_token_obo(config_pca, config_cca)
 
-    def test_arlington_acquire_token_device_code(self):
+    def test_arlington_acquire_token_device_flow(self):
         config = self.get_lab_user(usertype="cloud", azureenvironment=self.environment, publicClient="yes")
         config["scope"] = ["user.read"]
-        self._test_device_code(**config)
+        self._test_device_flow(**config)
 
 
 if __name__ == "__main__":
--- a/tests/test_e2e.py
+++ b/tests/test_e2e.py
@@ -518,7 +518,8 @@ class WorldWideTestCase(LabBasedTestCase):
         """
         self._test_acquire_token_by_auth_code(
             authority=self._build_b2c_authority("B2C_1_SignInPolicy"),
-            client_id="b876a048-55a5-4fc5-9403-f5d90cb1c852", port=3843,
+            client_id="b876a048-55a5-4fc5-9403-f5d90cb1c852",
+            port=3843,  # Lab defines 4 of them: [3843, 4584, 4843, 60000]
             scope=["https://msidlabb2c.onmicrosoft.com/msaapp/user_impersonation"]
             )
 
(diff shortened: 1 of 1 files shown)


abhidnya13:
cc: @henrik-me

tests/test_e2e.py
```diff
@@ -322,6 +322,12 @@ def setUpClass(cls):
     def tearDownClass(cls):
         cls.session.close()
 
+    @classmethod
+    def get_lab_app_object(cls, **query ):
```
rayluo: Does this API also have an online document? If so, let's leave its URL as comment, pretty much [like what we did here](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/165/files#diff-2f43f6151484c4ac714c0be06ac04a7cR342). That would be the only way to let the future maintainer know what those `**query` would be.

tests/test_e2e.py
```diff
@@ -505,6 +515,93 @@ def test_b2c_acquire_token_by_ropc(self):
             scope=["https://msidlabb2c.onmicrosoft.com/msidlabb2capi/read"],
             )
 
+    def test_arlington_acquire_token_by_ropc(self):
```
rayluo: Let's explore whether we could reuse most test logic among different clouds. These steps might work:

1. Change all the `def test_blah()` methods in this test class to `def _test_blah()` so that they won't run. This class would contain common test logic.
2. Create 2 sub-classes, such as `class WorldWideTestCase(LabBasedTestCase)` and `class ArlingtonTestCase(LabBasedTestCase)`, capture their difference in configuration. Define `def test_blah()` in these sub-classes to call their parent methods.

tests/test_e2e.py
```diff
@@ -505,6 +515,93 @@ def test_b2c_acquire_token_by_ropc(self):
...
+        config = self.get_lab_user(azureenvironment="azureusgovernment")
+        self._test_username_password(
+            password=self.get_lab_user_secret(config["lab_name"]), **config)
+
+    def test_arlington_acquire_token_by_client_secret(self):
```
rayluo: Historically, there was no explicit test case to cover confidential client in world-wide (WW) cloud. (We implicitly tested it by creating a confidential client as test infrastructure, though.)
That's why I LOVE seeing you add an explicit test case for confidential client this time! Let's see if we can follow the suggestion above, to try to backport this test case for WW cloud, in a reusable style.

tests/test_e2e.py
abhidnya13: I have separated this as per your suggestion. While doing that, I realized we can separate B2C and ADFS scenarios as well into their own subclasses as technically they are not using worldwide authorities. Thoughts ?

tests/test_e2e.py
abhidnya13: Thats good to have.  I think that should go in a separate PR though once this is merged. Lets keep the scope of this PR to Arlington test cases ? Let me know if that works for you. If yes, I can go and create an issue tracking this.

henrik-me: If not doing this now, then please create a github issue to refactor test cases for multiple cloud enablement. You can point to this comment. Expect that I will be asking for the same automation to run in all the clouds we support and that code should be well structured to handle that (similar discussion happening in other repo's at this time)

tests/test_e2e.py
henrik-me: (pls. see other comment) : pls. ensure you anticipate multiple clouds to be automated and that it should relatively easy to chose which ones to run, and e.g run all in a release build but only run one, two or a random selection for PR builds.

tests/test_e2e.py
```diff
@@ -336,67 +342,31 @@ def get_lab_user_secret(cls, lab_name="msidlab4"):
     def get_lab_user(cls, **query):  # https://docs.msidlab.com/labapi/userapi.html
         resp = cls.session.get("https://msidlab.com/api/user", params=query)
         result = resp.json()[0]
+        authority_base = "https://login.microsoftonline.com/"
+        graph_endpoint = "https://graph.microsoft.com/.default"
+        if "azureenvironment" in query and query["azureenvironment"] == "azureusgovernment":
+            authority_base = "https://login.microsoftonline.us/"
+            graph_endpoint = "https://graph.microsoft.us/.default"
```
rayluo, line 349: Ideally we should avoid "assign a default value; and then change it immediately" pattern. The following approach is more bullet-proof, and future-proof.

```suggestion
        _env = query.get("azureenvironment", "").lower()
        base = {
            "azureusgovernment": "https://login.microsoftonline.us",
            }.get(_env, "https://login.microsoftonline.com")
        scopes = {
            "azureusgovernment": ["https://graph.microsoft.us/.default"],
            }.get(_env, ["https://graph.microsoft.com/.default"])
```

You would also need to adjust the rest of this function accordingly.

tests/test_e2e.py
rayluo: To achieve Henrik's suggestion, let's rename the current test class from `SovereignCloudTestCase` to `ArlingtonCloudTestCase`, and define a property `environment = "azureusgovernment"`. All its test methods will reference such environment property. This way, in the future it would be very easy to expand to cover different clouds.

tests/test_e2e.py
```diff
@@ -336,67 +342,31 @@ def get_lab_user_secret(cls, lab_name="msidlab4"):
...
-            port=8080,
-            scopes=scopes,
+            config["client_id"],
+            authority=config["authority"],
+            port=config["port"],
+            scopes=config["scope"],
             )
         result = self.app.acquire_token_by_authorization_code(
-            ac, scopes, redirect_uri=redirect_uri)
+            ac, config["scopes"], redirect_uri=redirect_uri)
```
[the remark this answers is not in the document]
henrik-me: sometimes there is a reference to the passed in scope and sometimes it's a reference to the scopes from the config. Should this stick to only being be params passed in? Then config can be used for calling the functions?

---
In reply to: [401275819](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/165#discussion_r401275819) [](ancestors = 401275819)

henrik-me:
I would have expected to see something like this:

test_allenvironments(config)
{
  foreach (envriroment in config.TestEnvironments)
    test_envrionment(envrironment, config)
}

test_environment(environment, config)
{
  test_devicecode(environment, ...)
  test_xyz(environment, ...)
}

test_devicecode(environment, ....) {}
test_xyz(environment) {}


Are there many differences between the environments making that not easy to do? Am I reading the code in the wrong way?

henrik-me:
There seems to be several references to client id's etc. in the tests. Any reason the lab API is not used to get those as well as the secrets?

abhidnya13:
Hey @henrik-me,

1. The code structured right now is implicitly doing a similar thing as you have mentioned. By default it is running through each environment and if we want to test a specific environment, we can run the test cases under that class. We cannot have a single config for an environment as we have multiple different users as a part of each environment right now, for example- usertype: federated, onprem, cloud version(for a federated user): adfsv2, v3, v4 etc. So, the queries to pull up config will have these values along with the environment. The base methods that these calls for different types of flows within each environment are defined in a parent class and then depending on the kind of user in that environment we provide a config to these base methods.
2. Thats a good point. For the new tests I've added for the new Arlington test cases, I have pulled the client_ids from the Lab API. There are a few test cases which were already present that don't do this and it will be good to get those from LabAPI as well and we should be able to do it. Created an issue to track this [here](https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/180)

tests/test_e2e.py
```diff
@@ -472,38 +516,51 @@ def test_b2c_acquire_token_by_auth_code(self):
...
 
     def test_b2c_acquire_token_by_ropc(self):
-        self._test_username_password(
-            authority=self._build_b2c_authority("B2C_1_ROPC_Auth"),
-            client_id="e3b9ad76-9763-4827-b088-80c7a7888f79",
-            username="b2clocal@msidlabb2c.onmicrosoft.com",
-            password=self.get_lab_user_secret("msidlabb2c"),
-            scope=["https://msidlabb2c.onmicrosoft.com/msidlabb2capi/read"],
-            )
+        config = {"authority": self._build_b2c_authority("B2C_1_ROPC_Auth"),
```
rayluo, line 525: Given that the underlying `_test_username_password(...)` already accepts keyword arguments, don't you think the pre-existing calling pattern is actually more readable? Would suggest to revert changes to this test case. PS: The `**config` notation was only a "workaround" when the config was constructed by other helpers.

tests/test_e2e.py
```diff
@@ -472,38 +516,51 @@ def test_b2c_acquire_token_by_auth_code(self):
...
-            json.dumps(result.get("id_token_claims"), indent=4),
-            )
-        self.assertIn(
-            "access_token", result,
-            "{error}: {error_description}".format(
-                # Note: No interpolation here, cause error won't always present
-                error=result.get("error"),
-                error_description=result.get("error_description")))
-        self.assertCacheWorksForUser(result, scopes, username=None)
+        config = {"authority": self._build_b2c_authority("B2C_1_SignInPolicy"),
```
rayluo, line 519: Same suggestion to comment on [line 525](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/165/files#diff-2f43f6151484c4ac714c0be06ac04a7cR525).

tests/test_e2e.py
```diff
@@ -322,6 +322,12 @@ def setUpClass(cls):
     def tearDownClass(cls):
         cls.session.close()
 
+    @classmethod
+    def get_lab_app_object(cls, **query ): # https://msidlab.com/swagger/index.html
```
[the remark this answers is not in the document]
rayluo: The recent commit fixed the first half of the suggestion. The second half was specifically called out by the PEP 8 link above, quoted below:

> Inline comments should be separated by at least *two spaces* from the statement.

PS: If you mean to accept the previous suggestion but were afraid that "github suggestion" behavior would impact your local branch (thus chose to do it manually), you can try it out next time. After accepting a suggestion via github web UI, you would just need to do a "git pull" on your local desktop to bring in the new changes, generally without any merge conflict. :-)

tests/test_e2e.py
rayluo: Thanks for renaming this test class to `ArlingtonCloudTestCase`! Now we just need to remove all the "arlington" keyword from all its methods, so that in future we can easily copy-and-paste those test methods into other cloud, without needing to (remember to) change their names.

tests/test_e2e.py
```diff
@@ -472,39 +516,55 @@ def test_b2c_acquire_token_by_auth_code(self):
...
-        self.assertCacheWorksForUser(result, scopes, username=None)
 
     def test_b2c_acquire_token_by_ropc(self):
         self._test_username_password(
             authority=self._build_b2c_authority("B2C_1_ROPC_Auth"),
             client_id="e3b9ad76-9763-4827-b088-80c7a7888f79",
             username="b2clocal@msidlabb2c.onmicrosoft.com",
             password=self.get_lab_user_secret("msidlabb2c"),
-            scope=["https://msidlabb2c.onmicrosoft.com/msidlabb2capi/read"],
+            scope=["https://msidlabb2c.onmicrosoft.com/msidlabb2capi/read"]
```
rayluo, line 531:
```suggestion
            scope=["https://msidlabb2c.onmicrosoft.com/msidlabb2capi/read"],
```

Reasoning: [PEP 8 on Trailing commas](https://www.python.org/dev/peps/pep-0008/#when-to-use-trailing-commas)

---

PS: It is not necessarily that I paid attention to this kind of details. It is the github PR review UI highlighting the different lines in red-and-green in order to grab reviewer's attention. So, a universal hint for a PR author is to conduct a self-review on github PR UI to see whether each of those red-and-green snippets is intended and/or necessary.

tests/test_e2e.py
```diff
@@ -472,39 +516,55 @@ def test_b2c_acquire_token_by_auth_code(self):
...
         """
-        scopes = ["https://msidlabb2c.onmicrosoft.com/msaapp/user_impersonation"]
-        (self.app, ac, redirect_uri) = _get_app_and_auth_code(
-            "b876a048-55a5-4fc5-9403-f5d90cb1c852",
-            client_secret=self.get_lab_user_secret("MSIDLABB2C-MSAapp-AppSecret"),
+        self._test_acquire_token_by_auth_code(
             authority=self._build_b2c_authority("B2C_1_SignInPolicy"),
-            port=3843,  # Lab defines 4 of them: [3843, 4584, 4843, 60000]
-            scopes=scopes,
+            client_id="b876a048-55a5-4fc5-9403-f5d90cb1c852", port=3843,
```
rayluo, line 521:
```suggestion
            client_id="b876a048-55a5-4fc5-9403-f5d90cb1c852",
            port=3843,  # Lab defines 4 of them: [3843, 4584, 4843, 60000]
```

Suggests to keep the original inline comment. Back then, it took us a while to find those info from the doc ocean. And, if without such "justification", it would be weird to understand where that magic number "3843" came from.

PR #168, opened 2020-03-03
allow passing more parameters when generating auth uri

nwalsh1995:
We are interested in adding some more parameters to the `get_authorization_request_url`. Currently, it appears we can only send the parameters explicitly stated in the `get_authorization_request_url` method, while there are more [possible parameters](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest) than what is currently accepted.

Since the `client.build_auth_request_uri` already supports `kwargs`, we can pass kwargs to that function as well.

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -247,6 +247,7 @@ class ClientApplication(object):
             You will have to specify a value explicitly.
             Its valid values are defined in Open ID Connect specs
             https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest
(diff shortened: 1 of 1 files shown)


rayluo:
Discussed with @nwalsh1995 offline. This PR was mainly created to bring support to `nonce` parameter. So we developed another PR dedicated to `nonce`, and provide more feature there (including to validate the `nonce`).

Closing this PR here, and still thank @nwalsh1995 a lot for making this happen!

PR #173, opened 2020-03-18
Specifying and validating nonce in auth code flow

rayluo:
The `nonce` is a behavior [defined in Open ID Connect](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest).

This PR adds a `nonce` parameter into the `get_authorization_request_url(..., nonce=...)` so that it would be sent to Azure AD.

This PR also adds a `nonce` parameter into the `acquire_token_by_authorization_code(..., nonce=...)` so that this method will perform the nonce check on the returned id token, automatically. We [test it here](https://github.com/AzureAD/microsoft-authentication-library-for-python/compare/nonce-in-msal?expand=1#diff-2f43f6151484c4ac714c0be06ac04a7cR141). By the way, [validating nonce has been embedded in msal .net](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/4.10.0/src/client/Microsoft.Identity.Client/Platforms/iOS/iOSBroker.cs#L191), although it happens implicitly when using public client application.

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -229,6 +229,7 @@ class ClientApplication(object):
             redirect_uri=None,
             response_type="code",  # Can be "token" if you use Implicit Grant
             prompt=None,
+            nonce=None,
             **kwargs):
         """Constructs a URL for you to start a Authorization Code Grant.
 
@@ -247,6 +248,9 @@ class ClientApplication(object):
             You will have to specify a value explicitly.
             Its valid values are defined in Open ID Connect specs
             https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest
+        :param nonce:
+            A cryptographically random value used to mitigate replay attacks. See also
+            `OIDC specs <https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest>`_.
         :return: The authorization url as a string.
         """
         """ # TBD: this would only be meaningful in a new acquire_token_interactive()
@@ -276,6 +280,7 @@ class ClientApplication(object):
             redirect_uri=redirect_uri, state=state, login_hint=login_hint,
             prompt=prompt,
             scope=decorate_scope(scopes, self.client_id),
+            nonce=nonce,
             )
 
     def acquire_token_by_authorization_code(
@@ -286,6 +291,7 @@ class ClientApplication(object):
                 # REQUIRED, if the "redirect_uri" parameter was included in the
                 # authorization request as described in Section 4.1.1, and their
                 # values MUST be identical.
+            nonce=None,
             **kwargs):
         """The second half of the Authorization Code Grant.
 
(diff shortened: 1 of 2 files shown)


henrik-me:
Suggestion, add functionality which combines the two steps into one?

msal/application.py
```diff
@@ -247,6 +248,9 @@ def get_authorization_request_url(
             You will have to specify a value explicitly.
             Its valid values are defined in Open ID Connect specs
             https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest
+        :param nonce:
+            A hard-to-guess string used to mitigate replay attacks. See also
```
henrik-me: >hard-to-guess string [](start = 14, length = 20)

I suggest being a bit more specific here on the min complexity suggested

henrik-me:
>         return self.client.obtain_token_by_authorization_code(

does this already do nonce validation?

---
Refers to: msal/application.py:332 in ea137ae. [](commit_id = ea137ae00e81dfc30505dc5d511cdf4adf7baf64, deletion_comment = False)

msal/application.py
abhidnya13: We can write it as : `unguessable  value (ex : cryptographic hash of a random value)` instead of hard-to-guess as mentioned in the [OIDC Specs](https://openid.net/specs/openid-connect-core-1_0.html#NonceNotes).

tests/test_e2e.py
```diff
@@ -124,20 +124,21 @@ def test_username_password(self):
...
             authority=self.config.get("authority"),
             port=self.config.get("listen_port", 44331),
             scopes=self.config["scope"],
-            )
+            **kwargs)
 
     def test_auth_code(self):
         self.skipUnlessWithConfig(["client_id", "scope"])
-        (self.app, ac, redirect_uri) = self._get_app_and_auth_code()
+        nonce = "foo"
```
abhidnya13: Since nonce is an optional parameter, it makes sense to have a new test case for api calls with nonce parameter. So that our no-nonce test case is still there and we have a new one with nonce :)

msal/application.py
rayluo: Yes, that paragraph was what I referred to as "another full paragraph as implementation notes" in my previous message. Did not use that "cryptographic hash of the [cryptographically random] value" part because, albeit it sounds cool, but a "cryptographically random value" is already unguessable, and its hash does not make it more unguessable. That hash suggestion, in particular, is more for a web server client, yet our MSAL does not solely target web clients, so that part is not always applicable.

Taking everything in this thread into consideration, I think we can combine all our input here, and change that line to:

    A cryptographically random value used to mitigate replay attacks. See also (that link).

I will send out a new commit soon.

tests/test_e2e.py
rayluo: The test case you see here, is a test case for AAD. We still have other similar test cases for B2C and ADFS. Those test cases still operate in a no-nonce way (they may or may not support nonce anyway).

Besides, if our goal is to see whether we (i.e. MSAL) implement that *optional* logic correctly, then this PR itself serves as a white-box test already. We have 3 people reviewing this 19-line PR (and most of them are docs and tests), this is already a record-setting optimal reviewer-per-line ratio. I have high confidence on your reviews. :-)

abhidnya13: I know there are no-nonce test cases for B2C and ADFS but they serve the purpose to test B2C and ADFS. I just think instead of combining this test with our already existing test case, it will be better for debugging later if there is an issue relating this area.

rayluo: Ended up adding a test case for mismatching nonce, and retain the without-nonce test case. So we have 3 test cases now.

PR #181, opened 2020-04-13
Allow domain hint

### LINKED ISSUE #158: How to properly pass domain_hint?
Is there documentation or sample that can clearly show how to pass domain_hint as a query parameter when using MSAL for Python?

abhidnya13:
Allows user to pass in domain hint when generating authorization URL

---
UPDATED (by Ray):
This PR will fix #158. (This line will trigger an automatic close of that issue. This is probably fine, because we will cut a new release shortly after this merge.)

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -255,6 +255,7 @@ class ClientApplication(object):
         :param domain_hint:
             Provides a hint about the tenant or domain that the user should use to sign in. The value
             of the domain hint is a registered domain for the tenant.
+            https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow#request-an-authorization-code
         :return: The authorization url as a string.
         """
         """ # TBD: this would only be meaningful in a new acquire_token_interactive()
--- a/msal/application.py
+++ b/msal/application.py
@@ -230,6 +230,7 @@ class ClientApplication(object):
             response_type="code",  # Can be "token" if you use Implicit Grant
             prompt=None,
             nonce=None,
+            domain_hint=None,  # type: Optional[str]
             **kwargs):
         """Constructs a URL for you to start a Authorization Code Grant.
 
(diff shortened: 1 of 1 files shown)


msal/application.py
```diff
@@ -280,7 +285,7 @@ def get_authorization_request_url(
             redirect_uri=redirect_uri, state=state, login_hint=login_hint,
             prompt=prompt,
             scope=decorate_scope(scopes, self.client_id),
-            nonce=nonce,
+            nonce=nonce, domain_hint=domain_hint,
```
rayluo, line 288:
```suggestion
            nonce=nonce,
            domain_hint=domain_hint,
```
Besides the usual "minimizing the amounts of lines being changed" preference, we could actually make this PR contains all "green lines" and not a single "red line", which would become a practical visual clue to convey "this PR is adding a new feature, without needing to change any existing behavior". :-)

msal/application.py
```diff
@@ -251,6 +252,10 @@ def get_authorization_request_url(
         :param nonce:
             A cryptographically random value used to mitigate replay attacks. See also
             `OIDC specs <https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest>`_.
+        :param domain_hint:
+            Provides a hint about the tenant or domain that the user should use to sign in. The value
+            of the domain hint is a registered domain for the tenant.
```
rayluo, line 257: So, what can be the proper value to be used in this feature? This line suggests "a registered domain for the tenant" which sounds something like "contoso.com" (which is probably correct, based on [the link mentioned here](https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/158#issuecomment-584890191)), but the link at the next line below suggests two (and only two?) valid values as "consumers" or "organizations". Which one is correct?

If all those can be valid values, we can then choose to give a more clear description, something like this:

> Can be one of "consumers" or "organizations" or your tenant domain "contoso.com".
> If included, it will skip the email-based discovery process that user goes through on the sign-in page, leading to a slightly more streamlined user experience.
> And then provide a link (or even two) to existing docs to back it up.

abhidnya13: Yes it can be contoso.com too, I changed the doc string to what you mentioned

PR #185, opened 2020-04-15
Improve SNI docs

rayluo:
Improve SNI docs based on https://github.com/AzureAD/azure-activedirectory-library-for-python/issues/225

CC: @mbearup

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -96,9 +96,9 @@ class ClientApplication(object):
             client_claims=None, app_name=None, app_version=None):
         """Create an instance of application.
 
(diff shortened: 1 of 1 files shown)


msal/application.py
```diff
@@ -97,6 +97,7 @@ def __init__(
         """Create an instance of application.
 
         :param client_id: Your app has a client_id after you register it on AAD.
+
         :param client_credential:
```
mbearup, line 101: Consider a type hint here, e.g. `:param str client_credential:` Tools like pycharm pick up on these hints and make it easier to determine the value to pass in.

PR #186, opened 2020-04-23
Returning refresh token errors which were discarded

### LINKED ISSUE #184: Refresh token errors are discarded
`acquire_token_silent_with_error` appears to be intended to return the last error response received from AAD, and the documentation implies it does so, but there are cases in which it does not.

For example, if AAD rejects a refresh token, its error response is discarded (`at` is the response):
https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/5c1f0f4ad18a61587393c290b1094b5f8fd014d4/msal/application.py#L601-L609

When AAD rejected the token, that token was evicted from the cache:
https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/5c1f0f4ad18a61587393c290b1094b5f8fd014d4/msal/oauth2cli/oauth2.py#L449-L450

Subsequent calls to `_acquire_token_silent_by_finding_specific_refresh_token` may now return `None` because no refresh token remains in the cache. In such a case, `acquire_token_silent_with_error` returns `None`.

This behavior arguably makes sense because the difference between "no refresh token in the cache" and "only invalid refresh tokens in the cache" may not be interesting. But the documentation on `acquire_token_silent_with_error` implies it will distinguish these cases, so I describe the inconsistency here for comment and clarification.

abhidnya13:
Closes #184

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -633,16 +633,20 @@ class ClientApplication(object):
                 **kwargs)
             if at and "error" not in at:
                 return at
+        last_resp = None
         if app_metadata.get("family_id"):  # Meaning this app belongs to this family
             at = self._acquire_token_silent_by_finding_specific_refresh_token(
                 authority, scopes, dict(query, family_id=app_metadata["family_id"]),
                 **kwargs)
             if at and "error" not in at:
                 return at
+            last_resp = at
+
         # Either this app is an orphan, so we will naturally use its own RT;
         # or all attempts above have failed, so we fall back to non-foci behavior.
-        return self._acquire_token_silent_by_finding_specific_refresh_token(
+        resp = self._acquire_token_silent_by_finding_specific_refresh_token(
             authority, scopes, dict(query, client_id=self.client_id), **kwargs)
+        return resp or last_resp
 
     def _get_app_metadata(self, environment):
         apps = self.token_cache.find(  # Use find(), rather than token_cache.get(...)
--- a/msal/application.py
+++ b/msal/application.py
@@ -635,17 +635,16 @@ class ClientApplication(object):
                 return at
         last_resp = None
         if app_metadata.get("family_id"):  # Meaning this app belongs to this family
-            at = self._acquire_token_silent_by_finding_specific_refresh_token(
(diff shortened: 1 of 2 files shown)


msal/application.py
```diff
@@ -633,16 +633,18 @@ def _acquire_token_silent_by_finding_rt_belongs_to_me_or_my_family(
                 **kwargs)
             if at and "error" not in at:
                 return at
+        last_resp = None
         if app_metadata.get("family_id"):  # Meaning this app belongs to this family
-            at = self._acquire_token_silent_by_finding_specific_refresh_token(
+            last_resp = self._acquire_token_silent_by_finding_specific_refresh_token(
```
rayluo, line 638: Conceptually, the original `at` name hints it is (mainly) an AT response. Let's keep that semantics. You can use the following trick (not available in most other languages) to have BOTH `at` and `last_resp`.
```suggestion
            last_resp = at = self._acquire_token_silent_by_finding_specific_refresh_token(
```

msal/application.py
```diff
@@ -633,16 +633,18 @@ def _acquire_token_silent_by_finding_rt_belongs_to_me_or_my_family(
...
-            if at and "error" not in at:
-                return at
+            if last_resp and "error" not in last_resp:
+                return last_resp
         # Either this app is an orphan, so we will naturally use its own RT;
         # or all attempts above have failed, so we fall back to non-foci behavior.
         return self._acquire_token_silent_by_finding_specific_refresh_token(
-            authority, scopes, dict(query, client_id=self.client_id), **kwargs)
+            authority, scopes, dict(query, client_id=self.client_id),
+            **kwargs) or last_resp
```
rayluo, line 647: Yes, this is also how I would implement it. One question though. Have we tested how the new test case is (supposed to) failing if this `or last_resp` part is not present? I briefly tried that just now, but it seems still passing. Did I miss anything?

abhidnya13: I just tested it without `or last_resp` and it fails for me, not sure what are you missing ?

rayluo: My bad. (The things missing is a better pair of glasses, perhaps.)

PR #188, opened 2020-04-30
Improve error message on incorrect authority uri

rayluo:
Today a customer was asking for our support on an error he encountered.

> I have registered my App on Azure Portal but hitting:
> `json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)`
>
> when using:
>
> ```python
> auth = "https://login.microsoftonline.com/common/oauth2/nativeclient"
> app = PublicClientApplication(client_id, authority=auth)
> ```

And he further explained:

> I first went to this [MSAL repo's homepage](https://github.com/AzureAD/microsoft-authentication-library-for-python)
> there seems a bug in README (`config` was not defined but `config["scope"]` was called), so I went to [samples hosted inside same MSAL repo](https://github.com/AzureAD/microsoft-authentication-library-for-python/tree/dev/sample)

Our takeaways:

* Customers expect code snippet in README to be runnable, or at least, with syntactically-correct placeholders which would NOT throw any exception.
* We used to use (overly?) flexible placeholder like "..." in this project's README. Perhaps we need to switch to a narrow example such as "https://login.microsoftonline.com/common" which is suitable to perhaps majority of customers (with the price that it is technically incorrect for B2C or ADFS customers).
* Catch lower level exceptions, and convert them into a meaningful ValueError("Your input  did not work because ...")

This PR addresses such 3 findings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,11 @@ Before using MSAL Python (or any MSAL SDKs, for that matter), you will have to
 [register your application with the Microsoft identity platform](https://docs.microsoft.com/azure/active-directory/develop/quickstart-v2-register-an-app).
 
 Acquiring tokens with MSAL Python follows this 3-step pattern.
+(Note: That is the high level conceptual pattern.
+There will be some variations for different flows. They are demonstrated in
+[runnable samples hosted right in this repo](https://github.com/AzureAD/microsoft-authentication-library-for-python/tree/dev/sample).
+)
+
 
 1. MSAL proposes a clean separation between
    [public client applications, and confidential client applications](https://tools.ietf.org/html/rfc6749#section-2.1).
@@ -43,7 +48,9 @@ Acquiring tokens with MSAL Python follows this 3-step pattern.
 
    ```python
    from msal import PublicClientApplication
-   app = PublicClientApplication("your_client_id", authority="...")
+   app = PublicClientApplication(
+       "your_client_id",
+       "authority": "https://login.microsoftonline.com/Enter_the_Tenant_Name_Here")
    ```
 
    Later, each time you would want an access token, you start by:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -43,7 +48,9 @@ Acquiring tokens with MSAL Python follows this 3-step pattern.
 
    ```python
-   app = PublicClientApplication("your_client_id", authority="...")
+   app = PublicClientApplication(
+       "your_client_id",
+       "authority": "https://login.microsoftonline.com/Enter_the_Tenant_Name_Here")
```
abhidnya13, line 53: Suggestion: tenant-id would be more useful as tenant name can also mean tenant`s actual name  instead of xyz.onmicrosoft.com

rayluo: You might have a point. I don't necessarily know our official terminology here, I was just copying that entire phrase from our existing samples [here](https://github.com/Azure-Samples/ms-identity-python-webapp/blob/master/app_config.py#L12) and [there](https://github.com/Azure-Samples/ms-identity-python-daemon/blob/master/1-Call-MsGraph-WithSecret/parameters.json#L2), which have presumably been vetted by our PMs and doc writers.

I would say let's use this term as-is for now. You/we can bring this up as a separated suggestion, if desirable.

msal/authority.py
```diff
@@ -69,9 +69,16 @@ def __init__(self, authority_url, http_client, validate_authority=True):
                     authority.path,  # In B2C scenario, it is "/tenant/policy"
                     "" if tenant == "adfs" else "/v2.0" # the AAD v2 endpoint
                     ))
-        openid_config = tenant_discovery(
-            tenant_discovery_endpoint,
-            self.http_client)
+        try:
```
abhidnya13, line 72: Any specific reason why we catch it here and not where it originates , in line 132 ? Just curious as I went searching for  where this originates and considered the possibility of having this same try-except block there

rayluo: It is mainly because the input `authority_url` would not be available there.

PR #191, opened 2020-05-12
import_refresh_token() for RT migration

rayluo:
This PR proposes an experimental API for migrating refresh token (RT) from other source (typically ADAL Python, but not necessarily only ADAL Python) to MSAL Python.

TL;DR: This PR proposes an `import_refresh_token(old_rt_string, scope_list)` as first-class citizen (i.e. not hiding from main API). As hinted by its name, its successful return value contains no RT, not even AT, not currently return an ID token (this one is TBD).

```python
app = msal.PublicClientApplication(
    "client_id", authority="https://login.microsoftonline.com/...", token_cache=...)

# NEW API
result = app.import_refresh_token(old_rt, ["scope1", "scope2"])  # <-- NEW API
if "error" in result:
    print("Migration unsuccessful. Error: ", json.dumps(result, indent=2))
else:
    print("Migration is successful")

# From now on, the RT is saved inside MSAL's cache,
# and becomes available in normal MSAL coding pattern. For example:
...
result = app.acquire_token_silent(["scope1"], ...)
```

So that there will be no way to (ab)use such migration API alone as a normal flow. The programming model would be to complete the RT migration upfront, and then come back to the normal MSAL coding model.

Reviewers can focus on its [usage pattern (highlighted at the bottom of this page)](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/191/files#diff-615899d8b1282d9f785b5404cf87e0ccR60-R68), rather than its actual implementation.

Background:

1. To simplify token migration, we do not need to migrate existing access token (AT). We just need to migrate RTs. Because: (1) ATs are short-lived, old ATs might already been expired when the migration happens. (2) Once we have RTs migrated, they will be automatically used to obtain new ATs.
2. To migrate RTs, a shortcut is to use the old RT to redeem an AT, during such process, a new RT will typically also return, and then MSAL's existing token cache mechanism will kick in, and store the new RT and AT properly. This means to call an imaginary API roughly named as `acquire_token_by_refresh_token(...)`.
3. However, historically, ADAL/MSAL has been designed to downplay the RT concept.
   * Many MSALs do not provide an `acquire_token_by_refresh_token(rt, scope, ...)` API. Instead, MSALs provide a higher level API `acquire_token_silent(scope, ...)` which uses RT from internal cache. For example, [MSAL .Net only provides `AcquireTokenByRefreshToken(...)` in an indirect way](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/Adal-to-Msal#adalnet-v2x-migration-to-msalnet-v3x):
      > As this method is intended for scenarios that are not typical, it is not readily accessible with the IConfidentialClientApplication without first casting it to IByRefreshToken.
      > You will see an access token and ID token returned in your AuthenticationResult while your new refresh token is stored in the cache.

     But, even so, an `AcquireTokenByRefreshToken()` is arguably still provided and documented, and it can be used to acquire AT.

   * MSAL Java provides `AcquireTokenByRefreshToken(...)` out-of-the-box. It returns no RT, either, but it returns AT, etc..

### DIFF
--- a/sample/migrate_rt.py
+++ b/sample/migrate_rt.py
@@ -1,0 +1,75 @@
+"""
+The configuration file would look like this:
+
+{
+    "authority": "https://login.microsoftonline.com/organizations",
+    "client_id": "your_client_id",
+    "scope": ["User.ReadBasic.All"],
+        // You can find the other permission names from this document
+        // https://docs.microsoft.com/en-us/graph/permissions-reference
+}
+
+You can then run this sample with a JSON configuration file:
+
+    python sample.py parameters.json
+"""
+
+import sys  # For simplicity, we'll read config file from 1st CLI param sys.argv[1]
+import json
+import logging
+
+import msal
+
+
+# Optional logging
+# logging.basicConfig(level=logging.DEBUG)  # Enable DEBUG log for entire script
+# logging.getLogger("msal").setLevel(logging.INFO)  # Optionally disable MSAL DEBUG logs
+
+config = json.load(open(sys.argv[1]))
+
+def get_rt_via_old_app():
+    # Let's pretend this is an old app powered by ADAL
+    app = msal.PublicClientApplication(
+        config["client_id"], authority=config["authority"])
+    flow = app.initiate_device_flow(scopes=config["scope"])
+    if "user_code" not in flow:
+        raise ValueError(
+            "Fail to create device flow. Err: %s" % json.dumps(flow, indent=4))
+    print(flow["message"])
+    sys.stdout.flush()  # Some terminal needs this to ensure the message is shown
+
+    # Ideally you should wait here, in order to save some unnecessary polling
+    # input("Press Enter after signing in from another device to proceed, CTRL+C to abort.")
+
+    result = app.acquire_token_by_device_flow(flow)  # By default it will block
+    assert "refresh_token" in result, "We should have a successful result"
+    return result["refresh_token"]
+
+try:  # For easier testing, we try to reload a RT from previous run
+    old_rt = json.load(open("rt.json"))[0]
+except:  # If that is not possible, we acquire a RT
+    old_rt = get_rt_via_old_app()
+    json.dump([old_rt], open("rt.json", "w"))
+
+# Now we will try to migrate this old_rt into a new app powered by MSAL
+
+token_cache = msal.SerializableTokenCache()
+assert token_cache.serialize() == '{}', "Token cache is initially empty"
+app = msal.PublicClientApplication(
+    config["client_id"], authority=config["authority"], token_cache=token_cache)
+result = app.import_refresh_token(old_rt, config["scope"])
+if "error" in result:
+    print("Migration unsuccessful. Error: ", json.dumps(result, indent=2))
+else:
+    print("Migration is successful")
+    logging.debug("Token cache contains: %s", token_cache.serialize())
+
+# From now on, the RT is saved inside MSAL's cache,
+# and becomes available in normal MSAL coding pattern. For example:
+accounts = app.get_accounts()
+if accounts:
+    account = accounts[0]  # Assuming end user pick this account
+    result = app.acquire_token_silent(config["scope"], account)
+    if "access_token" in result:
+        print("RT is available in MSAL's cache, and can be used to acquire new AT")
+
--- a/msal/application.py
+++ b/msal/application.py
@@ -700,6 +700,28 @@ class ClientApplication(object):
                     "you must include a string parameter named 'key_id' "
                     "which identifies the key in the 'req_cnf' argument.")
 
+    def import_refresh_token(self, refresh_token, scopes):
+        """Import an RT from elsewhere into MSAL's token cache.
+
+        :param str refresh_token: The old refresh token, as a string.
+
+        :param list scopes:
+            The scopes associate with this old RT.
+            Each scope needs to be in the Microsoft identity platform (v2) format.
+            https://docs.microsoft.com/en-us/azure/active-directory/develop/migrate-python-adal-msal#scopes-not-resources
+
+        :return:
+            * A dict contains "error" and some other keys, when error happened.
+            * A dict contains no "error" key.
+        """
+        result = self.client.obtain_token_by_refresh_token(
+            refresh_token,
+            decorate_scope(scopes, self.client_id),
+            rt_getter=lambda rt: rt,
+            on_updating_rt=False,
+            )
+        return {} if "error" not in result else result  # Returns NO token
+
 
 class PublicClientApplication(ClientApplication):  # browser app or mobile app
 


bgavrilMS:
There is a similar feature in MSAL.NET that @trwalke implemented - see https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/Adal-to-Msal#adalnet-v2x-migration-to-msalnet-v3x

rayluo:
A comparison between 2 flavors of API:

<table>
<tr>
<th>If we use a dedicated API which does NOT return an AT</th>
<th>If we use an acquire_token_by_refresh_token() which returns an AT</th>
</tr>
<tr>
<td>

```python
app = msal.PublicClientApplication(
    "client_id", ...)

# Proposal A
result = app.import_refresh_token(
    old_rt, ["scope1", "scope2"])
if "error" in result:
    print(
        "Error in Migration:",
        json.dumps(result))
else:
    print("Migration is successful")

# From now on,
# the RT is saved inside MSAL's cache,
# and becomes available in normal
# MSAL coding pattern. For example:
...
result = app.acquire_token_silent(
    ["scope1"], ...)
```
</td>

<td>

```python
app = msal.PublicClientApplication(
    "client_id", ...)

# Proposal B
result = app.acquire_token_by_refresh_token(
    old_rt, ["scope1", "scope2"])
if "error" in result:
    print(
        "Error in Migration:",
        json.dumps(result))
else:
    print("Migration is successful")

# From now on,
# the RT is saved inside MSAL's cache.

# But at the same time,
# the app could skip the normal
# MSAL coding pattern. For example:
use_the_token(result["access_token"])
```
</td>
</tr>
</table>

bgavrilMS:
I believe the idea with `acquire_token_by_refresh_token` is that it can be used in a similar style as other acquire_token_* methods.

```py
try:
       # if this works, migration is not needed or it already happened
       acquire_token_silent 
except UIRequiredException:
       # migration path ... can be removed after some time
       rt = get_legacy_refresh_token
       acquire_token_by_refersh_token 
except UIRequiredException:
      # no RT available, get a new one
      acquire_token_by_X
```
What are your thoughts on full scenario using `import_refresh_token` ?

rayluo:
There has been some offline discussion since sending out this PR. We will end up choosing one solution soon. But here in this comment, for the sake of answering @bgavrilMS 's [question above](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/191#issuecomment-627984143), and leaving a trace here for future reference:

* The formula: `import_refresh_token()` + `acquire_token_silent()` = [MSAL .Net's `it is not readily accessible with the IConfidentialClientApplication without first casting it to IByRefreshToken.`](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/Adal-to-Msal#adalnet-v2x-migration-to-msalnet-v3x) + `acquire_token_by_refresh_token()`

  + MSAL .Net's "needing to first cast it to a different interface" is an implicit yet required step, which effectively becomes part of that `acquire_token_by_refresh_token()` API calling pattern. A perception of "there is only one `AcquireTokenByRefreshToken()` API needed to be called", is not the full picture.

  + In Python language, there is technical limitation to do magic as "to cast an object as another", so one reason of the `import_refresh_token()` idea is to satisfy that original desire to hide the `AcquireTokenByRefreshToken()` behind a cast. BTW, MSAL C++ happens to come up with a migration function with exact same name `void ImportRefreshToken()`.

  + That being said, I'm also told that "hiding `AcquireTokenByRefreshToken()` behind a casting" is not a hard requirement, this time.

* Regardless of which flavor of API we end up using (because they ARE equivalent - see the "formula" above), an MSAL-powered app can still have freedom to choose their own two migration strategies.

  + One approach is to migrate RTs IN A BATCH within first few minutes, when a newer version of app (Azure CLI) is run for its first time on an end user's machine and detects a legacy RT persistence.

    ```python
    for rt, scopes in get_legacy_rt(...):
        app.import_refresh_token(rt, scopes)
    os.rename("adal_token_cache.bin", "legacy_cache_to_be_deleted.bin")
    # Migration completed. From now on, everything follows the normal MSAL way
    ```
    With this strategy, you do NOT even need to call an `acquire_token_silent()` immediately DURING the migration.

    The revised formula: `import_refresh_token()` ≈ [MSAL .Net's `it is not readily accessible with the IConfidentialClientApplication without first casting it to IByRefreshToken.`](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/Adal-to-Msal#adalnet-v2x-migration-to-msalnet-v3x) + `acquire_token_by_refresh_token()`

  + Another is to migrate RTs GRADUALLY across many days or even weeks, when the end user is using our app (such as Azure CLI) daily. That pattern is roughly demonstrated in C# (derived from @bgavrilMS 's [question above](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/191#issuecomment-627984143), modified with nested `try...catch...`).

    ```c#
    try
    {
           // if this works, migration is not needed or it already happened
           AcquireTokenSilent (...)
    }
    catch UIRequiredException
    {
        try
        {
               // migration path ... can be removed after some time
               var rt = GetLegacyRefreshToken(...)
               AcquireTokenByRefershToken(...)
        }
        catch UIRequiredException
        {
              // no RT available, get a new one
              AcquireTokenByXyz(...)
        }
    }
    ```

sample/migrate_rt.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+# Optional logging
+# logging.basicConfig(level=logging.DEBUG)  # Enable DEBUG log for entire script
+# logging.getLogger("msal").setLevel(logging.INFO)  # Optionally disable MSAL DEBUG logs
+
+config = json.load(open(sys.argv[1]))
+
+def get_rt_via_old_app():
+    # Let's pretend this is an old app powered by ADAL
```
abhidnya13, line 31: I think we can just use ADAL here to avoid confusion for people trying to migrate instead of this comment and then using an MSAL app instead.

abhidnya13:
Put my review comments on the other PR 193 to avoid duplication of comments

henrik-me:
>  With this strategy, you do NOT even need to call an acquire_token_silent() immediately DURING the migration.

And the way this is done in Java and .NET you don't need to get the result and do anything with it nor do you need to call acquire_token_silent if you needed to use the result right away.

acquire_token_by_refresh_token() gets my vote

rayluo:
> > With this strategy, you do NOT even need to call an acquire_token_silent() immediately DURING the migration.
>
> And the way this is done in Java and .NET you don't need to get the result and do anything with it nor do you need to call acquire_token_silent if you needed to use the result right away.

This is true. Regardless of whether we implement `import_refresh_token()` or `acquire_token_by_refresh_token()`, migrating-in-a-batch seems a better migration strategy. Interestingly, `import_refresh_token()`'s returning-nothing behavior will presumably lead app developer to the migrate-in-batch route; while on the contrary, `acquire_token_by_refresh_token()`'s returning-AT behavior could mislead app developer to Migrate-Gradually route.

> acquire_token_by_refresh_token() gets my vote

Yep. #193 implements `acquire_token_by_refresh_token()`.

PR #193, opened 2020-05-14
acquire_token_by_refresh_token() for RT migration

rayluo:
This PR implements the right-hand side of [this comparison](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/191#issuecomment-627640783). We can go with this direction, and then this PR will close #191.

The API calling pattern is largely the same as #191, if we do not really need to use its return value:

```python
app = msal.PublicClientApplication(
    "client_id", ...)

# Proposal B
result = app.acquire_token_by_refresh_token(
    old_rt, ["scope1", "scope2"])
if "error" in result:
    print(
        "Error in Migration:",
        json.dumps(result))
else:
    print("Migration is successful")
    # The new RT is saved inside MSAL's cache.
    # You could use the new AT, IDT from result at this point

```

There are still two higher level RT migration strategies:

  + One approach is to migrate RTs IN A BATCH within minutes, when a newer version of app (Azure CLI) is run for its first time on an end user's machine and detects a legacy RT persistence.

    ```python
    if os.path.exists("adal_token_cache.bin"):
        for rt, scopes in get_legacy_rt(...):
            app.acquire_token_by_refresh_token(rt, scopes)
        os.rename("adal_token_cache.bin", "legacy_cache_to_be_deleted.bin")
        # Migration completed. From now on,
        # everything follows the normal MSAL way,
        # i.e. acquire_token_silent(scopes)
    main_program(...)  # Main program starts
    ```
    With this strategy, you may safely discard the return value of each `acquire_token_by_refresh_token()` call, DURING the migration.

    The comparison: `acquire_token_by_refresh_token()` ≈ [MSAL .Net's `it is not readily accessible with the IConfidentialClientApplication without first casting it to IByRefreshToken.`](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/Adal-to-Msal#adalnet-v2x-migration-to-msalnet-v3x) + `acquire_token_by_refresh_token()`

  + Another approach is to migrate RTs GRADUALLY across many days or even weeks, when the end user is using our app (such as Azure CLI) frequently. But then you would have to use `acquire_token_by_refresh_token()` in your normal code path. The following demo is copied from [here](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/191#issuecomment-628159543).

    ```c#
    try
    {
           // if this works, migration is not needed or it already happened
           AcquireTokenSilent (...)
    }
    catch UIRequiredException
    {
        try
        {
               // migration path ... can be removed after some time
               var rt = GetLegacyRefreshToken(...)
               AcquireTokenByRefershToken(...)
        }
        catch UIRequiredException
        {
              // no RT available, get a new one
              AcquireTokenByXyz(...)
        }
    }
    ```

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -700,6 +700,28 @@ class ClientApplication(object):
                     "you must include a string parameter named 'key_id' "
                     "which identifies the key in the 'req_cnf' argument.")
 
+    def import_refresh_token(self, refresh_token, scopes):
+        """Import an RT from elsewhere into MSAL's token cache.
+
+        :param str refresh_token: The old refresh token, as a string.
+
+        :param list scopes:
+            The scopes associate with this old RT.
+            Each scope needs to be in the Microsoft identity platform (v2) format.
+            https://docs.microsoft.com/en-us/azure/active-directory/develop/migrate-python-adal-msal#scopes-not-resources
+
+        :return:
+            * A dict contains "error" and some other keys, when error happened.
+            * A dict contains no "error" key.
+        """
+        result = self.client.obtain_token_by_refresh_token(
+            refresh_token,
+            decorate_scope(scopes, self.client_id),
+            rt_getter=lambda rt: rt,
+            on_updating_rt=False,
+            )
+        return {} if "error" not in result else result  # Returns NO token
+
 
 class PublicClientApplication(ClientApplication):  # browser app or mobile app
 
--- a/msal/application.py
+++ b/msal/application.py
@@ -700,27 +700,35 @@ class ClientApplication(object):
                     "you must include a string parameter named 'key_id' "
                     "which identifies the key in the 'req_cnf' argument.")
 
-    def import_refresh_token(self, refresh_token, scopes):
-        """Import an RT from elsewhere into MSAL's token cache.
+    def acquire_token_by_refresh_token(self, refresh_token, scopes):
+        """Acquire token(s) based on a refresh token (RT) obtained from elsewhere.
+
+        You use this method only when you have old RTs from elsewhere,
+        and now you want to migrate them into MSAL.
+        Calling this method results in new tokens automatically storing into MSAL.
+
+        You do NOT need to use this method if you are already using MSAL.
+        MSAL maintains RT automatically inside its token cache,
+        and an access token can be retrieved
+        when you call :func:`~acquire_token_silent`.
 
         :param str refresh_token: The old refresh token, as a string.
 
         :param list scopes:
             The scopes associate with this old RT.
             Each scope needs to be in the Microsoft identity platform (v2) format.
-            https://docs.microsoft.com/en-us/azure/active-directory/develop/migrate-python-adal-msal#scopes-not-resources
+            See `Scopes not resources <https://docs.microsoft.com/en-us/azure/active-directory/develop/migrate-python-adal-msal#scopes-not-resources>`_.
 
         :return:
             * A dict contains "error" and some other keys, when error happened.
-            * A dict contains no "error" key.
+            * A dict contains no "error" key means migration was successful.
         """
-        result = self.client.obtain_token_by_refresh_token(
+        return self.client.obtain_token_by_refresh_token(
             refresh_token,
             decorate_scope(scopes, self.client_id),
             rt_getter=lambda rt: rt,
             on_updating_rt=False,
             )
-        return {} if "error" not in result else result  # Returns NO token
 
 
 class PublicClientApplication(ClientApplication):  # browser app or mobile app
--- a/sample/migrate_rt.py
+++ b/sample/migrate_rt.py
@@ -25,51 +25,43 @@ import msal
 # logging.basicConfig(level=logging.DEBUG)  # Enable DEBUG log for entire script
 # logging.getLogger("msal").setLevel(logging.INFO)  # Optionally disable MSAL DEBUG logs
 
+def get_preexisting_rt_and_their_scopes_from_elsewhere():
+    # Maybe you have an ADAL-powered app like this
+    #   https://github.com/AzureAD/azure-activedirectory-library-for-python/blob/1.2.3/sample/device_code_sample.py#L72
+    # which uses a resource rather than a scope,
+    # you need to convert your v1 resource into v2 scopes
+    # See https://docs.microsoft.com/azure/active-directory/develop/azure-ad-endpoint-comparison#scopes-not-resources
+    # You may be able to append "/.default" to your v1 resource to form a scope
+    # See https://docs.microsoft.com/azure/active-directory/develop/v2-permissions-and-consent#the-default-scope
+
+    # Or maybe you have an app already talking to Microsoft identity platform v2,
+    # powered by some 3rd-party auth library, and persist its tokens somehow.
+
+    # Either way, you need to extract RTs from there, and return them like this.
+    return [
+        ("old_rt_1", ["scope1", "scope2"]),
+        ("old_rt_2", ["scope3", "scope4"]),
+        ]
+
+
+# We will migrate all the old RTs into a new app powered by MSAL
 config = json.load(open(sys.argv[1]))
-
-def get_rt_via_old_app():
-    # Let's pretend this is an old app powered by ADAL
-    app = msal.PublicClientApplication(
-        config["client_id"], authority=config["authority"])
-    flow = app.initiate_device_flow(scopes=config["scope"])
-    if "user_code" not in flow:
-        raise ValueError(
-            "Fail to create device flow. Err: %s" % json.dumps(flow, indent=4))
-    print(flow["message"])
-    sys.stdout.flush()  # Some terminal needs this to ensure the message is shown
-
-    # Ideally you should wait here, in order to save some unnecessary polling
-    # input("Press Enter after signing in from another device to proceed, CTRL+C to abort.")
-
-    result = app.acquire_token_by_device_flow(flow)  # By default it will block
-    assert "refresh_token" in result, "We should have a successful result"
-    return result["refresh_token"]
-
-try:  # For easier testing, we try to reload a RT from previous run
-    old_rt = json.load(open("rt.json"))[0]
-except:  # If that is not possible, we acquire a RT
-    old_rt = get_rt_via_old_app()
-    json.dump([old_rt], open("rt.json", "w"))
-
-# Now we will try to migrate this old_rt into a new app powered by MSAL
-
-token_cache = msal.SerializableTokenCache()
-assert token_cache.serialize() == '{}', "Token cache is initially empty"
 app = msal.PublicClientApplication(
-    config["client_id"], authority=config["authority"], token_cache=token_cache)
-result = app.import_refresh_token(old_rt, config["scope"])
-if "error" in result:
-    print("Migration unsuccessful. Error: ", json.dumps(result, indent=2))
-else:
-    print("Migration is successful")
-    logging.debug("Token cache contains: %s", token_cache.serialize())
+    config["client_id"], authority=config["authority"],
+    # token_cache=...  # Default cache is in memory only.
+                       # You can learn how to use SerializableTokenCache from
+                       # https://msal-python.rtfd.io/en/latest/#msal.SerializableTokenCache
+    )
 
-# From now on, the RT is saved inside MSAL's cache,
-# and becomes available in normal MSAL coding pattern. For example:
-accounts = app.get_accounts()
-if accounts:
-    account = accounts[0]  # Assuming end user pick this account
-    result = app.acquire_token_silent(config["scope"], account)
-    if "access_token" in result:
-        print("RT is available in MSAL's cache, and can be used to acquire new AT")
+# We choose a migration strategy of migrating all RTs in one loop
+for old_rt, scopes in get_preexisting_rt_and_their_scopes_from_elsewhere():
+    result = app.acquire_token_by_refresh_token(old_rt, scopes)
+    if "error" in result:
+        print("Discarding unsuccessful RT. Error: ", json.dumps(result, indent=2))
 
+print("Migration completed")
+
+# From now on, those successfully-migrated RTs are saved inside MSAL's cache,
+# and becomes available in normal MSAL coding pattern, which is NOT part of migration.
+# You can refer to:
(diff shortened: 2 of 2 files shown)


sample/migrate_rt.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+# Optional logging
+# logging.basicConfig(level=logging.DEBUG)  # Enable DEBUG log for entire script
+# logging.getLogger("msal").setLevel(logging.INFO)  # Optionally disable MSAL DEBUG logs
+
+config = json.load(open(sys.argv[1]))
+
+def get_rt_via_old_app():
+    # Let's pretend this is an old app powered by ADAL
```
abhidnya13, line 31: Suggestion: I think we should  use ADAL here. It is confusing to understand when the comment mentions that this is an app powered by ADAL but is using MSAL in the very next line.

sample/migrate_rt.py
```diff
@@ -0,0 +1,75 @@
...
+app = msal.PublicClientApplication(
+    config["client_id"], authority=config["authority"], token_cache=token_cache)
+result = app.acquire_token_by_refresh_token(old_rt, config["scope"])
+if "error" in result:
+    print("Migration unsuccessful. Error: ", json.dumps(result, indent=2))
+else:
+    print("Migration is successful")
+    logging.debug("Token cache contains: %s", token_cache.serialize())
+
+# From now on, the RT is saved inside MSAL's cache,
```
abhidnya13, line 67: This statement won't hold true for the cases when migration was not successful. What should be done in that case ?

sample/migrate_rt.py
```diff
@@ -0,0 +1,75 @@
...
+
+    # Ideally you should wait here, in order to save some unnecessary polling
+    # input("Press Enter after signing in from another device to proceed, CTRL+C to abort.")
+
+    result = app.acquire_token_by_device_flow(flow)  # By default it will block
+    assert "refresh_token" in result, "We should have a successful result"
+    return result["refresh_token"]
+
+try:  # For easier testing, we try to reload a RT from previous run
+    old_rt = json.load(open("rt.json"))[0]
```
abhidnya13, line 49: Its not very clear to me what "rt.json" is expected to have. Just as we have a sample for how config.json but it still creates confusion, atleast mentioning how it will possibly look like can help.

sample/migrate_rt.py
```diff
@@ -0,0 +1,75 @@
...
+except:  # If that is not possible, we acquire a RT
+    old_rt = get_rt_via_old_app()
+    json.dump([old_rt], open("rt.json", "w"))
```
abhidnya13, line 52: Why do we want to store this old rt back in the file in this sample when we are not really retrieving from the file again anywhere later ?

sample/migrate_rt.py
```diff
@@ -0,0 +1,75 @@
...
+
+# Now we will try to migrate this old_rt into a new app powered by MSAL
+
+token_cache = msal.SerializableTokenCache()
```
abhidnya13, line 56: May be a link to the actual usage for persisting it would be helpful to avoid any confusion. https://msal-python.readthedocs.io/en/latest/#msal.SerializableTokenCache

sample/migrate_rt.py
```diff
@@ -0,0 +1,75 @@
...
+    logging.debug("Token cache contains: %s", token_cache.serialize())
+
+# From now on, the RT is saved inside MSAL's cache,
+# and becomes available in normal MSAL coding pattern. For example:
+accounts = app.get_accounts()
+if accounts:
+    account = accounts[0]  # Assuming end user pick this account
+    result = app.acquire_token_silent(config["scope"], account)
+    if "access_token" in result:
+        print("RT is available in MSAL's cache, and can be used to acquire new AT")
```
abhidnya13, line 74: Suggestion : `RT is available in MSAL's cache, and was used to acquire new AT` will be more clear of a message, suggesting the sample did what its supposed to do.

msal/application.py
```diff
@@ -700,6 +700,35 @@ def _validate_ssh_cert_input_data(self, data):
                     "you must include a string parameter named 'key_id' "
                     "which identifies the key in the 'req_cnf' argument.")
 
+    def acquire_token_by_refresh_token(self, refresh_token, scopes):
+        """Acquire token(s) based on a refresh token (RT) obtained from elsewhere.
+
+        In normal MSAL scenarios, you do NOT need to use this method.
```
abhidnya13, line 706: Normal MSAL scenarios does not give a very clear idea on what are these normal scenarios. May be we can write the whole paraphrase like this or any other way to be more clear:
```
You use this method only when you have old RTs from elsewhere,
and now you want to migrate them into MSAL.
Calling this method results in new tokens automatically storing into MSAL

You do NOT need to use this method if you are already using MSAL.
MSAL maintains RT automatically inside its token cache, and an access token can be retrieved
when you call :func:`~acquire_token_silent`.

**Comment by henrik-me** [comment]
this one gets my vote

File: msal/application.py (hunk 7 above)
Comment (henrik-me) [inline msal/application.py:706]: >In normal MSAL scenarios, you do NOT need to use this method. [](start = 8, length = 61)

[https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/1cef43d185bb0a1dcc013d2ec232ae9d4c6c053a/src/client/Microsoft.Identity.Client/IByRefreshToken.cs](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/1cef43d185bb0a1dcc013d2ec232ae9d4c6c053a/src/client/Microsoft.Identity.Client/IByRefreshToken.cs):
        /// Acquires an access token from an existing refresh token and stores it, and the refresh token, in
        /// the user token cache, where it will be available for further AcquireTokenSilent calls.
        /// This method can be used in migration to MSAL from ADAL v2, and in various integration
        /// scenarios where you have a RefreshToken available.
        /// See [https://aka.ms/msal-net-migration-adal2-msal2](https://aka.ms/msal-net-migration-adal2-msal2).

You can also get inspiration on the text from the Java API. Should be similar.

Please add aka.ms links to and create a wiki page allowing you to explain this in more details as well as updating content and linking to it easily when customers have questions.

File: msal/application.py [hunk 8]
```diff
@@ -700,6 +700,35 @@ def _validate_ssh_cert_input_data(self, data):
...
+        You use this method only when you have old RTs from elsewhere,
+        and now you want to migrate them into MSAL.
+        Calling this method results in new tokens automatically storing into MSAL.
+
+        :param str refresh_token: The old refresh token, as a string.
+
+        :param list scopes:
+            The scopes associate with this old RT.
+            Each scope needs to be in the Microsoft identity platform (v2) format.
+            https://docs.microsoft.com/en-us/azure/active-directory/develop/migrate-python-adal-msal#scopes-not-resources
```
henrik-me, line 719: >https://docs.microsoft.com/en-us/azure/active-d [](start = 12, length = 47)

aka.ms link?

sample/migrate_rt.py
henrik-me: +1 to use this as a way to demonstrate a typical scenario and explain also the resource to scope fun

---
In reply to: [424830506](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/193#discussion_r424830506) [](ancestors = 424830506)

sample/migrate_rt.py
henrik-me: aka.ms

---
In reply to: [424837036](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/193#discussion_r424837036) [](ancestors = 424837036)

sample/migrate_rt.py
```diff
@@ -0,0 +1,75 @@
...
+    old_rt = get_rt_via_old_app()
+    json.dump([old_rt], open("rt.json", "w"))
+
+# Now we will try to migrate this old_rt into a new app powered by MSAL
+
+token_cache = msal.SerializableTokenCache()
+assert token_cache.serialize() == '{}', "Token cache is initially empty"
+app = msal.PublicClientApplication(
+    config["client_id"], authority=config["authority"], token_cache=token_cache)
+result = app.acquire_token_by_refresh_token(old_rt, config["scope"])
```
henrik-me, line 60: >result [](start = 0, length = 6)

You can also validate that this has an AT and an RT, in case it's not an error state.

sample/migrate_rt.py
```diff
@@ -0,0 +1,75 @@
...
+else:
+    print("Migration is successful")
+    logging.debug("Token cache contains: %s", token_cache.serialize())
+
+# From now on, the RT is saved inside MSAL's cache,
+# and becomes available in normal MSAL coding pattern. For example:
+accounts = app.get_accounts()
+if accounts:
+    account = accounts[0]  # Assuming end user pick this account
+    result = app.acquire_token_silent(config["scope"], account)
```
henrik-me, line 72: >result [](start = 4, length = 6)

this should be compared to the above result evaluating if those are the same (expected to be)

msal/application.py
```diff
@@ -700,6 +700,35 @@ def _validate_ssh_cert_input_data(self, data):
...
+
+        :return:
+            * A dict contains "error" and some other keys, when error happened.
+            * A dict contains no "error" key means migration was successful.
+        """
+        return self.client.obtain_token_by_refresh_token(
+            refresh_token,
+            decorate_scope(scopes, self.client_id),
+            rt_getter=lambda rt: rt,
+            on_updating_rt=False,
```
henrik-me, line 729: >on_updating_rt=False [](start = 12, length = 20)

Why is this needed?

rayluo: Long story short, it is the internal helper's behavior, which should be considered implementation detail. The fact that we now provide this new `acquire_token_by_refresh_token()` as first-class citizen API, hides such detail from caller.

msal/application.py
rayluo: Instead of creating yet another short link, the next commit will change such naked url into "[Scopes not resources](https://docs.microsoft.com/en-us/azure/active-directory/develop/migrate-python-adal-msal#scopes-not-resources)", then using its canonical long url will not be a problem.

msal/application.py
rayluo: We probably do not need to create yet another wiki page. After this PR is released, we can modify the [existing doc](https://docs.microsoft.com/en-us/azure/active-directory/develop/migrate-python-adal-msal#migrate-existing-refresh-tokens-for-msal-python) accordingly.

sample/migrate_rt.py
rayluo: Interestingly, this script was created as a [quick-and-dirty test driver](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/193/files#diff-615899d8b1282d9f785b5404cf87e0ccR48-R52), and it is already unfortunate that it ended up with much more lines in preparing an old RT than demonstrating the actual migration (thus distracting a sample reader). AND now it does attract more attentions and comments than necessary.

Here is the thing.

* The old app does not have to be an ADAL Python app. It could also be an app developer wanting to migrate his/her app powered by a 3rd-party generic OAuth2 library to MSAL.
* Even if the old app is using ADAL Python, there is actually not a specific thing as "adal python token cache". ADAL Python never enforced a particular token cache persistence format. To test such migration, we would need to call an old (adal?) app, acquire tokens, INVENT an ad-hoc token storage format, pretend to read token from it, ..., etc.. Our ad-hoc token storage format does not necessarily match what an ADAL Python app developer's situation, thus might just be noise to them.

Now I incline to replace this entire `def get_rt_via_old_app()` implementation into something like this:

```python
def get_rt_via_old_app():
    # Maybe you have an ADAL-powered app like this
    # https://github.com/AzureAD/azure-activedirectory-library-for-python/blob/dev/sample/device_code_sample.py
    # or you have an app already talking to Microsoft identity platform v2,
    # powered by some 3rd-party auth library.
    # And you already persist its tokens somehow.
    # You are expected to extract RTs from there, and return it like this.
    return "old_rt", ["scope1", "scope2"]
```

which is very close to the [previous doc here](https://docs.microsoft.com/en-us/azure/active-directory/develop/migrate-python-adal-msal#migrate-existing-refresh-tokens-for-msal-python).

sample/migrate_rt.py
rayluo: We do not have to do that here to further pollute this already too-complicated sample. Checking "error" field is the official (and only) way to check whether a migration was successful, which is already [documented in its API Reference](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/193/files#diff-f7ae931bcdab4b14f3a256baf7d6922aR721-R723).

sample/migrate_rt.py
rayluo: Good catch. Will fix in next commit.

sample/migrate_rt.py
rayluo: I realize you are probably think about some unit test case. Yes we should probably add some, but again, those unit test checks will not and should not show up in this sample.

sample/migrate_rt.py
rayluo: Well, I purposely use the word `can`. :-) Because that AT is more likely an old AT obtained during migration. No new AT is being requested.

Again, this is exactly an example that this sample begins to distract readers. Let's try to come up with something as concise as possible. Stay tuned.

rayluo:
Quoted a test feedback from @qianwens:

> I have tried your PR and it works perfectly. I can use an ADAL refresh token to generate the MSAL token cache with your method and CLI command works well with the generated MSAL token.
>
> Thanks very much for implementing this function in such a short time.

sample/migrate_rt.py
abhidnya13: Yes, makes sense. However here, we are still skipping the part where we explain what scope1 and scope2 mean and if they are still using resources to acquire this RT what would that conversion from resource to scope look like.

rayluo: See the latest commit.

msal/application.py
rayluo: > After this PR is released, we can modify the [existing doc](https://docs.microsoft.com/en-us/azure/active-directory/develop/migrate-python-adal-msal#migrate-existing-refresh-tokens-for-msal-python) accordingly.

UPDATE:
A new [doc PR for the migration docs](https://github.com/MicrosoftDocs/azure-docs-pr/pull/116270) has been created just now.

PR #205, opened 2020-06-09
Application initializer does not make tenant discovery calls

### LINKED ISSUE #187: ClientApplication initializer sends HTTP requests
Constructing `ClientApplication` entails at least one HTTP request, for tenant discovery. This adds all the costs and failure modes associated with the network to constructing the class. In my own code I therefore defer constructing (subclasses of) `ClientApplication` until I want to authenticate. That's a feasible workaround but please consider taking a similar approach in msal, deferring discovery until a method requiring I/O is called.

abhidnya13:
UPDATE: This will fix #187


msal/application.py
```diff
@@ -194,12 +194,16 @@ def __init__(
                 self.http_client.request, timeout=timeout)
         self.app_name = app_name
         self.app_version = app_version
+        self.authority_url = authority
+        self.validate_authority = validate_authority
         self.authority = Authority(
```
rayluo, line 199: Since we are heading to this direction anyway, let's do it thoroughly, by also lazy-building the Authority instance. One of the approaches to do this, is to add something like this at each API which needs to use `self.authority`:

```python
if not self.authority.is_initialized:
    self.authority.initialize()  # It will send out network calls which is currently inside Authority constructor
```

tests/public_cert.txt
```diff
@@ -0,0 +1,117 @@
+-----BEGIN CERTIFICATE-----
```
rayluo, line 1: Is checking-in such cert intended? I'm not seeing it being used here.

msal/application.py
```diff
@@ -194,12 +194,16 @@ def __init__(
                 self.http_client.request, timeout=timeout)
         self.app_name = app_name
         self.app_version = app_version
+        self.authority_url = authority
+        self.validate_authority = validate_authority
```
rayluo, line 198: It seems like these 2 lines are not needed in the current approach?

msal/authority.py
```diff
@@ -53,15 +53,20 @@ def __init__(self, authority_url, http_client, validate_authority=True):
             performed.
         """
         self._http_client = http_client
-        authority, self.instance, tenant = canonicalize(authority_url)
-        parts = authority.path.split('/')
-        is_b2c = any(self.instance.endswith("." + d) for d in WELL_KNOWN_B2C_HOSTS) or (
+        self.authority, self.instance, self.tenant = canonicalize(authority_url)
```
rayluo, line 56: I haven't tried this, but it seems to be a possible direction to explore. What if we just record the 3 raw input here as internal members:

```python
self._http_client = http_client
self._authority_url = authority_url
self._validate_authority = validate_authority
self._is_initialized = False
```

Ant then move ALL the initialization into `initialize()`? That way, the change would likely be cleaner.

msal/authority.py
```diff
@@ -53,15 +53,20 @@ def __init__(self, authority_url, http_client, validate_authority=True):
...
+        parts = self.authority.path.split('/')
+        self.is_b2c = any(self.instance.endswith("." + d) for d in WELL_KNOWN_B2C_HOSTS) or (
             len(parts) == 3 and parts[2].lower().startswith("b2c_"))
-        if (tenant != "adfs" and (not is_b2c) and validate_authority
+        self.authority_url = authority_url
+        self.validate_authority = validate_authority
+        self.is_initialized = False
+
+    def initialize(self):
```
rayluo, line 64: Also a possible trick here. We could define all the actual initialization work in a `__initialize()`. And then the public `initialize()` would just allow [reentrancy](https://en.wikipedia.org/wiki/Reentrancy_(computing)).

Once we reach that far, we can go back to revisit how the application.py would be able to call this.

msal/authority.py
rayluo: Getting close. :-)

Suggest to give all these internal properties a leading underscore. That is the only conventional approach to scare downstream app developers away from making a direct dependencies of those properties.

msal/authority.py
```diff
@@ -53,11 +53,22 @@ def __init__(self, authority_url, http_client, validate_authority=True):
...
+        if not self.is_initialized:
+            self.__initialize()
+            self.is_initialized = True
+
+    def __initialize(self):
```
rayluo, line 65: Looking better and better now. :-)  Perhaps we can try to go one step further, by doing this:

```suggestion
    def __initialize(self, authority_url, http_client, validate_authority):
```

so that its implementation would probably be exactly the same as the previous `__init__()`.

msal/application.py
```diff
@@ -198,8 +198,9 @@ def __init__(
                 authority or "https://login.microsoftonline.com/common/",
                 self.http_client, validate_authority=validate_authority)
             # Here the self.authority is not the same type as authority in input
+        self.client = None
         self.token_cache = token_cache or TokenCache()
-        self.client = self._build_client(client_credential, self.authority)
+        self.client_credential = client_credential
```
rayluo, line 203:
```suggestion
        self._client_credential = client_credential
```
We won't change pre-existing property names, but from now on, let's be more conscious on using leading underscore on internal property names.

msal/application.py
```diff
@@ -701,6 +707,10 @@ def _validate_ssh_cert_input_data(self, data):
                     "you must include a string parameter named 'key_id' "
                     "which identifies the key in the 'req_cnf' argument.")
 
+    def _setup_before_first_network_call(self):
```
rayluo, line 710: Perhaps this can be renamed as `_get_client(self)` and also internally have the same "reentrance" trick we used in authority.py? That way the callers might be able to just do `return self._get_client().obtain_token_by_xyz(...)`.

msal/application.py
```diff
@@ -825,8 +819,7 @@ def acquire_token_by_username_password(
             - A successful response would contain "access_token" key,
             - an error response would contain "error" and usually "error_description".
         """
-        if not self.client:
-            self._setup_before_first_network_call()
+        self._get_client()
```
rayluo, line 822: This line functionally works. It is a different pattern here, when comparing to your implementations in other `acquire_token_with_xyz()` methods. Which one do you think is better, and why? :-)

msal/application.py
```diff
@@ -701,6 +705,12 @@ def _validate_ssh_cert_input_data(self, data):
                     "you must include a string parameter named 'key_id' "
                     "which identifies the key in the 'req_cnf' argument.")
 
+    def _get_client(self):
```
rayluo, line 708: Here comes a totally subjective suggestion. I think it would be better to move this method higher in this source file, to right below the pre-existing `_build_client()`. This way, the 2 correlated methods stay together, thus easier to read and understand for future maintainers. They will thank us. :-)

msal/application.py
abhidnya13: So, for this one we need some attributes of Authority even before we need our client. In previous implementation, those would get configured in the initialization of the authority in the app __init__ so we can do network calls in authority and client initialization as _get_client().obtain_token_xyz(). For this one, since we need the authority attributes, I chose to initialize the client in the first step.

msal/application.py
abhidnya13: I think thats a good suggestion, will move it up in the next commits

msal/application.py
rayluo: Oh I see. Then I would suggest, instead of adding this line here, we add an explicit `self.authority.initialize()` at line 828.5 and then go back to the same `return self._get_client().obtain_tokan_by_xyz(...)` pattern we used elsewhere.

abhidnya13:
Tested the public api and added a few more changes

msal/application.py
```diff
@@ -862,16 +863,16 @@ def _acquire_token_by_username_password_federated(
         GRANT_TYPE_SAML1_1 = 'urn:ietf:params:oauth:grant-type:saml1_1-bearer'
         grant_type = {
             SAML_TOKEN_TYPE_V1: GRANT_TYPE_SAML1_1,
-            SAML_TOKEN_TYPE_V2: self.client.GRANT_TYPE_SAML2,
+            SAML_TOKEN_TYPE_V2: self._get_client().GRANT_TYPE_SAML2,
```
rayluo, line 866: For this one and the next one and the `encode_saml_assertion`, we can probably go with a different calling pattern like this:
```suggestion
            SAML_TOKEN_TYPE_V2: Client.GRANT_TYPE_SAML2,
```

Actually, since you missed line 874's `self.client.encode_saml_assertion`, it seems your testing strategy is to "run it (initially by test automation, and now by manual test), and see if it is broken, if so, fix it". That is not enough. We should aim for have a white-box decision BEFORE attempting to run the test cases. In this PR, at this point, we would know we need to search all occurrences of pre-existing `self.client` and then review each and every one to see whether/how to adjust it.

abhidnya13: I actually did the part where I looked through every occurrence of self.client and then decide if its is needed to be adjusted and I thought these changes are the minimun needed. However for cases you have suggested, if it does not really matter to initialize a client as we are pulling constant values, we can make such a change.

msal/application.py
```diff
@@ -809,6 +820,7 @@ def acquire_token_by_username_password(
             - A successful response would contain "access_token" key,
             - an error response would contain "error" and usually "error_description".
         """
+        self.authority.initialize()
```
rayluo, line 823: Nit: Suggest to move this line to line 829.5 i.e. right where it is needed. So that if/when in future we would refactor the authority logic for ADFS, we would be less likely to forget this otherwise-separated `self.authority.initialize()`.

msal/application.py
rayluo: So, does that mean the latest commit already address all the existing `self.client`? If you intended to leave some of them unchanged, we can discuss specifically. Right now I do not know whether those left-behind are intended or not.

abhidnya13: So, all the self.clients which were passed in functions as parameters which were called on self.client object were initially not changed as client object will get initialized before function parameters are evaluated. However, that can depend and vary on how python compiler processes a function call. To avoid this dependency, we realized after the offline conversation that all the places where self.client is still used are global variables and do not really depend on the specific instance of the class, so they have been changed to use Client.variable_name instead.

abhidnya13:
@rayluo summarized our offline conversation in the comment to give a context on the changes made!

PR #209, opened 2020-06-19
Migration should fail gracefully even on wrong RT

### LINKED ISSUE #208: acquire_token_by_refresh_token - AttributeError: 'str' object has no attribute 'get'
I'm getting the following error when trying to request an access_token via refreh_token:  

```
File "/usr/local/lib/python3.7/site-packages/msal/application.py", line 730, in acquire_token_by_refresh_token
    on_updating_rt=False,
  File "/usr/local/lib/python3.7/site-packages/msal/oauth2cli/oauth2.py", line 528, in obtain_token_by_refresh_token
    (on_removing_rt or self.on_removing_rt)(token_item)  # Discard old RT
  File "/usr/local/lib/python3.7/site-packages/msal/token_cache.py", line 241, in remove_rt
    assert rt_item.get("credential_type") == self.CredentialType.REFRESH_TOKEN
AttributeError: 'str' object has no attribute 'get'
```

Does anybody know why this happens? The refresh token should be valid (tested locally).

rayluo:
Fix #208

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -284,7 +284,7 @@ class ClientApplication(object):
             Can be one of "consumers" or "organizations" or your tenant domain "contoso.com".
             If included, it will skip the email-based discovery process that user goes
             through on the sign-in page, leading to a slightly more streamlined user experience.
-            More information on possible values 
+            More information on possible values
             `here <https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow#request-an-authorization-code>`_ and
             `here <https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-oapx/86fb452d-e34a-494e-ac61-e526e263b6d8>`_.
         :return: The authorization url as a string.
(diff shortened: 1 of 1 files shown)


msal/application.py
```diff
@@ -284,7 +284,7 @@ def get_authorization_request_url(
             Can be one of "consumers" or "organizations" or your tenant domain "contoso.com".
             If included, it will skip the email-based discovery process that user goes
             through on the sign-in page, leading to a slightly more streamlined user experience.
-            More information on possible values
```
abhidnya13, line 287: Is this change needed ?

rayluo: Absolutely not. :-) And really glad that you notice this!

It was, however, an unnecessary trailing white space in the first place, which we failed to notice it when [it was introduced in this PR](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/197/files#diff-f7ae931bcdab4b14f3a256baf7d6922aR287). :-(

That has to be fixed otherwise our future PR(s) would still encounter this conversation. I/we could have create a dedicate PR only to fix that trailing space.

By the way, do you happen to know whether PyCharm can be configured to automatically trim all the trailing spaces? That's the first thing I do whenever I install a new editor.

abhidnya13: Yes unfortunate that it got introduced in the first place, I guess there should be a setting to configure this in PyCharm,  will do this configuration so these kind of spaces aren't introduced next time onwards :)

PR #220, opened 2020-06-30
Removing hardcoded client_ids from test environment

### LINKED ISSUE #180: Pull client_id from Lab API query
Modify the existing test environment to get client_id strings from a query to the Lab API instead of hardcoding them.

abhidnya13:
List of client_ids changed:
- onprem ADFS2019 test cases - Using values from here : https://docs.msidlab.com/accounts/opadfsv2019.html
- OBO flow does not have a query exposed yet so storing the client ids as environment variables for now
- B2C test cases - Using values from https://msidlab.com/api/App?azureenvironment=azureb2ccloud

Fixes #180

### DIFF
--- a/tests/test_e2e.py
+++ b/tests/test_e2e.py
@@ -510,7 +510,6 @@ class WorldWideTestCase(LabBasedTestCase):
         os.getenv("LAB_OBO_PUBLIC_CLIENT_ID"),
         "Public client id can be found here https://docs.msidlab.com/flows/onbehalfofflow.html")
     def test_acquire_token_obo(self):
-
         config = self.get_lab_user(usertype="cloud")
 
         config_cca = {}
--- a/tests/test_e2e.py
+++ b/tests/test_e2e.py
@@ -503,28 +503,29 @@ class WorldWideTestCase(LabBasedTestCase):
     @unittest.skipUnless(
         os.getenv("LAB_OBO_CLIENT_SECRET"),
         "Need LAB_OBO_CLIENT SECRET from https://msidlabs.vault.azure.net/secrets/TodoListServiceV2-OBO/c58ba97c34ca4464886943a847d1db56")
+    @unittest.skipUnless(
+        os.getenv("LAB_OBO_CONFIDENTIAL_CLIENT_ID"),
+        "Confidential client id can be found here https://docs.msidlab.com/flows/onbehalfofflow.html")
+    @unittest.skipUnless(
+        os.getenv("LAB_OBO_PUBLIC_CLIENT_ID"),
+        "Public client id can be found here https://docs.msidlab.com/flows/onbehalfofflow.html")
     def test_acquire_token_obo(self):
-        if os.getenv("LAB_OBO_CONFIDENTIAL_CLIENT_ID") and os.getenv("LAB_OBO_PUBLIC_CLIENT_ID"):
-            config = self.get_lab_user(usertype="cloud")
 
-            config_cca = {}
-            config_cca.update(config)
-            config_cca["client_id"] = os.getenv("LAB_OBO_CONFIDENTIAL_CLIENT_ID")
-            config_cca["scope"] = ["https://graph.microsoft.com/.default"]
-            config_cca["client_secret"] = os.getenv("LAB_OBO_CLIENT_SECRET")
+        config = self.get_lab_user(usertype="cloud")
 
-            config_pca = {}
-            config_pca.update(config)
-            config_pca["client_id"] = os.getenv("LAB_OBO_PUBLIC_CLIENT_ID")
-            config_pca["password"] = self.get_lab_user_secret(config_pca["lab_name"])
-            config_pca["scope"] = ["api://%s/read" % config_cca["client_id"]]
+        config_cca = {}
+        config_cca.update(config)
+        config_cca["client_id"] = os.getenv("LAB_OBO_CONFIDENTIAL_CLIENT_ID")
+        config_cca["scope"] = ["https://graph.microsoft.com/.default"]
+        config_cca["client_secret"] = os.getenv("LAB_OBO_CLIENT_SECRET")
 
-            self._test_acquire_token_obo(config_pca, config_cca)
-        else:
-            logger.info("ENV variables LAB_OBO_CONFIDENTIAL_CLIENT_ID and/or LAB_OBO_PUBLIC_CLIENT_ID are not defined.")
-            # See https://docs.msidlab.com/flows/onbehalfofflow.html for details
-            raise unittest.SkipTest("Env variables were not found for OBO flow")
+        config_pca = {}
+        config_pca.update(config)
+        config_pca["client_id"] = os.getenv("LAB_OBO_PUBLIC_CLIENT_ID")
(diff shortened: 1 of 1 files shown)


tests/test_e2e.py
```diff
@@ -482,7 +482,6 @@ def test_ropc_adfs2019_onprem(self):
         # Configuration is derived from https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/4.7.0/tests/Microsoft.Identity.Test.Common/TestConstants.cs#L250-L259
         config = self.get_lab_user(usertype="onprem", federationProvider="ADFSv2019")
         config["authority"] = "https://fs.%s.com/adfs" % config["lab_name"]
-        config["client_id"] = "PublicClientId"
```
rayluo, line 485: Glad that this works. Finally. The previous `"PublicClientId"` thing was too magical. :-)

tests/test_e2e.py
```diff
@@ -506,21 +504,27 @@ def test_adfs2019_onprem_acquire_token_by_auth_code(self):
...
-        config_cca["client_id"] = "f4aa5217-e87c-42b2-82af-5624dd14ee72"
-        config_cca["scope"] = ["https://graph.microsoft.com/.default"]
-        config_cca["client_secret"] = os.getenv("LAB_OBO_CLIENT_SECRET")
-
-        config_pca = {}
-        config_pca.update(config)
-        config_pca["client_id"] = "c0485386-1e9a-4663-bc96-7ab30656de7f"
-        config_pca["password"] = self.get_lab_user_secret(config_pca["lab_name"])
-        config_pca["scope"] = ["api://%s/read" % config_cca["client_id"]]
+        if os.getenv("LAB_OBO_CONFIDENTIAL_CLIENTID") and os.getenv("LAB_OBO_PUBLIC_CLIENTID"):
```
rayluo, line 507: * The term `client_id` defined in OAuth2 specs has long been affecting our naming convention in this code base, namely, defined as TWO words connected by underscore. Please follow such naming convention.
* This condition can go into line 504, can't it? Then the majority of this function would remain the same.

tests/test_e2e.py
```diff
@@ -535,20 +539,23 @@ def test_b2c_acquire_token_by_auth_code(self):
...
             password="***"  # From https://aka.ms/GetLabUserSecret?Secret=msidlabb2c
         """
+        config = self.get_lab_app_object(azureenvironment="azureb2ccloud")
         self._test_acquire_token_by_auth_code(
             authority=self._build_b2c_authority("B2C_1_SignInPolicy"),
-            client_id="b876a048-55a5-4fc5-9403-f5d90cb1c852",
+            client_id=config["appId"],
             port=3843,  # Lab defines 4 of them: [3843, 4584, 4843, 60000]
-            scope=["https://msidlabb2c.onmicrosoft.com/msaapp/user_impersonation"]
+            scope=config["defaultScopes"].split(',')  # Default scopes returned by lab are
```
rayluo, line 547: Why do we want to document the exact scopes return by lab here? Is that important?

tests/test_e2e.py
abhidnya13: I'll change the values of variables to have underscores in them,
Also yes you can have them in line 504, however the reason its not added there is because the source of these values are different that the source of the secret in line 504. So, if we end up adding these there, it will be difficult to understand exactly what is missing and where to go to fetch the value. Also once we have the API available, these should hopefully be gone as well.

tests/test_e2e.py
abhidnya13: You are right, not really needed, removed them in the next commit

rayluo: Thanks for making the change. With the comments now gone, the missing of a trailing comma comes into attention. Python happens to allow an extra comma following the last parameter of a function, and we can take advantage of that.

tests/test_e2e.py
rayluo: Glad that you take into consideration on where/how to present a helpful description. But then let's explore whether we can avoid change our code structure. Possible solutions:

1. Can two `skipUnless(...)` be chained? I haven't encountered such a need before so I don't know. Please try it out and let me know. :-)
2. Or we can always use one `skipUnless(..., """a long long multi-line description as two paragraphs""")`
3. I know at least No.2 above would work but, for the sake of completeness, even if both of them would fail, we could always choose to postpone this change until the corresponding lab api arrives. In other words, it is not worthy to alter the otherwise-optimal code structure for a not-really-helpful temporary adjustment. (In case you don't realize it, the introduction of these 2 environment variables would not give us the flexibility from calling a not-yet-available lab api, and, it would make our own testing on local machine more troublesome than before. So if you/we did not set up our local testing environment due to the troublesome setup, this change alone would make it even less likely to happen.)

abhidnya13: I didnt really see a simple way to just chain the `skipUnless(..)` together, we can definitely do the second option. I however think that I'll prefer keeping it this way. I just think that adding too many conditions in the skipUnless can be confusing to debug if one of these environment variables goes missing. Also about your concern, we are not really changing the existing code structure, we are just adding some things keeping the existing code structure as is.  I understand that it might be harder to setup local testing environment as there will be an additional step to set these environment variables in local testing environment but I feel its a good enough one time setup to do if doing that setup means we would not be  exposing these values in the code.

rayluo: > I didnt really **see** a simple way to just chain the skipUnless(..) together

It seems to work on my machine (TM). :-)

```python
# This is the content of foo.py
import unittest


CONDITION_A = "Y"
CONDITION_B = "Y"


class TestFoo(unittest.TestCase):

    @unittest.skipUnless(CONDITION_A == "Y", "condition A not met")
    @unittest.skipUnless(CONDITION_B == "Y", "condition B not met")
    def test_foo(self):
        raise ValueError("If you see this, this test case was run")


unittest.main()
```

And then `python foo.py -v`

abhidnya13: Works for me too ! Thanks for this, did not realize chaining the skipUnless in this way is possible. I have made the change, please take a look :)

tests/test_e2e.py
```diff
@@ -535,20 +539,22 @@ def test_b2c_acquire_token_by_auth_code(self):
...
             password="***"  # From https://aka.ms/GetLabUserSecret?Secret=msidlabb2c
         """
+        config = self.get_lab_app_object(azureenvironment="azureb2ccloud")
         self._test_acquire_token_by_auth_code(
             authority=self._build_b2c_authority("B2C_1_SignInPolicy"),
-            client_id="b876a048-55a5-4fc5-9403-f5d90cb1c852",
+            client_id=config["appId"],
             port=3843,  # Lab defines 4 of them: [3843, 4584, 4843, 60000]
-            scope=["https://msidlabb2c.onmicrosoft.com/msaapp/user_impersonation"]
+            scope=config["defaultScopes"].split(',')
```
rayluo, line 547:
```suggestion
            scope=config["defaultScopes"].split(','),
```
Quoted from [PEP8](https://www.python.org/dev/peps/pep-0008/#when-to-use-trailing-commas):

> The pattern is to put each value (etc.) on a line by itself, always adding a trailing comma, and add the close parenthesis/bracket/brace on the next line.

abhidnya13: Updated :)

Issue #223, opened 2020-07-06
Acquiring token silently with account set to None fails in Azure Government


marlonsingleton:
**Describe the bug**

UPDATE by Ray: There is also an internal issue created [here](https://portal.microsofticm.com/imp/v3/incidents/details/195289801/home?fullScreen=incident-summary), which is currently under investigation by our service side team.

**To Reproduce**
1. Go to our [off-the-shelf samples](https://github.com/AzureAD/microsoft-authentication-library-for-python/tree/dev/sample) and pick one that is closest to your usage scenario. You should not need to modify the sample.
2. Follow the description of the sample, typically at the beginning of it, to prepare a `config.json` containing your test configurations
3. Run such sample, typically by `python sample.py config.json`
4. See the error
5. In this bug report, tell us the sample you choose, paste the content of the config.json with your test setup (which you can choose to skip your credentials, and/or mail it to our developer's email).
```
import msal
import os
import sys
from pprint import pprint

SCOPES = ["https://graph.microsoft.us/.default"]

app = msal.ConfidentialClientApplication(
    client_id="xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
    client_credential=['CLIENT_SECRET'],
    authority='https://login.microsoftonline.us/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx',
)

result = app.acquire_token_silent(scopes=SCOPES, account=None)
if not result:
    result = app.acquire_token_for_client(scopes=SCOPES)

if 'access_token' not in result:
    msg = "Could not acquire Azure access token: {error}: {descr}".format(
        error=result.get('error'),
        descr=result.get('error_description'),
    )
    pprint(msg)
    sys.exit(1)

pprint("Finished")
pprint(result)
```

**Expected behavior**
```
#result should equal None, evaluate if statement then acquire token for client
#works as expected in Azure Public cloud

result = app.acquire_token_silent(scopes=SCOPES, account=None)
if not result:
    result = app.acquire_token_for_client(scopes=SCOPES)
```
**What you see instead**
```
Traceback starts with:
 httplib_response = self._make_request(

Traceback ends with:
(Caused by SSLError(SSLCertVerificationError("hostname 'login.usgovcloudapi.net' doesn't match either of 'graph.aadg.microsoftonline.us', '*.b2clogin.us'")))
```

**The MSAL Python version you are using**
Paste the output of this
[...]

PR #225, opened 2020-07-07
Adding test case for acquire_token_silent in Arlington

abhidnya13:
This PR adds a test case to reproduce the error found in #223. It is expected to fail right now.

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -552,6 +552,12 @@ class ClientApplication(object):
             return result
         final_result = result
         for alias in self._get_authority_aliases(self.authority.instance):
+            if not self.token_cache.find(
+                    self.token_cache.CredentialType.REFRESH_TOKEN,
+                    target=scopes,
+                    query={"environment": alias}):
+                # Skip heavy weight logic when RT for this alias doesn't exist
+                continue
             the_authority = Authority(
                 "https://" + alias + "/" + self.authority.tenant,
                 self.http_client,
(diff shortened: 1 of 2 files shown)


msal/application.py
```diff
@@ -551,6 +551,13 @@ def acquire_token_silent_with_error(
             return result
         final_result = result
         for alias in self._get_authority_aliases(self.authority.instance):
+            if not self.token_cache.find(
+                    self.token_cache.CredentialType.REFRESH_TOKEN,
+                    target=scopes,
+                    query={"environment": alias}):
+                # Skip heavy weigh logic when RT for this alias doesn't exist.
```
rayluo, line 558: @abhidnya13 
I went ahead and made some changes, but github does not allow me to set you as reviewer (presumably because you were the PR creator). Please leave your review as comments here.

abhidnya13: This would also skip finding an AT if there exists one though which is done in line 594

rayluo: True. The reasoning here is to use a simplest-possible, good-enough logic as a heuristic, rather than duplicates most of the real `acquire_token_silent(...)` logic here. Given that RT basically always exists in an active alias, this is probably a good-enough approach to decide whether an alias is worth further searching.

abhidnya13: That is true that this is a good enough approach given the reason you mentioned. I think we can merge this in now to provide a quick fix but may be this is worth exploring and putting some more time to refactor the code sometime later to implement a foolproof approach instead which won't skip finding an AT

PR #230, opened 2020-07-15
Changing case of messageID in WS-Trust request document

### LINKED ISSUE #228: WS-Trust Request Fails because of casing issue in XML body
I include the fix for this towards the bottom -- very simple.

**Describe the bug**
When sending a WS-Trust request to some 3rd-Party Federations, message fails because of improper casing in MessageID

**To Reproduce**
(You are not expected to do this.)
Set up a WS-Trust relationship with a product like NetIQ Access Manager and then use MSAL to authenticate using a method like "acquire_token_by_username_password"

Error returned from Access Manager is:
"A required header representing a Message Addressing Property is not present, Problem header:{http://www.w3.org/2005/08/addressing}MessageID"

To Fix: On line 82 of wtrust_request.py , 
change:
<wsa:messageID>urn:uuid:{message_id}</wsa:messageID>
to:
<wsa:MessageID>urn:uuid:{message_id}</wsa:MessageID>
(i.e. just capitalize the "m" in "MessageID")

1.4.1

**Additional context**
The error is actually happening in Sun's XML parsing libraries, so other products doing WS-Trust in Java may be affected.

abhidnya13:
Closes #228

### DIFF
--- a/msal/wstrust_request.py
+++ b/msal/wstrust_request.py
@@ -79,7 +79,7 @@ def _build_rst(username, password, cloud_audience_urn, endpoint_address, soap_ac
     return """<s:Envelope xmlns:s='{s}' xmlns:wsa='{wsa}' xmlns:wsu='{wsu}'>
         <s:Header>
             <wsa:Action s:mustUnderstand='1'>{soap_action}</wsa:Action>
-            <wsa:messageID>urn:uuid:{message_id}</wsa:messageID>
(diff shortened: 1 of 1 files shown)


abhidnya13:
Also verified it from from middleware team that Brent's stuff uses MessageId (with capital M) and is following https://www.w3.org/2005/08/addressing/ which mentions the same. Even the older spec has capital M :http://schemas.xmlsoap.org/ws/2004/08/addressing/. Merging this now

PR #240, opened 2020-08-06
Client capabilities

### LINKED ISSUE #174: [Feature Request] Client capabilities support
Smart Session Management (SSM) sometimes referred to as the long lived tokens (LLT) scenario require MSAL to allow developers to resolve challenges (formatted as claims request parameters) to be resolved silently rather than interactively.  Current releases of MSAL assume that interaction is required whenever a claims request parameter is included in the request.  

This change in behavior requires the client to indicate that it's aware of a new calling pattern relative to the use of MSAL.

See additional details of requirements in:
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1545

API/Design details:
https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=%2FClientCapabilities%2Foverview.md&version=GBdev
**Is your feature request related to a problem? Please describe.**

Related internal work item:
https://identitydivision.visualstudio.com/Engineering/_workitems/edit/824806

abhidnya13:
<sub>Linking back to relevant issue (on behalf of Abhi ^_^) for easier cross-check during reviewing</sub> This will close #174

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -97,7 +97,8 @@ class ClientApplication(object):
             token_cache=None,
             http_client=None,
             verify=True, proxies=None, timeout=None,
-            client_claims=None, client_capabilities=None, app_name=None, app_version=None):
+            client_claims=None, app_name=None, app_version=None,
+            client_capabilities=None):
         """Create an instance of application.
 
         :param str client_id: Your app has a client_id after you register it on AAD.
@@ -146,11 +147,6 @@ class ClientApplication(object):
                     "jti": a_random_uuid
                 }
 
-        :param list[str] client_capabilities
-            Allows configuration of one or more client capabilities, e.g. "llt"
-            MSAL will transform these into special claims request.
-            See https://openid.net/specs/openid-connect-core-1_0-final.html#ClaimsParameter for
-            details on claim requests.
         :param str authority:
             A URL that identifies a token authority. It should be of the format
             https://login.microsoftonline.com/your_tenant
@@ -184,15 +180,16 @@ class ClientApplication(object):
         :param app_version: (optional)
             You can provide your application version for Microsoft telemetry purposes.
             Default value is None, means it will not be passed to Microsoft.
+        :param list[str] client_capabilities: (optional)
+            Allows configuration of one or more client capabilities, e.g. ["CP1"].
+            MSAL will combine them into
+            `claims parameter <https://openid.net/specs/openid-connect-core-1_0-final.html#ClaimsParameter`_
+            which you will later provide via one of the acquire-token request.
         """
         self.client_id = client_id
         self.client_credential = client_credential
         self.client_claims = client_claims
-        self.client_capabilities = {
-                "accessToken": {
-                    "xms_cc": client_capabilities
-                }
-            } if client_capabilities else None
+        self.client_capabilities = client_capabilities
         if http_client:
             self.http_client = http_client
         else:
@@ -226,8 +223,6 @@ class ClientApplication(object):
         if self.app_version:
             default_headers['x-app-ver'] = self.app_version
         default_body = {"client_info": 1}
-        if self.client_capabilities:
-            default_body["claims"] = self.client_capabilities
         if isinstance(client_credential, dict):
             assert ("private_key" in client_credential
                     and "thumbprint" in client_credential)
@@ -261,6 +256,19 @@ class ClientApplication(object):
             on_removing_rt=self.token_cache.remove_rt,
             on_updating_rt=self.token_cache.update_rt)
 
+    def _merge_claims_capabilities(self, claims):
+        client_capabilities_json = {}
+        if self.client_capabilities:
+            client_capabilities_json = {
+                "access_token": {
+                    "xms_cc": self.client_capabilities
+                }
+            }
+        for key in claims:
+            claims.update(client_capabilities_json.get(key))
+
+        return claims
+
     def get_authorization_request_url(
             self,
             scopes,  # type: list[str]
@@ -272,6 +280,7 @@ class ClientApplication(object):
             prompt=None,
             nonce=None,
             domain_hint=None,  # type: Optional[str]
+            claims=None,  # type: Optional[dict]
             **kwargs):
         """Constructs a URL for you to start a Authorization Code Grant.
 
@@ -300,6 +309,10 @@ class ClientApplication(object):
             More information on possible values
             `here <https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow#request-an-authorization-code>`_ and
             `here <https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-oapx/86fb452d-e34a-494e-ac61-e526e263b6d8>`_.
+        :param claims:
+             The claims Authentication Request parameter requests that specific Claims be returned
+             from the UserInfo Endpoint and/or in the ID Token and/or Access Token.
+             It is represented as a JSON object containing lists of Claims being requested from these locations.
         :return: The authorization url as a string.
         """
         """ # TBD: this would only be meaningful in a new acquire_token_interactive()
@@ -321,6 +334,7 @@ class ClientApplication(object):
             self.http_client
             ) if authority else self.authority
 
+        claims = self._merge_claims_capabilities(claims)
         client = Client(
             {"authorization_endpoint": the_authority.authorization_endpoint},
             self.client_id,
@@ -332,7 +346,7 @@ class ClientApplication(object):
             scope=decorate_scope(scopes, self.client_id),
             nonce=nonce,
             domain_hint=domain_hint,
-            claims=self.client_capabilities
+            claims=claims
             )
 
     def acquire_token_by_authorization_code(
@@ -344,6 +358,7 @@ class ClientApplication(object):
                 # authorization request as described in Section 4.1.1, and their
                 # values MUST be identical.
             nonce=None,
+            claims=None,
             **kwargs):
         """The second half of the Authorization Code Grant.
 
@@ -369,6 +384,11 @@ class ClientApplication(object):
             same nonce should also be provided here, so that we'll validate it.
             An exception will be raised if the nonce in id token mismatches.
 
+        :param claims:
+            The claims Authentication Request parameter requests that specific Claims be returned
+            from the UserInfo Endpoint and/or in the ID Token and/or Access Token.
+            It is represented as a JSON object containing lists of Claims being requested from these locations.
+
         :return: A dict representing the json response from AAD:
 
             - A successful response would contain "access_token" key,
@@ -389,6 +409,7 @@ class ClientApplication(object):
                 CLIENT_CURRENT_TELEMETRY: _build_current_telemetry_request_header(
                     self.ACQUIRE_TOKEN_BY_AUTHORIZATION_CODE_ID),
                 },
+            data=dict(kwargs.pop("data", {}), claims=claims),
             nonce=nonce,
             **kwargs)
 
@@ -792,7 +813,7 @@ class PublicClientApplication(ClientApplication):  # browser app or mobile app
         flow[self.DEVICE_FLOW_CORRELATION_ID] = correlation_id
         return flow
 
-    def acquire_token_by_device_flow(self, flow, **kwargs):
+    def acquire_token_by_device_flow(self, flow, claims=None, **kwargs):
         """Obtain token by a device flow object, with customizable polling effect.
 
         :param dict flow:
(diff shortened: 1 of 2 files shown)


tests/test_application.py
```diff
@@ -267,3 +267,42 @@ def helper(scopes, account, authority, *args, **kwargs):
...
+        self.authority_url_in_app = "https://login.microsoftonline.com/common"
+        self.scopes = ["s1", "s2"]
+        self.client_id = "my_app"
+        self.claims = "{\"id_token\": {\"auth_time\": {\"essential\": true}}}"
+
+    def test_merged_claims_returned_correctly(self):
+        app = ClientApplication(
+            self.client_id,
+            authority=self.authority_url_in_app, client_capabilities=["llt", "ssm"])
+        merged_claims = "{\"id_token\": {\"auth_time\": {\"essential\": true}}, \"access_token\": {\"xms_cc\": {\"values\": [\"llt\", \"ssm\"]}}}"
```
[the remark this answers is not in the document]
Avery-Dunn: My Python is rusty so I could be reading it wrong, but the format of the merged_claims value here looks correct, based on the values of client_capabilities and self.claims that I think get used in the test.

Check out the Claims/ClientCapabilities/ClientCapabilitiesAndClaimsJson values in [this constants file](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/master/tests/Microsoft.Identity.Test.Common/TestConstants.cs) of the .NET tests, or the very bottom of [this test](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/dev/src/test/java/com/microsoft/aad/msal4j/AuthorizationRequestUrlParametersTest.java) in the Java tests for other examples of the claims/client capabilities input and the expected output of merging the two together.

msal/application.py
```diff
@@ -249,6 +256,29 @@ def _build_client(self, client_credential, authority):
             on_removing_rt=self.token_cache.remove_rt,
             on_updating_rt=self.token_cache.update_rt)
 
+    def _merge_claims_and_capabilities(self, claims):
+        if claims:
+            claims = json.loads(claims)
```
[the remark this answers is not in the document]
abhidnya13: All of the other SDKs take in claims as a json string. Line 299 should have been updated but regardless, we need to decide on what we want to take in as an input. I don't have a strong preference but if we want to be consistent across SDKs we can take it in as a string

tests/test_application.py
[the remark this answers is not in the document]
abhidnya13: Yes I referred to these test cases as well, however both of these don not add  `xms_cc` part as a value in id_token field, I might be reading it wrong but I see in the above constants you mentioned we end up adding it only to `access_token` . However, as @rayluo  mentioned, in the [SSM](https://microsoft.sharepoint-df.com/:w:/t/SmartSessionManagementforOfficeClients/EU-SBJAwblNIlqlRwSFFDGUBbZng0DvLJAKNNB8Wy8hBKA?e=bKoiwM) doc mentioned it does say if you need both id_token and the access_token it needs to be added in both of these values. But from the client_capabilities [spec](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=%2FClientCapabilities%2Foverview.md&version=GBdev&_a=preview) we only need to add it as an access_token claim so this probably should be clarified. But if you do follow client capabilities spec, it only relates xm_cc as access_token claim

msal/application.py
```diff
@@ -79,6 +79,30 @@ def extract_certs(public_cert_content):
...
+            for key in client_capabilities_dict:
+                claims_node = claims.get(key, {})
+                if claims_node:
+                    claims_node.update(client_capabilities_dict[key])
+                else:
+                    claims[key] = client_capabilities_dict[key]
+
+        else:
+            return json.dumps(client_capabilities_dict)
+    return json.dumps(claims)
```
rayluo, line 103: Currently, if claims input is `None`, the output would become `null`. This might not be what we want?

Also, we should have a test case for such edge case.

msal/application.py
```diff
@@ -309,6 +346,7 @@ def get_authorization_request_url(
             self.http_client
             ) if authority else self.authority
 
+        claims = _merge_claims_and_capabilities(self.client_capabilities, claims)
```
rayluo, line 349: Assigning a different value into the same input parameter is error-prone. I would suggest to either use a different variable name:

```suggestion
        _claims = _merge_claims_and_capabilities(self.client_capabilities, claims)
```

or see the comment below.

msal/application.py
```diff
@@ -320,6 +358,7 @@ def get_authorization_request_url(
             scope=decorate_scope(scopes, self.client_id),
             nonce=nonce,
             domain_hint=domain_hint,
+            claims=claims,
```
rayluo, line 361: (continue from above)
or we can do it inline:

```suggestion
            claims=_merge_claims_and_capabilities(self.client_capabilities, claims),
```

msal/application.py
```diff
@@ -376,6 +421,7 @@ def acquire_token_by_authorization_code(
                 CLIENT_CURRENT_TELEMETRY: _build_current_telemetry_request_header(
                     self.ACQUIRE_TOKEN_BY_AUTHORIZATION_CODE_ID),
                 },
+            data=dict(kwargs.pop("data", {}), claims=claims),
```
rayluo, line 424: Did we forget to merge claims beforehand?

msal/application.py
```diff
@@ -830,16 +885,17 @@ def acquire_token_by_username_password(
...
             }
+        claims = _merge_claims_and_capabilities(self.client_capabilities, claims)
         if not self.authority.is_adfs:
             user_realm_result = self.authority.user_realm_discovery(
                 username, correlation_id=headers[CLIENT_REQUEST_ID])
             if user_realm_result.get("account_type") == "Federated":
                 return self._acquire_token_by_username_password_federated(
                     user_realm_result, username, password, scopes=scopes,
-                    headers=headers, **kwargs)
+                    headers=headers, data=dict(kwargs.pop("data", {}), claims=claims), **kwargs)
```
rayluo, line 895: The new addition is already significantly long enough to occupy a line. We do not squeeze it into exist line, then. (Same suggestion on elsewhere, too.)
```suggestion
                    data=dict(kwargs.pop("data", {}), claims=claims),
                    headers=headers, **kwargs)
```

tests/test_application.py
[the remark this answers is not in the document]
rayluo: Documenting current decision: based on an offline mail thread with stakeholders, this topic is not yet fully settled. In such case, we will go with providing this Client Capability behavior for access_token only, because this gives us more flexibility when/if we will also introduce this to id_token at a later time.

tests/test_application.py
```diff
@@ -267,3 +268,31 @@ def helper(scopes, account, authority, *args, **kwargs):
         with self.assertRaises(ExpectedBehavior):
             self.app.acquire_token_silent(["different scope"], self.account)
 
+
+class TestApplicationForClientCapabilities(unittest.TestCase):
+
+    def test_merged_claims_returned_correctly(self):
```
rayluo, line 274: I believe our org encourages a more informational test method naming pattern "condition_a_b_c_should_result(...)" e.g. "test_divide_by_zero_should_raise_exception()". We might not always enforce that, but the current name is just too vague to differentiate it from its cohort. :-)
```suggestion
    def test_capability_and_id_token_claims(self):
```

tests/test_application.py
```diff
@@ -267,3 +268,31 @@ def helper(scopes, account, authority, *args, **kwargs):
...
+        client_capabilities = ["llt", "ssm"]
+        claims = "{\"id_token\": {\"auth_time\": {\"essential\": true}}}"
```
rayluo, line 276: Perhaps it is time to utilize Python's triple-quotation string? You can also use it to break down those long lines.
```suggestion
        claims = '''{"id_token": {"auth_time": {"essential": true}}}'''
```

msal/application.py
```diff
@@ -79,6 +79,37 @@ def extract_certs(public_cert_content):
...
+    elif capabilities:
+        client_capabilities_dict = {
+            "access_token": {
+                "xms_cc": {
+                    "values": capabilities
+                }
+            }
+        }
+        return json.dumps(client_capabilities_dict)
+    return None
```
rayluo, line 110: Utilize the full power of Python's dict to cut 30-ish lines into 6. :-)
```suggestion
def _merge_claims_and_capabilities(capabilities, claims):
    # Represent capabilities as {"access_token": {"xms_cc": {"values": capabilities}}}
    # and then merge/add it into incoming claims
    if capabilities:
        claims_dict = json.loads(claims) if claims else {}
        for key in ["access_token"]:  # We could add "id_token" if we'd decide to
            claims_dict.setdefault(key, {}).update(xms_cc={"values": capabilities})
        return json.dumps(claims_dict)
    return claims
```

msal/application.py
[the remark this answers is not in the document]
rayluo: Since we now decide to actually go with string, let's document so in each of the APIs.

msal/application.py
```diff
@@ -795,7 +835,10 @@ def acquire_token_by_device_flow(self, flow, **kwargs):
...
         return self.client.obtain_token_by_device_flow(
             flow,
-            data=dict(kwargs.pop("data", {}), code=flow["device_code"]),
+            data=dict(
+                kwargs.pop("data", {}),
+                code=flow["device_code"],
+                claims=_merge_claims_and_capabilities(self._client_capabilities, claims)),
                 # 2018-10-4 Hack:
                 # during transition period,
                 # service seemingly need both device_code and code parameter.
```
rayluo, line 844: My bad. Did not realize the previous change split this comment with its corresponding code
```suggestion
                code=flow["device_code"],  # 2018-10-4 Hack:
                    # during transition period,
                    # service seemingly need both device_code and code parameter.
                claims=_merge_claims_and_capabilities(self._client_capabilities, claims),
                ),
```

tests/test_application.py
```diff
@@ -267,3 +268,41 @@ def helper(scopes, account, authority, *args, **kwargs):
...
+                        "access_token": {"nbf": {"essential": true, "value": "1563308371"},
+                                        "xms_cc": {"values": ["llt", "ssm"]}}}'''
+        # Comparing  dictionaries as JSON object order differs based on python version
+        self.assertEqual(
+            json.loads(merged_claims),
+            json.loads(_merge_claims_and_capabilities(client_capabilities, claims)))
+
+    def test_no_capabilities_only_claims_merge(self):
+        claims = '''{"id_token": {"auth_time": {"essential": true}}}'''
+        self.assertEqual(claims, _merge_claims_and_capabilities(None, claims))
```
rayluo, line 298: The existing assertion is logically wrong. Let's be consistent to use your `json.loads(...)` pattern (which is great! Thanks!).
```suggestion
        self.assertEqual(
            json.loads(...),
            json.loads(_merge_claims_and_capabilities(...))
```

msal/application.py
```diff
@@ -79,6 +79,16 @@ def extract_certs(public_cert_content):
...
 
+def _merge_claims_and_capabilities(capabilities, claims):
+    # Represent capabilities as {"access_token": {"xms_cc": {"values": capabilities}}}
+    # and then merge/add it into incoming claims
+    if capabilities:
+        claims_dict = json.loads(claims) if claims else {}
+        for key in ["access_token"]:  # We could add "id_token" if we'd decide to
+            claims_dict.setdefault(key, {}).update(xms_cc={"values": capabilities})
+        return json.dumps(claims_dict)
+    return claims
```
rayluo, line 90: Less indentation is probably slightly better
```suggestion
    if not capabilities:
        return claims
    claims_dict = json.loads(claims) if claims else {}
    for key in ["access_token"]:  # We could add "id_token" if we'd decide to
        claims_dict.setdefault(key, {}).update(xms_cc={"values": capabilities})
    return json.dumps(claims_dict)
```

abhidnya13:
As per internal discussion, changing the `claims` parameter to `claims_challenge` in the latest commit.

msal/application.py
```diff
@@ -917,6 +1005,11 @@ def acquire_token_on_behalf_of(self, user_assertion, scopes, **kwargs):
 
         :param str user_assertion: The incoming token already received by this app
         :param list[str] scopes: Scopes required by downstream API (a resource).
+        :param claims_challenge:
+            The claims_challenge parameter requests specific claims requested by the resource provider
+            in the form of a claims_challenge directive in the www-authenticate header to be
+            returned from the UserInfo Endpoint and/or in the ID Token and/or Access Token.
+            It is a string of a JSON object which contains lists of claims being requested from these locations..
```
jiasli, line 1012: Duplicate periods (.).

tests/test_application.py
```diff
@@ -267,3 +268,45 @@ def helper(scopes, account, authority, *args, **kwargs):
         with self.assertRaises(ExpectedBehavior):
             self.app.acquire_token_silent(["different scope"], self.account)
 
+
+class TestApplicationForClientCapabilities(unittest.TestCase):
+
+    def test_capabilities_and_id_token_claims_merge(self):
+        client_capabilities = ["llt", "ssm"]
```
jiasli, line 275: Are `llt` and `ssm` mocked-up values? Per the [doc](https://microsoft.sharepoint.com/:w:/t/aad/protocols/ERSWYtOQB45GgG4e2q3Cz00B6C36zi4gAs6JhdQG_wvVeQ?rtime=yEUj_nvd1kg&wdLOR=cB0CF5D1C-739A-4A91-983A-DBBDB9E775A6):

> The currently known values are: CP1.

This may confuse the readers.

msal/application.py
rayluo: @abhidnya13 Jiashuo has sharp eyes... :-)

tests/test_application.py
rayluo: Good point. Perhaps we should have changed them into "foo", "bar". @abhidnya13

msal/application.py
abhidnya13: He does :-) Jiashuo,  I have created a [new PR](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/250/files) addressing this and the other concern you have, please take a look :-) For some reason, I cannot add you as a reviewer.

PR #249, opened 2020-08-31
Permit response_mode in get_authorization_request_url

blokje:
Adds support for response_mode in get_authorization_request_url in order to override the default value set by the authentication server. For example this allows having form_post response in OpenID connect instead of the default fragment.

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -258,6 +258,7 @@ class ClientApplication(object):
             state=None,  # Recommended by OAuth2 for CSRF protection
             redirect_uri=None,
             response_type="code",  # Can be "token" if you use Implicit Grant
+            response_mode=None, # Can be "form_post", "fragment" or "query"
             prompt=None,
             nonce=None,
             domain_hint=None,  # type: Optional[str]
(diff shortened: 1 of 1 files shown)


rayluo:
Hi Rick, thanks for your input here. While the change seems simple, we want to better understand your scenario.

I believe you were using `response_type="token"` Implicit Flow (which you referred to as "OpenID Connect") and then its default `response_mode="fragment"` behavior is inconvenient for your Python web app. Therefore you would like to have such change.

But the industry is [moving away from Implicit Flow, and change to use Auth Code Flow (i.e. `resposne="code"`), for security reasons](https://stackoverflow.com/questions/13387698/why-is-there-an-authorization-code-flow-in-oauth2-when-implicit-flow-works-s).

If at all possible, we would suggest you to stick with MSAL's default `response_type="code"` and then you likely do not need to customize its default behavior at all. We have an existing [web app sample](https://github.com/Azure-Samples/ms-identity-python-webapp) for that.

blokje:
Hi Rayluo,

You are absolutely right and we seemed to. be stuck on some older legacy part of our code. We will take a look if we can integrate the Auth Code Flow. And close this PR as this not the road MSAL is going anyway :-)

Thanks for your feedback and prompt response.

PR #250, opened 2020-09-01
Changes in client capabilities as per AzCLI feedback

abhidnya13:
- Removing an extra period
- Using `foo` and `bar` as mock client capability values to avoid confusion for developers.


jiasli (approved):
Looks great!

yonzhan:
@abhidnya13  Thanks a lot:)

Issue #242, opened 2020-08-19
Device code endpoint malformed for some authorities


chlowell:
`ClientApplication._build_client` forms the device code endpoint with `urljoin`:

https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/f285074d6d5700457e7e80eed79560ab86c77a9a/msal/application.py#L234-L238

This assumes `authority.token_endpoint` does not have a trailing slash. The URL is incorrect when it does:

```
>>> urljoin("https://foo.bar/token", "devicecode")
'https://foo.bar/devicecode'
>>> urljoin("https://foo.bar/token/", "devicecode")
'https://foo.bar/token/devicecode'
```

`token_endpoint` comes directly from the authority's openid metadata:
https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/f285074d6d5700457e7e80eed79560ab86c77a9a/msal/authority.py#L94

So the upshot is that should an authority give its token endpoint with a trailing slash, as does my Azure Stack's authority, `ClientApplication.intiate_device_flow` will fail.

rayluo:
Historically, there was no "device_authorization_endpoint" available from tenant discovery, so we had to hardcode such fragile logic. Now it seems AAD provides that info (see [this example](https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration)).

I do not currently have an Azure Stack environment. @chlowell can you paste its openid configuration output, to let us see whether it also contain the device_authorization_endpoint?

rayluo:
@chlowell confirmed offline that "both the Stack and ADFS servers I'm working with return device_authorization_endpoint".

@abhidnya13 let's refactor this to use the device endpoint returned from the discovery, and still fall back to the existing logic when/if that info is not available.

Issue #211, opened 2020-06-23
Need to investigate some sign-in error during test automation


rayluo:
I happen to notice 3+ such error message in our test automation logs:

    AADSTS50053: The account is locked, you've tried to sign in too many times with an incorrect user ID or password.

For example, [this one](https://travis-ci.org/github/AzureAD/microsoft-authentication-library-for-python/jobs/701078031#L2162). While this is not an issue to this MSAL Python library, we might want to investigate to see whether our testing account is still up-to-date.

UPDATE at 7-21: Same error messages can still be found in our latest test automation logs, such as [this one](https://travis-ci.org/github/AzureAD/microsoft-authentication-library-for-python/jobs/710194383).

PR #252, opened 2020-09-03
MSAL Python 1.5.0

abhidnya13:
- Added support for setting [client capabilities](https://github.com/AzureAD/microsoft-authentication-library-for-python/wiki/Client-capabilities) to enable CAE(Continuous Access Evaluation) (#240, #174)
- Device code endpoint is now fetched from open-id configuration, if available. (#245, #242)
- Fixes in test cases (#239, #211)


rayluo (approved):
SSM has been re-branded to CAE. Do not use that old name in our release note. Other than that, everything looks good. Providing an approval in advance.

PR #260, opened 2020-10-06
Acquire token interactive  using system browser

abhidnya13:
Original goal:
- Adding state ~and pkce~
- ~Can we use a custom browser~
- Cleaning up the code and structure
- Reference docs

---

Latest Update:

The latest commit implements:
- `state`, `nonce` and `PKCE` protection are already implemented by recent refactoring. This PR is built on top of that, thus including these security behavior. This PR focus on the new `acquire_token_interactive()` API.
- Reference docs is available [in staging environment](https://msal-python.readthedocs.io/en/docs-staging/#msal.PublicClientApplication.acquire_token_interactive), and will show up [in production](https://msal-python.readthedocs.io/en/latest/#msal.PublicClientApplication.acquire_token_interactive) once this PR is merged.
- Telemetry with API ID (the new API ID is also updated into the [internal Telemetry table](https://microsoft-my.sharepoint-df.com/:x:/p/sagonzal/EXSrr4vM1utAqQQfD6bMln4BYKwjrqh3cagiNJWPVNjLzw?e=G5FybL)
- We will work out a sample at a later time

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -107,6 +107,7 @@ class ClientApplication(object):
     ACQUIRE_TOKEN_BY_DEVICE_FLOW_ID = "622"
     ACQUIRE_TOKEN_FOR_CLIENT_ID = "730"
     ACQUIRE_TOKEN_BY_AUTHORIZATION_CODE_ID = "832"
+    ACQUIRE_TOKEN_INTERACTIVE = "169"
     GET_ACCOUNTS_ID = "902"
     REMOVE_ACCOUNT_ID = "903"
 
@@ -318,7 +319,6 @@ class ClientApplication(object):
 
         :param list scope:
             It is a list of case-sensitive strings.
-            Some ID provider can accept empty string to represent default scope.
         :param str redirect_uri:
             Optional. If not specified, server will use the pre-registered one.
         :param str state:
@@ -997,6 +997,78 @@ class PublicClientApplication(ClientApplication):  # browser app or mobile app
             raise ValueError("Public Client should not possess credentials")
         super(PublicClientApplication, self).__init__(
             client_id, client_credential=None, **kwargs)
+
+    def acquire_token_interactive(
+            self,
+            scopes,  # type: list[str]
+            prompt=None,
+            login_hint=None,  # type: Optional[str]
+            domain_hint=None,  # type: Optional[str]
+            claims_challenge=None,
+            timeout=None,
+            port=None,
+            **kwargs):
+        """Acquire token interactively i.e. via a local browser.
+
+        :param list scope:
+            It is a list of case-sensitive strings.
+        :param str prompt:
+            By default, no prompt value will be sent, not even "none".
+            You will have to specify a value explicitly.
+            Its valid values are defined in Open ID Connect specs
+            https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest
+        :param str login_hint:
+            Optional. Identifier of the user. Generally a User Principal Name (UPN).
+        :param domain_hint:
+            Can be one of "consumers" or "organizations" or your tenant domain "contoso.com".
+            If included, it will skip the email-based discovery process that user goes
+            through on the sign-in page, leading to a slightly more streamlined user experience.
+            More information on possible values
+            `here <https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow#request-an-authorization-code>`_ and
+            `here <https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-oapx/86fb452d-e34a-494e-ac61-e526e263b6d8>`_.
+
+        :param claims_challenge:
+            The claims_challenge parameter requests specific claims requested by the resource provider
+            in the form of a claims_challenge directive in the www-authenticate header to be
+            returned from the UserInfo Endpoint and/or in the ID Token and/or Access Token.
+            It is a string of a JSON object which contains lists of claims being requested from these locations.
+
+        :param int timeout:
+            This method will block the current thread.
+            This parameter specifies the timeout value in seconds.
+            Default value ``None`` means wait indefinitely.
+
+        :param int port:
+            The port to be used to listen to an incoming auth response.
+            By default we will use a system-allocated port.
+            (The rest of the redirect_uri is hard coded as ``http://localhost``.)
+
+        :return:
+            - A dict containing no "error" key,
+              and typically contains an "access_token" key,
+              if cache lookup succeeded.
+            - A dict containing an "error" key, when token refresh failed.
+        """
+        self._validate_ssh_cert_input_data(kwargs.get("data", {}))
+        claims = _merge_claims_challenge_and_capabilities(
+            self._client_capabilities, claims_challenge)
+        return self.client.obtain_token_by_browser(
+            scope=decorate_scope(scopes, self.client_id) if scopes else None,
+            redirect_uri="http://localhost:{port}".format(
+                # Hardcode the host, for now. AAD portal rejects 127.0.0.1 anyway
+                port=port or 0),
+            prompt=prompt,
+            login_hint=login_hint,
+            domain_hint=domain_hint,
+            timeout=timeout,
+            auth_params={"claims": claims},
+            data=dict(kwargs.pop("data", {}), claims=claims),
+            headers={
+                CLIENT_REQUEST_ID: _get_new_correlation_id(),
+                CLIENT_CURRENT_TELEMETRY: _build_current_telemetry_request_header(
+                    self.ACQUIRE_TOKEN_INTERACTIVE),
+                },
+            **kwargs)
 
     def initiate_device_flow(self, scopes=None, **kwargs):
         """Initiate a Device Flow instance,
--- a/tests/test_e2e.py
+++ b/tests/test_e2e.py
@@ -441,6 +441,39 @@ class LabBasedTestCase(E2eTestCase):
                 error_description=result.get("error_description")))
         self.assertCacheWorksForUser(result, scope, username=None)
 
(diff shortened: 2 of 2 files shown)


msal/application.py
```diff
@@ -823,6 +836,31 @@ def acquire_token_by_refresh_token(self, refresh_token, scopes):
...
 
+    def acquire_token_interactive(self,
+            scopes,  # type: list[str]
+            # additional_scope=None,  # type: Optional[list]
+            login_hint=None,  # type: Optional[str]
+            port=None,
+            prompt=None,
+            domain_hint=None,  # type: Optional[str]
+            claims_challenge=None,
+            **kwargs):
```
rayluo, line 847: Remove this param, for now.

msal/application.py
```diff
@@ -823,6 +836,31 @@ def acquire_token_by_refresh_token(self, refresh_token, scopes):
...
+            port = _get_open_port()
+        redirect_uri = "http://localhost:" + str(port)
+        auth_url = self.get_authorization_request_url(scopes=scopes,
+                                                      login_hint=login_hint,
+                                                      redirect_uri=redirect_uri,
+                                                      response_type="code",
+                                                      prompt=prompt,
+                                                      domain_hint=domain_hint,
+                                                      claims_challenge=claims_challenge,
+                                                      **kwargs)
```
rayluo, line 858:
```suggestion
        auth_url = self.get_authorization_request_url(
            scopes=scopes,
            login_hint=login_hint,
            redirect_uri="http://localhost:%d" % port,
            response_type="code",
            prompt=prompt,
            domain_hint=domain_hint,
            claims_challenge=claims_challenge,
            )
```

msal/application.py
```diff
@@ -59,6 +62,16 @@ def _get_new_correlation_id():
     return str(uuid.uuid4())
 
 
+def _get_open_port():
+    import socket
+    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    s.bind(("", 0))
+    s.listen(1)
```
rayluo, line 69: There are various implementations out there. Some of them do not contain this line. Any investigation on whether that is necessary or not?

msal/application.py
rayluo: Given that we do not handle `**kwargs` in the implementation, I mean to say let's remove this `**kwargs` from this method's signature.

msal/application.py
```diff
@@ -306,6 +309,8 @@ def get_authorization_request_url(
             nonce=None,
             domain_hint=None,  # type: Optional[str]
             claims_challenge=None,
+            code_challenge_method=None,
+            code_challenge=None,
```
rayluo, line 313: `acquire_token_interactive` and PKCE are two different features. I would suggest we separate them into 2 different PRs, so that their code review, subsequent wikis/samples (if any), and even future conversation reference ... would all be self-contained.

We may then choose to get the PKCE PR finished first.

msal/application.py
abhidnya13: Yes, I have already removed it in the latest commit

msal/application.py
abhidnya13: I think PKCE can be considered as a sub feature of acquire_token_interactive since PKCE as a whole involves two steps, one is generating `code_verifier` and `code_challenge`  and then sending `code_challenge` in step 1 to /oauth2/authorize request and the sending `code_verifier` in part 2 which is acquire_token_by_authorization code. But if we want to separate it into 2 PRs for the ease of code review, I can separate PKCE API support into a different PR 

However utilizing this PKCE feature(generating the code_verifier and code_challenge and then sending it in two steps) will still be in this PR. The actual support to be able to pass such parameters to the functions will be added once the PKCE PR is merged. 

PS: To be honest these changes were made to also get an E2E test working with the code_verifier and code_challenge being actually sent, how does it look like when a wrong code_verifier is sent, what error is encountered etc

rayluo: Your "PS" part is exactly the reason why the PKCE part of implementation, test, review, etc. should be in its own PR. Once that part is done, this acquire_token_interactive() can simply use it, without needing to test it.

Once you open a PKCE PR on the existing auth code flow, we will be get that reviewed and merged quickly.

msal/application.py
```diff
@@ -852,8 +852,7 @@ def acquire_token_interactive(self,
             port=None,
             prompt=None,
             domain_hint=None,  # type: Optional[str]
-            claims_challenge=None,
-            **kwargs):
+            claims_challenge=None):
```
rayluo, line 855: Minor thing: In case you don't already know, this can be written in the following way, so that when we eventually add that `**kwargs` (or whatever new parameters), it will generate minimal red-and-green lines in the new PR.

```suggestion
            claims_challenge=None,
            ):
```

msal/application.py
rayluo: @abhidnya13 : I've adjusted some of the goals in this PR's description, to align with what we discussed last week. All those extra features you identified can be a good-to-have in near future, but they just do not have/need to be mixed up all together. Let's focus on the topic(s) fundamental to interactive auth, and get them done.

abhidnya13: I agree that we should have a fundamental interactive auth PR and that should be this one. I still think PKCE should be included here too instead of taking it up later, I agree though having it in a different PR makes more sense. And the last part about customizing such a behavior is good-to-have in near future. That near future can be soon if we agree on what we want to include as I have spent enough time looking into the possibility of customizing, I do not want to leave it here. I agree though that, it should not be a part of this PR and should not block us from getting a fundamental interactive auth up and running.

msal/application.py
```diff
@@ -823,6 +836,34 @@ def acquire_token_by_refresh_token(self, refresh_token, scopes):
...
+        request_state = str(uuid.uuid4())
+        auth_url = self.get_authorization_request_url(
+            scopes=scopes,
+            login_hint=login_hint,
+            redirect_uri="http://localhost:%d" % port,
+            response_type="code",
+            state=request_state,
+            prompt=prompt,
+            domain_hint=domain_hint,
+            claims_challenge=claims_challenge,)
```
henrik-me, line 859: >) [](start = 46, length = 1)

I think this one has to be on the next line according to Ray's comment

henrik-me (approved):
:shipit:

rayluo (commented):
* Adding 2 implementation-level comments below, however those part of code structure would likely be adjusted anyway, when you carry out the authcode.py refactor which you brought up days ago. Do you have any idea to discuss in our upcoming meeting?
* And, what is your thought on [this one](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/260#discussion_r500648798)?

msal/application.py
```diff
@@ -823,6 +836,34 @@ def acquire_token_by_refresh_token(self, refresh_token, scopes):
...
+        auth_code, received_state = obtain_auth_code(port, auth_uri=auth_url)
+        if request_state != received_state:
```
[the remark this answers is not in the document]
abhidnya13: have some thoughts around this, would like to sync up. Making the change should not be a big part but have some concerns that I wanted to discuss.

msal/application.py
abhidnya13: s.listen() will make the socket open for accepting connections. 1 is the value of `backlog`. So, s.listen(1) means that it is accepting one simultaneous connection. This part of code is inspired from [here](https://stackoverflow.com/questions/2838244/get-open-tcp-port-in-python/2838309#2838309).

msal/oauth2cli/authcode.py
```diff
@@ -36,25 +36,11 @@ def obtain_auth_code(listen_port, auth_uri=None):
     :param auth_uri: If provided, this function will try to open a local browser.
     :return: Hang indefinitely, until it receives and then return the auth code.
     """
-    exit_hint = "Visit http://localhost:{p}?code=exit to abort".format(p=listen_port)
-    logger.warning(exit_hint)
     if auth_uri:
-        page = "http://localhost:{p}?{q}".format(p=listen_port, q=urlencode({
```
rayluo, line 42: The latest commit(s) move this part to a test case. Is it possible to still have this functionality here? We may be able to have the presence or absence of a new parameter `obtain_auth_code(..., text="Open this link to sign in. You may use incognito window")` to toggle this test functionality.

msal/oauth2cli/authcode.py
```diff
@@ -69,14 +55,16 @@ def browse(auth_uri):
...
 class AuthCodeReceiver(BaseHTTPRequestHandler):
     def do_GET(self):
         # For flexibility, we choose to not check self.path matching redirect_uri
         #assert self.path.startswith('/THE_PATH_REGISTERED_BY_THE_APP')
         qs = parse_qs(urlparse(self.path).query)
         if qs.get('code'):  # Then store it into the server instance
-            ac = self.server.authcode = qs['code'][0]
-            self._send_full_response('Authcode:\n{}'.format(ac))
+            self.server.auth_code = qs['code'][0]
+            self.server.state = qs.get('state', [None])[0]
```
rayluo, line 66: Is it possible to have the initial state pre-polulated inside `self.server`, so that we can check the state here and only accept a `code` when the `state` matches?

msal/application.py
```diff
@@ -823,6 +836,37 @@ def acquire_token_by_refresh_token(self, refresh_token, scopes):
...
             )
 
+    def acquire_token_interactive(self,
+            scopes,  # type: list[str]
+            # additional_scope=None,  # type: Optional[list]
+            login_hint=None,  # type: Optional[str]
+            port=None,
+            prompt=None,
+            domain_hint=None,  # type: Optional[str]
+            claims_challenge=None,
```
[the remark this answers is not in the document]
abhidnya13: Do you mean `claims_challenge` ?  I have removed everything related to PKCE from here.

rayluo (approved):

PR #261, opened 2020-10-12
Adding API to support PKCE in interactive auth

### LINKED ISSUE #255: PKCE Support
**Description**
As of today, there does not seem to be a straightforward way to support [PKCE](https://tools.ietf.org/html/rfc7636) through the use of query parameters `code_challenge` and `code_challenge_method`. See "[Request an authorization code](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow#request-an-authorization-code)" and  "[Request an access token](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow#request-an-access-token)" for more information.

Is there some guideline of how to achieve this? Is there a work around?

`python -c "import msal; print(msal.__version__)"` gives 1.5.0

abhidnya13:
Sub-task of #260.

Resolves #255

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -293,6 +293,8 @@ class ClientApplication(object):
             nonce=None,
             domain_hint=None,  # type: Optional[str]
             claims_challenge=None,
+            code_challenge=None,
+            code_challenge_method=None,
             **kwargs):
         """Constructs a URL for you to start a Authorization Code Grant.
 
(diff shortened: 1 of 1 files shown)


henrik-me (approved):
:shipit:

rayluo:
Closing this one, as we now go with a different direction at #276 which is in better position to provide PKCE support by NOT providing PKCE parameters. :-)

PR #262, opened 2020-10-13
Do not add account when performing acquire_token_silent()

### LINKED ISSUE #258: Potential loss of sign-out
**Describe the bug**
Loss-of-signout could happen.

**To Reproduce**
We did not see this in the wild. But these are steps to possibly hit the issue.

1. app A and app B share same token cache (i.e. they belong to same family of apps)
2. app B calls `acquire_token_silent()` and is in the middle of waiting network I/O to complete
3. end user John Doe signs out in app A via `remove_account()` (and intend to lend the device to user Jane Doe)
4. app B finishes network I/O, and adds the same account into cache, thus ends up still keeping John Doe's account for both app A and app B.

**Expected behavior**
App B should honor the sign out operation from app A.

**What you see instead**
We did not see this bug in the wild. But these are steps to possibly hit the issue.

All.

**Additional context**
A solution: to enforce that "save token" operation is always "UPDATE" operation in context of "Silent call". And not "ADD" operation.

rayluo:
This will fix #258

### DIFF
--- a/msal/token_cache.py
+++ b/msal/token_cache.py
@@ -172,7 +172,7 @@ class TokenCache(object):
                     at["key_id"] = data.get("key_id")
                 self.modify(self.CredentialType.ACCESS_TOKEN, at, at)
 
-            if client_info:
+            if client_info and event.get("add_account") is not False:
                 account = {
                     "home_account_id": home_account_id,
                     "environment": environment,
--- a/msal/token_cache.py
+++ b/msal/token_cache.py
@@ -172,7 +172,7 @@ class TokenCache(object):
                     at["key_id"] = data.get("key_id")
                 self.modify(self.CredentialType.ACCESS_TOKEN, at, at)
 
-            if client_info and event.get("add_account") is not False:
+            if client_info and not event.get("skip_account_creation"):
                 account = {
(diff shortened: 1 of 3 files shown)


abhidnya13:
Hey Ray, I went through the changes and from what I understand is we skip adding an account in the case of an acquire token silent. So the only place where an account will be added is when we acquire token by rest of the methods.

msal/token_cache.py
```diff
@@ -172,7 +172,7 @@ def __add(self, event, now=None):
                     at["key_id"] = data.get("key_id")
                 self.modify(self.CredentialType.ACCESS_TOKEN, at, at)
 
-            if client_info:
+            if client_info and event.get("add_account") is not False:
```
abhidnya13, line 175: Is `is not False` necessary ?

rayluo: Unfortunately, it is currently necessary, because all other `acquire_token_by_xyz()` methods historically send no `add_account` flag at all, yet we want to treat such default case as "Yes we add an account".

That being said, you identified a bad judgement call I made here. Now I incline to change this flag's name to `skip_account`, so that we would be able to simply use `if client_info and not skip_account:` here.

PR #266, opened 2020-10-21
Release 1.5.1

rayluo:
* Bugfix: We now cache tokens by specified environment, not by OIDC Discovery. This won't matter most of the time, but it can be needed when your tenant is in transitional state while migrating to a different cloud. (#247)
* Bugfix: We now make sure one app's sign-out operation would be successful even when another app is acquiring token from cache at the same time. (#258, #262)

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -235,6 +235,7 @@ class ClientApplication(object):
             "authorization_endpoint": authority.authorization_endpoint,
             "token_endpoint": authority.token_endpoint,
             "device_authorization_endpoint":
+                authority.device_authorization_endpoint or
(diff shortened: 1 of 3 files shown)


abhidnya13 (approved):
🚢

cdeler:
Hello,
Do you know if we can include https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/267 into the release (if it's approved)?

rayluo:
@cdeler Thanks for your effort on that new PR. We will take a look there. It won't make it to this patch release 1.5.1 anyway, because your PR is conceptually a new feature. Our version number follows [Semantic Versioning](https://semver.org/).

PR #267, opened 2020-10-21
Added extra_headers to ClientApplication

cdeler:
Having enabled Auth Code Grant Flow for SPA in our project, we broke
Device Code flow for the Azure app. The only way to fix this error is to
pass an extra "Origin" header to this endpoint
    `https://login.microsoftonline.com/<tenant>/oauth2/v2.0/token`

After that the app initialisation might look like:
```python
application = msal.PublicClientApplication(
          client_id,
          authority=authority,
          extra_headers={"Origin": "http://example.com"})
flow = application.initiate_device_flow(scopes=scopes)
```

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -109,7 +109,7 @@ class ClientApplication(object):
             http_client=None,
             verify=True, proxies=None, timeout=None,
             client_claims=None, app_name=None, app_version=None,
-            client_capabilities=None):
+            client_capabilities=None, extra_headers=None):
         """Create an instance of application.
 
         :param str client_id: Your app has a client_id after you register it on AAD.
@@ -191,6 +191,11 @@ class ClientApplication(object):
         :param app_version: (optional)
             You can provide your application version for Microsoft telemetry purposes.
             Default value is None, means it will not be passed to Microsoft.
+        :param dict extra_headers: (optional)
+            Allows you to pass extra headers to each OAuth client request,
+            adding them to the client session. msal uses requests.Session to requesting,
+            so more information you can find here:
+            https://requests.readthedocs.io/en/master/user/advanced/#session-objects
         :param list[str] client_capabilities: (optional)
             Allows configuration of one or more client capabilities, e.g. ["CP1"].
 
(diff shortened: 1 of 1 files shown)


rayluo:
Hi @cdeler , thanks for bringing this to our attention. While your implementation in this PR looks straightforward, we want to better understand the background of this change, i.e. how to reproduce the issue you encountered.

* When you said you "enable Auth Code Grant Flow for SPA", did you mean you follow these ["Migrate javascript SPA from implicit grant to auth code flow" instructions](https://docs.microsoft.com/en-us/azure/active-directory/develop/migrate-spa-implicit-to-auth-code)?

* Was Device Flow the only flow which was broken? Do you happen to also use other flows with your app?

cdeler:
### The root cause

Hello @rayluo 

> When you said you "enable Auth Code Grant Flow for SPA", did you mean you follow these "Migrate javascript SPA from implicit grant to auth code flow" instructions?

Yes. Some context:
We have an application. We need to have an access to the application using both device flow & from web UI.

Updating Azure AD application to SPA in UI causes additional ("Origin") header requirement in requests to https://login.microsoftonline.com/<tenant>/oauth2/v2.0/token.

In both examples below I get the same error like "Tokens issued for the 'Single-Page Application' client-type may only be redeemed via cross-origin requests"

### Code examples:
From the python msal library, it means that we are unable to either run device flow or issue "access_token" using given "refresh_token":
Example 1
```python
assert isintance(application, msal.PublicClientApplication)
flow = application.initiate_device_flow(scopes=scopes)
if "user_code" not in flow:
    raise RuntimeException( f"ERROR, Unable to create device flow. Error: {pprint.pformat(flow)}")

print(flow["message"])
input("Press any key to continue...")
token = application.acquire_token_by_device_flow(flow)
# do something with token
```

Example 2:
```python
assert isintance(application, msal.PublicClientApplication)
refresh_token = {"credential_type": "RefreshToken", "refresh_token": "<put_your_refresh_token_here>"}
token = application.client.obtain_token_by_refresh_token(token, scope=scopes)
```

rayluo:
@cdeler , since this PR unblocks you, you can stick with it for a longer time. Meanwhile, our team is investigating this, to see why/whether this workaround is needed in the first place. We will reply back to this issue when new findings become available.

cdeler:
@rayluo thank you

I have a workaround, so I'll be waiting for updates :-)

rayluo:
@cdeler I'm told that the fix for this, has been rolled out on the service, so, your existing Device Flow app should now work, without needing the workaround. Please verify. This PR will soon be closed without merging.

cdeler:
I confirm that the device flow works without updating the headers.

Thank you 🎉 🚀

PR #269, opened 2020-10-22
Bubble http exceptions so apps could catch them

### LINKED ISSUE #263: Getting "Unable to get authority configuration" Error when Authority is properly sent
**Describe the bug**
We are using MSAL in production environment in out first party application.
we are using ConfidentialClientApplication with certificate.

in the logs I see this error popping up:
_"ValueError: Unable to get authority configuration for https://login.microsoftonline.com/<AAD context-id>. Authority would typically be in a format of https://login.microsoftonline.com/your_tenant_name"_
(I removed the AAD context id for privacy here)

for exactly the same context-id from AAD the same request with the same authority url can pass multiple times and then randomly fail with the same message.

**Expected behavior**
if this is an error that is not caused by the authority url input but by something else the message returned should be different.
we need to be able to differentiate the types of errors, so that if this is caused by something transient we can ignore it.

msal version is 1.5.0

rayluo:
This can likely close #263 .

@AlmogBentz You can install this branch for your smoke testing:

    pip install git+https://github.com/AzureAD/microsoft-authentication-library-for-python.git@differentiate-oidc-discovery-errors

After this patch, when the lower level network-relevant error (came from requests) occurs, they would be bubbled up as-is, therefore you won't see the "Authority would typically be in a format of ..." error message. You could catch them and treat them differently, therefore satisfying your desire to handle different exception differently.

### DIFF
--- a/msal/authority.py
+++ b/msal/authority.py
@@ -83,7 +83,7 @@ class Authority(object):
             openid_config = tenant_discovery(
                 tenant_discovery_endpoint,
                 self.http_client)
-        except ValueError:  # json.decoder.JSONDecodeError in Py3 subclasses this
+        except ValueError:
             raise ValueError(
                 "Unable to get authority configuration for {}. "
                 "Authority would typically be in a format of "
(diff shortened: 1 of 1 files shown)


rayluo:
Did not (yet?) hear back from @AlmogBentz . @abhidnya13 please help take a look.

msal/authority.py
```diff
@@ -140,8 +140,17 @@ def instance_discovery(url, http_client, **kwargs):
...
-    raise MsalServiceError(status_code=resp.status_code, **payload)
+    if resp.status_code == 200:
+        payload = json.loads(resp.text)  # It could raise ValueError
+        if 'authorization_endpoint' in payload and 'token_endpoint' in payload:
+            return payload  # Happy path
+        raise ValueError("OIDC Discovery does not provide enough information")
+    if 400 <= resp.status_code < 500:
+        # Nonexist tenant would hit this path
+        # e.g. https://login.microsoftonline.com/nonexist_tenant/v2.0/.well-known/openid-configuration
+        raise ValueError("OIDC Discovery endpoint rejects our request")
```
abhidnya13, line 151: We can be a little more verbose on what needs to be done by the app. If this is hit when non existing tenant is specified, we can say specify an action item on app developer based on what needs to be done. I am not sure if the app developer will always understand what a OIDC discovery endpoint is and how is it formed since its an internal MSAL Implementation detail

msal/authority.py
```diff
@@ -140,8 +140,17 @@ def instance_discovery(url, http_client, **kwargs):
...
+    # Transient network error would hit this path
+    resp.raise_for_status()
+    raise RuntimeError(  # A fallback here, in case resp.raise_for_status() is no-op
```
abhidnya13, line 154: Is there a reason to move away from `MsalServiceError` ?

abhidnya13 (changes requested):
Looks good to me overall, just some questions/suggestions around supportability !

msal/authority.py
rayluo: Good point. Actually, this exception will be [caught by ourselves at a different layer](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/269/files#diff-dea3df43cd06f684c7e95c0dcd265456fc0e2ed36cbdf7a72c6bc4dbd8af8f9aR86), and we provided information there.

msal/authority.py
rayluo: Per our customized http client expectation/requirement, `resp.raise_for_status()` should throw proper exception. The line here is really just a fallback. If it really hits to this line, it is conceptually a http client error, not a service error.

abhidnya13 (approved):
:shipit:

PR #271, opened 2020-11-01
Add support for acquiring a token with a pre-signed JWT

lochiiconnectivity:
JWT.

Useful for where the signing takes place externally for example using
Azure Key Vault (AKV).

AKV sample included.

This is the MSAL version of [This pull](https://github.com/AzureAD/azure-activedirectory-library-for-python/pull/196)

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -153,7 +153,7 @@ class ClientApplication(object):
 
             *Added in version 1.13.0*:
             It can also be a completly pre-signed assertion that you've assembled yourself.
-            Simply pass a container containing only the key "client_assertion", like this:
+            Simply pass a container containing only the key "client_assertion", like this::
 
                 {
                     "client_assertion": "...a JWT with claims aud, exp, iss, jti, nbf, and sub..."
@@ -1093,4 +1093,3 @@ class ConfidentialClientApplication(ClientApplication):  # server-side web app
                     self.ACQUIRE_TOKEN_ON_BEHALF_OF_ID),
                 },
             **kwargs)
-
--- a/msal/application.py
+++ b/msal/application.py
@@ -268,9 +268,9 @@ class ClientApplication(object):
                     and "thumbprint" in client_credential) or
                     "client_assertion" in client_credential)
             headers = {}
+            client_assertion_type = Client.CLIENT_ASSERTION_TYPE_JWT
             if 'client_assertion' in client_credential:
                 client_assertion = client_credential['client_assertion']
-                client_assertion_type = Client.CLIENT_ASSERTION_TYPE_JWT
             else:
                 if 'public_certificate' in client_credential:
                     headers["x5c"] = extract_certs(client_credential['public_certificate'])
@@ -290,7 +290,6 @@ class ClientApplication(object):
                 client_assertion = assertion.create_regenerative_assertion(
                     audience=authority.token_endpoint, issuer=self.client_id,
                     additional_claims=self.client_claims or {})
-                client_assertion_type = Client.CLIENT_ASSERTION_TYPE_JWT
         else:
             default_body['client_secret'] = client_credential
         server_configuration = {
--- a/msal/application.py
+++ b/msal/application.py
@@ -151,12 +151,12 @@ class ClientApplication(object):
             "The provided signature value did not match the expected signature value",
             you may try use only the leaf cert (in PEM/str format) instead.
 
-            **NEW**
-            it can also be a completly pre-signed JWT that you've assembled yourself
-            simply pass a container containing only the key "client_assertion", like this:
+            *Added in version 1.13.0*:
+            It can also be a completly pre-signed assertion that you've assembled yourself.
+            Simply pass a container containing only the key "client_assertion", like this:
 
                 {
-                    "client_assertion": "..."
+                    "client_assertion": "...a JWT with claims aud, exp, iss, jti, nbf, and sub..."
                 }
 
         :param dict client_claims:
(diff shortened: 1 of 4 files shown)


rayluo:
Nice to meet you, again, David! :-)  We will definitely take a look into this.

lochiiconnectivity:
**prod**

msal/application.py
```diff
@@ -151,6 +151,14 @@ def __init__(
             "The provided signature value did not match the expected signature value",
             you may try use only the leaf cert (in PEM/str format) instead.
 
+            **NEW**
+            it can also be a completly pre-signed JWT that you've assembled yourself
+            simply pass a container containing only the key "jwt", like this:
+
+                {
+                    "jwt": "..."
```
rayluo, line 159: Our API surface would prefer to use a name referring to the intention rather than the data format. Besides, MSAL .Net already supports [confidential client authenticated by a signed assertion](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/Client-Assertions#signed-assertions), with the API `.WithClientAssertion(...)`. So, we would suggest to change the parameter name of this feature from `jwt` to `client_assertion`. Would you be able to change that here and in those test cases in this PR?

sample/vault_jwt_sample.py
```diff
@@ -0,0 +1,134 @@
...
+# notice we give account parameter as None.
+result = app.acquire_token_silent(config["scope"], account=None)
+
+if not result:
+    logging.info("No suitable token exists in cache. Let's get a new one from AAD.")
+    result = app.acquire_token_for_client(scopes=config["scope"])
+
+if "access_token" in result:
+    # Calling graph using the access token
+    graph_data = requests.get(  # Use token to call downstream service
```
rayluo, line 126: This sample would work because it is intended to be used as a one-time command-line invocation. We can accept it as-is (after making those `jwt` to `client_assertion` changes.

Out of curiosity, is your real app in production a long-lived app? If so, you would encounter assertion expiration after 10 minutes, wouldn't you?

rayluo (changes requested):
Thank you, David, for your patience, and your comprehensive work on this PR! It is rare to see a contributor PR with this level of test coverage. You are awesome!

We added some comments below. If you can make some adjustment, that would be great.

msal/application.py
lochiiconnectivity: done.

sample/vault_jwt_sample.py
lochiiconnectivity: it's a short lived command line app, it shouldn't be running for more than ten minutes.

msal/application.py
```diff
@@ -256,28 +264,33 @@ def _build_client(self, client_credential, authority):
...
+                        _str2bytes(client_credential["passphrase"]),
+                        backend=default_backend(),  # It was a required param until 2020
+                        )
+                assertion = JwtAssertionCreator(
+                    unencrypted_private_key, algorithm="RS256",
+                    sha1_thumbprint=client_credential.get("thumbprint"), headers=headers)
+                client_assertion = assertion.create_regenerative_assertion(
+                    audience=authority.token_endpoint, issuer=self.client_id,
+                    additional_claims=self.client_claims or {})
+                client_assertion_type = Client.CLIENT_ASSERTION_TYPE_JWT
```
jiasli, line 293: No need to repeat this statement twice in both `if` and `else`.

jiasli:
PR tiles is malformed.

msal/application.py
lochiiconnectivity: done

rayluo:
Thanks @jiasli for the review, too!

@lochiiconnectivity , I added some changes mainly to the docs, as yet another [PR to your feature branch](https://github.com/claranet/microsoft-authentication-library-for-python/pull/1). Please help proofread and merge that PR when done. We will take care of this PR after that.

msal/application.py
```diff
@@ -151,6 +151,14 @@ def __init__(
             "The provided signature value did not match the expected signature value",
             you may try use only the leaf cert (in PEM/str format) instead.
 
+            *Added in version 1.13.0*:
+            It can also be a completly pre-signed assertion that you've assembled yourself.
+            Simply pass a container containing only the key "client_assertion", like this:
```
rayluo, line 156:
```suggestion
            Simply pass a container containing only the key "client_assertion", like this::
```
The [restructuredtext needs a double-colon](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#literal-blocks) here.

By the way, this suggestion can be accepted by one-click from the Github web interface.

tests/test_e2e.py
```diff
@@ -277,6 +277,16 @@ def test_subject_name_issuer_authentication(self):
         self.assertIn('access_token', result)
         self.assertCacheWorksForApp(result, scope)
 
+    def test_pre_signed_jwt_authentication(self):
```
rayluo, line 280:
```suggestion
    def test_client_assertion(self):
```

By the way, this suggestion can be accepted by one-click from the Github web interface.

rayluo (approved):
Feature-wise, this is completed. Approving now.

Somehow github reports a conflict which prevents my merge attempt. @lochiiconnectivity would you mind fixing this by a rebase-to-dev or merge-from-dev?

lochiiconnectivity:
I've resolved the merge conflict by hand, no need for rebase - it was minor.

rayluo:
Thanks David for driving this to the finish line! It will be included in our next release. At that time we will leave another note in this PR, so you will receive a notification. Meanwhile, relax and be proud. :-)

rayluo:
Hi Dave (@lochiiconnectivity ), I mean to reply your email today but somehow my response won't be sent. Just letting you know that we are still working on another feature which was also scheduled in the next release but took longer than expected. Don't worry. Our next release will automatically include all the features currently in dev branch.

Meanwhile, there is a workaround for you. You can use the following line to install the "cutting-edge" MSAL which includes your PR. Later when MSAL 1.13 is released, you simply switch that line back to the normal "pip install msal", without any code changes in your project.

    pip install git+https://github.com/AzureAD/microsoft-authentication-library-for-python.git@dev


Thanks again for your patience and understanding.

lochiiconnectivity:
Thanks, but we're waiting to publish software tied to 1.13 (as in msal=1.13 in requirements.txt), we can't do this without a published module , so we want you to publish 1.13 as soon as possible please.

rayluo:
Hi Dave (@lochiiconnectivity ), thanks for your patience. MSAL 1.13 has been released just now! Also thank you very much for your contribution!

Issue #232, opened 2020-07-21
Accept an encrypted key as client credential


chlowell:
[Documentation](https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/2f5774c7328261160d4201dfee0af7310fe86ab4/msal/application.py#L112) indicates keys should be provided to `ConfidentialClientApplication` as strings, PEM encoded. The documentation doesn't mention how to provide an encrypted key, and there's nothing obvious like a `password` parameter, so on the face of it `ConfidentialClientApplication` doesn't accept encrypted keys.

The implementation allows something of a workaround. It so happens that the given private key gets passed along to PyJWT, [which will accept a private key as deserialized by `cryptography`](https://github.com/jpadilla/pyjwt/issues/320#issuecomment-353236716). So this works today:
```py
from cryptography.hazmat.primitives import serialization
private_key = serialization.load_pem_private_key(pem_bytes, password=password, backend=default_backend())
ConfidentialClientApplication(client_credentials={"private_key": private_key, ...})
```
But that relies on today's implementation and isn't a substitute for formal support.

rayluo:
@abhidnya13 Let's add one more key (something like `passphrase` or `private_key_passphrase`?) into the existing `client_credentials` input parameter, and then internally we can use the implementation described above. Also, if the input `private_key_passphrase` is a string, we can support automatically convert it into bytes.

PR #272, opened 2020-11-02
MSAL Python 1.6.0

rayluo:
* New Feature: `ConfidentialClientApplication` accepts private key encrypted by a passphrase. (#232, #270)
* Enhancement: Provides different exception and messages while encountering transient error during tenant discovery (#263, #269)


abhidnya13 (approved):
🚢

Issue #297, opened 2021-01-26
Client.obtain_token_by_refresh_token fails when token_item is a str


jiasli:
**Describe the bug**
Even though `Client.obtain_token_by_refresh_token` checks the type of `token_item` at 

https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/381fe584df39f69a952d6e685758f93f581c2b96/msal/oauth2cli/oauth2.py#L801-L802

and calls `rt_getter` accordingly, it doesn't check the type of `token_item` at

https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/381fe584df39f69a952d6e685758f93f581c2b96/msal/oauth2cli/oauth2.py#L809-L810

to make sure `token_item` is converted to a `dict` before calling `self.on_updating_rt`.

This causes `update_rt` to fail at L248:

https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/0845404ef7ec6a95e0fecf123a815e06dc683d85/msal/token_cache.py#L247-L248

**To Reproduce**
Pass a refresh token in `str` as `token_item` to `Client.obtain_token_by_refresh_token`.

**Expected behavior**
Either:
- `Client.obtain_token_by_refresh_token` should succeed with `token_item` as a `str` either by
  - converting `token_item` to a `dict` before calling `on_updating_rt`
  - forcing `on_updating_rt=False`
- `Client.obtain_token_by_refresh_token` should reject `token_item` as a `str`

**What you see instead**

```
  File "D:\cli\microsoft-authentication-library-for-python\msal\application.py", line 714, in acquire_token_silent
    result = self.acquire_token_silent_with_error(
  File "D:\cli\microsoft-authentication-library-for-python\msal\application.py", line 767, in acquire_token_silent_with_error
    result = self._acquire_token_silent_from_cache_and_possibly_refresh_it(
  File "D:\cli\microsoft-authentication-library-for-python\msal\application.py", line 840, in _acquire_token_silent_from_cache_and_possibly_refresh_it
    return self._acquire_token_silent_by_finding_rt_belongs_to_me_or_my_family(
  File "d:\cli\azure-cli\src\azure-cli-core\azure\cli\core\_msal.py", line 17, in _acquire_token_silent_by_finding_rt_belongs_to_me_or_my_family
    return self._acquire_token_silent_by_finding_specific_refresh_token(
[...]

PR #298, opened 2021-01-26
Pass kwargs to acquire_token_by_refresh_token

jiasli:
- Pass `kwargs` to `ClientApplication.acquire_token_by_refresh_token` so that `data` can be passed to `Client.obtain_token_by_refresh_token`.
- Call `_validate_ssh_cert_input_data` to make sure `data` is valid.


rayluo (commented):
The changes in this PR extends the functionality of `acquire_token_by_refresh_token()`. It is fine, but I'd like us to have a conclusion on #297 first, before we permanently commit to this new behaviors you added here.

PR #316, opened 2021-03-03
Change RTD link to a more family-friendly option

tonybaloney:


rayluo:
Ah, I did not know you already started working on this. How about you review #317? It is more completed there.

PR #318, opened 2021-03-03
Only trigger CI/CD when a PR is labelled

rayluo:
Experimental


rayluo (commented):
This works as expected. The CI/CD would be triggered only after a PR is labelled.

PR #319, opened 2021-03-04
Refactor docs into a different theme, improve nav

tonybaloney:
- show inherited members for the client classes
- group to an API section
- add navbar links to GitHub and the docs.microsoft.com page

this change would require changing the readthedocs configuration to install the `docs/requirements.txt` packages

Screen Shot 2021-03-04 at 11 17 31 am


docs/requirements.txt
```diff
@@ -0,0 +1 @@
+furo
```
rayluo, line 1: > this change would require changing the readthedocs configuration to install the docs/requirements.txt packages

Is such new dependency strictly necessary? Our docs on RTD was mean to be a no-frills rendering, because we tried to minimize our effort on maintenance.

Currently, that new [`furo` module happens to requires Python 3.5+](https://pypi.org/project/furo/), and RTD's default build interpretor happens to be Python 2.7 (who would know that...), so [it failed](https://readthedocs.org/projects/msal-python/builds/13156354/). Then I tried switching the RTD interpretor into Python 3.x, but then somehow it [failed with other error](https://readthedocs.org/projects/msal-python/builds/13156392/).

FWIW, Microsoft Docs team is also planning to generate and host MSAL docs on docs.microsoft.com. Perhaps there would be a unified scheme there.

docs/index.rst
```diff
@@ -71,11 +65,3 @@ See `SerializableTokenCache` for example.
 
 .. autoclass:: msal.SerializableTokenCache
    :members:
-
-
-Indices and tables
```
rayluo, line 76: :+1:

docs/index.rst
```diff
@@ -33,33 +28,32 @@ MSAL proposes a clean separation between
 They are implemented as two separated classes,
 with different methods for different authentication scenarios.
 
-PublicClientApplication
+ClientApplication
```
rayluo, line 31: The pre-existing section header (now changed to "API", which is fine), and the two paragraphs above, were trying to promote the 2 (not 3) classes as our conceptual API. Given that you now bring in the `:inherited-members` trick (Yay!), I think we can simply NOT mention parent class `ClientApplication` at all.

docs/conf.py
```diff
@@ -14,17 +14,20 @@
 #
 import os
 import sys
-sys.path.insert(0, os.path.abspath('..'))
+from datetime import date
+
+sys.path.insert(0, os.path.abspath(".."))
```
rayluo, line 19: Out of curiosity, why are there so many `'` to `"` changes? Are you using [Black](https://black.readthedocs.io/en/stable/the_black_code_style.html#strings)? Ideally, in a PR, we should aim for minimal noises (i.e. minimize the red-and-green sections).

docs/requirements.txt
tonybaloney: Fixed in next commit I hope by including the parent directory so it will install the requirements of the package

docs/index.rst
tonybaloney: Removed that class

docs/conf.py
tonybaloney: Reverted those changes, I still had auto-format on

rayluo (approved):
Nice work! [Staged docs look good](https://msal-python.readthedocs.io/en/docs-staging/).

docs/index.rst
```diff
@@ -35,31 +30,22 @@ with different methods for different authentication scenarios.
...
---------------
-Both PublicClientApplication and ConfidentialClientApplication
-have following methods inherited from their base class.
-You typically do not need to initiate this base class, though.
-
-.. autoclass:: msal.ClientApplication
+.. autoclass:: msal.ConfidentialClientApplication
    :members:
-
-   .. automethod:: __init__
```
rayluo, line 56: Hi @tonybaloney , remember this PR we did last month? Today I noticed that [the new documentation for `PublicClientApplication`](https://msal-python.readthedocs.io/en/stable/#publicclientapplication) and [`ConfidentialClientApplication`](https://msal-python.readthedocs.io/en/stable/#confidentialclientapplication) lose [the detailed documentation on `__init__()` available from older docs](https://msal-python.readthedocs.io/en/testing-docs/#shared-methods). Would you mind if we bring back the `Shared Methods` approach here (while keeping your new theme)?

PR #333, opened 2021-03-29
Support launching browser in WSL Ubuntu 18.04

### LINKED ISSUE #332: Support auth code flow in WSL Ubuntu 18.04
It will be good for MSAL to support auth code flow in WSL Ubuntu 18.04.

`webbrowser` doesn't work in WSL Ubuntu 18.04 due to the lack of `www-browser`:

```sh
$ python3 -V
Python 3.6.9
$ python3 -c "import webbrowser; print(webbrowser.open('https://microsoft.com'))"
False
$ python3 -c "import webbrowser; print(webbrowser.get())"
  File "<string>", line 1, in <module>
  File "/usr/lib/python3.6/webbrowser.py", line 51, in get
    raise Error("could not locate runnable browser")
webbrowser.Error: could not locate runnable browser
$ which www-browser
$
```

WSL Ubuntu 20.04 already supports this:

```sh
$ python3 -V
Python 3.8.5
$ python3 -c "import webbrowser; print(webbrowser.get().name)"
www-browser
$ which www-browser
/usr/bin/www-browser
```

Azure CLI supports WSL Ubuntu 18.04 by

https://github.com/Azure/azure-cli/blob/3bff786b9e337aa35bb4cc64a09e5600a35663f8/src/azure-cli-core/azure/cli/core/util.py#L637-L648

```py
def open_page_in_browser(url):
    import subprocess
    import webbrowser
    platform_name, _ = _get_platform_info()


    if is_wsl():   # windows 10 linux subsystem
        try:
            # https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_powershell_exe
            # Ampersand (&) should be quoted
            return subprocess.call(['powershell.exe', '-Command', 'Start-Process "{}"'.format(url)])
        except OSError:  # WSL might be too old  # FileNotFoundError introduced in Python 3
            pass
```

Without this functionality, Azure CLI will have a regression when using MSAL.

jiasli:
Fix #332

The code is migrated from Azure CLI:

https://github.com/Azure/azure-cli/blob/3bff786b9e337aa35bb4cc64a09e5600a35663f8/src/azure-cli-core/azure/cli/core/util.py#L642-L648

```py
    if is_wsl():   # windows 10 linux subsystem
        try:
            # https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_powershell_exe
            # Ampersand (&) should be quoted
            return subprocess.call(['powershell.exe', '-Command', 'Start-Process "{}"'.format(url)])
        except OSError:  # WSL might be too old  # FileNotFoundError introduced in Python 3
            pass
```

`_browse` now tries `webbrowser.open` first. If that fails and WSL is detected, launch browser with `powershell.exe`.

As for why `powershell.exe` is preferred over `cmd.exe`, it is discussed in https://github.com/Azure/azure-cli/pull/12851.


jiasli:
`www-browser` internally also uses `powershell.exe`

https://github.com/wslutilities/wslu/blob/221e85457c01e92db675cc7b24a49dfcdfacc6a9/src/wslu-header#L156

```py
	"$(interop_prefix)$(sysdrive_prefix)"/Windows/System32/WindowsPowerShell/v1.0/powershell.exe -NoProfile -NonInteractive –ExecutionPolicy Bypass -Command "[Console]::OutputEncoding = [System.Text.Encoding]::UTF8; [Console]::InputEncoding = [System.Text.Encoding]::GetEncoding($cp); $@"
```

But we can keep it simple.

msal/oauth2cli/authcode.py
```diff
@@ -33,9 +33,34 @@ def obtain_auth_code(listen_port, auth_uri=None):  # Historically only used in t
...
+    browser_opened = webbrowser.open(auth_uri)  # Use default browser. Customizable by $BROWSER
+
+    # In WSL which doesn't have www-browser, try launching browser with PowerShell
+    if not browser_opened and is_wsl():
+        try:
+            import subprocess
+            # https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_powershell_exe
+            # Ampersand (&) should be quoted
+            exit_code = subprocess.call(
+                ['powershell.exe', '-NoProfile', '-Command', 'Start-Process "{}"'.format(auth_uri)])
```
jiasli, line 59: `-NoProfile` is necessary (https://github.com/Azure/azure-cli/pull/15361).

msal/oauth2cli/authcode.py
```diff
@@ -33,9 +33,34 @@ def obtain_auth_code(listen_port, auth_uri=None):  # Historically only used in t
...
 
+def is_wsl():
+    # "Official" way of detecting WSL: https://github.com/Microsoft/WSL/issues/423#issuecomment-221627364
+    # Run `uname -a` to get 'release' without python
+    #   - WSL 1: '4.4.0-19041-Microsoft'
+    #   - WSL 2: '4.19.128-microsoft-standard'
+    import platform
+    uname = platform.uname()
+    platform_name = getattr(uname, 'system', None).lower()
+    release = getattr(uname, 'release', None).lower()
```
rayluo, line 44: These 2 lines did not feel right. The default value `None` won't have the `lower()` method. Let's use this chance to make it compatible with some [other versions of Python](https://docs.python.org/2.7/library/platform.html#platform.uname) which do not return uname as a namedtuple.

```suggestion
    platform_name = getattr(uname, 'system', uname[0]).lower()
    release = getattr(uname, 'release', uname[2]).lower()
```

tests/test_authcode.py
```diff
@@ -0,0 +1,26 @@
+from unittest.mock import patch
```
rayluo, line 1: Azure CLI as an application can move on to use newer version of Python, but MSAL as a library still tries to keep compatibility with older Python when possible.

Here, `mock` is not a built-in module in Python 2.7, so the [build actions are failing](https://github.com/AzureAD/microsoft-authentication-library-for-python/actions). I'd suggest either declare a test dependency of external `mock` when running on Python 2.7, or simply remove this test case. Seriously, for logic as simple as this pr, I do not think we have to rely on test case like this.

rayluo (changes requested):

PR #349, opened 2021-04-13
get_accounts() will group equivalent accounts and return the merged result

rayluo:
This way, the caller of `get_accounts(...)` (which is typically downstream application's account picker) would no longer need to choose among multiple equivalent accounts. This change is functionally equivalent in terms of the subsequent `acquire_token_silent(..., account=account)` call.

CC: @jiasli @chlowell

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -642,10 +642,25 @@ class ClientApplication(object):
         return accounts
 
     def _find_msal_accounts(self, environment):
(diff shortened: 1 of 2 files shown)


jiasli (approved):

jiasli:
Tested with latest Azure CLI. Works like a charm!

PR #354, opened 2021-04-29
Use image map to guide customer to proper sample

rayluo:
[Experimenting a "MAP" concept](https://msal-python.readthedocs.io/en/docs-staging/) to hopefully guide people find "where am I (i.e. what am I going to build)" easier.


navyasric:
@rayluo This is a good idea. I noticed that the SPA and web app icons are repeating in the first diagram.

PR #358, opened 2021-05-06
Reginal endpoint support

### LINKED ISSUE #295: Feature: Enable regional support
[Feature 954484](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/954484): Enable azure ad region separation in MSAL Python

rayluo:
This resolves #295.

Feature wise this is completed.

~~TODO: Currently the authority validation is still performed. Would need to decide whether that is desirable. CC: @bgavrilMS~~ Now MSAL will ignore connection error during Authority Validation if app developer opted in to use `azure_region`.

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -9,7 +9,6 @@ import logging
 import sys
 import warnings
 from threading import Lock
-import os
 
 import requests
 
@@ -20,6 +19,7 @@ from .wstrust_request import send_request as wst_send_request
 from .wstrust_response import *
 from .token_cache import TokenCache
 import msal.telemetry
+from .region import _detect_region
 
 
 # The __init__.py will import this. Not the other way around.
@@ -261,7 +261,7 @@ class ClientApplication(object):
             # Requests, does not support session - wide timeout
             # But you can patch that (https://github.com/psf/requests/issues/3341):
             self.http_client.request = functools.partial(
-                self.http_client.request, timeout=timeout)
+                self.http_client.request, timeout=timeout or 2)
 
             # Enable a minimal retry. Better than nothing.
             # https://github.com/psf/requests/blob/v2.25.1/requests/adapters.py#L94-L108
@@ -290,11 +290,8 @@ class ClientApplication(object):
             self._telemetry_buffer, self._telemetry_lock, api_id,
             correlation_id=correlation_id, refresh_reason=refresh_reason)
 
-    def _detect_region(self):
-        return os.environ.get("REGION_NAME")  # TODO: or Call IMDS
-
     def _get_regional_authority(self, central_authority):
-        self._region_detected = self._region_detected or self._detect_region()
+        self._region_detected = self._region_detected or _detect_region(self.http_client)
         if self._region_configured and self._region_detected != self._region_configured:
             logger.warning('Region configured ({}) != region detected ({})'.format(
                 repr(self._region_configured), repr(self._region_detected)))
@@ -369,27 +366,28 @@ class ClientApplication(object):
             on_updating_rt=self.token_cache.update_rt)
 
         regional_client = None
-        regional_authority = self._get_regional_authority(authority)
-        if regional_authority:
-            regional_configuration = {
-                "authorization_endpoint": regional_authority.authorization_endpoint,
-                "token_endpoint": regional_authority.token_endpoint,
-                "device_authorization_endpoint":
-                    regional_authority.device_authorization_endpoint or
-                    urljoin(regional_authority.token_endpoint, "devicecode"),
-                }
-            regional_client = Client(
-                regional_configuration,
-                self.client_id,
-                http_client=self.http_client,
-                default_headers=default_headers,
-                default_body=default_body,
-                client_assertion=client_assertion,
-                client_assertion_type=client_assertion_type,
-                on_obtaining_tokens=lambda event: self.token_cache.add(dict(
-                    event, environment=authority.instance)),
-                on_removing_rt=self.token_cache.remove_rt,
-                on_updating_rt=self.token_cache.update_rt)
+        if client_credential:  # Currently regional endpoint only serves some CCA flows
+            regional_authority = self._get_regional_authority(authority)
+            if regional_authority:
+                regional_configuration = {
+                    "authorization_endpoint": regional_authority.authorization_endpoint,
+                    "token_endpoint": regional_authority.token_endpoint,
+                    "device_authorization_endpoint":
+                        regional_authority.device_authorization_endpoint or
+                        urljoin(regional_authority.token_endpoint, "devicecode"),
+                    }
+                regional_client = Client(
+                    regional_configuration,
+                    self.client_id,
+                    http_client=self.http_client,
+                    default_headers=default_headers,
+                    default_body=default_body,
+                    client_assertion=client_assertion,
+                    client_assertion_type=client_assertion_type,
+                    on_obtaining_tokens=lambda event: self.token_cache.add(dict(
+                        event, environment=authority.instance)),
+                    on_removing_rt=self.token_cache.remove_rt,
+                    on_updating_rt=self.token_cache.update_rt)
         return central_client, regional_client
 
     def initiate_auth_code_flow(
--- a/msal/application.py
+++ b/msal/application.py
@@ -328,7 +328,12 @@ class ClientApplication(object):
         if region_to_use:
             logger.info('Region to be used: {}'.format(repr(region_to_use)))
             regional_host = ("{}.login.microsoft.com".format(region_to_use)
-                if central_authority.instance == "login.microsoftonline.com"
+                if central_authority.instance in (
+                    # The list came from https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/358/files#r629400328
+                    "login.microsoftonline.com",
+                    "login.windows.net",
+                    "sts.windows.net",
+                    )
                 else "{}.{}".format(region_to_use, central_authority.instance))
             return Authority(
                 "https://{}/{}".format(regional_host, central_authority.tenant),
--- a/msal/application.py
+++ b/msal/application.py
@@ -282,10 +282,24 @@ class ClientApplication(object):
 
         self.app_name = app_name
         self.app_version = app_version
-        self.authority = Authority(
+
+        # Here the self.authority will not be the same type as authority in input
+        try:
+            self.authority = Authority(
                 authority or "https://login.microsoftonline.com/common/",
                 self.http_client, validate_authority=validate_authority)
-            # Here the self.authority is not the same type as authority in input
+        except ValueError:  # Those are explicit authority validation errors
+            raise
+        except Exception:  # The rest are typically connection errors
+            if validate_authority and region:
+                # Since caller opts in to use region, here we tolerate connection
+                # errors happened during authority validation at non-region endpoint
+                self.authority = Authority(
+                    authority or "https://login.microsoftonline.com/common/",
+                    self.http_client, validate_authority=False)
+            else:
+                raise
+
         self.token_cache = token_cache or TokenCache()
         self._region_configured = region
         self._region_detected = None
--- a/msal/application.py
+++ b/msal/application.py
@@ -119,7 +119,7 @@ class ClientApplication(object):
             verify=True, proxies=None, timeout=None,
             client_claims=None, app_name=None, app_version=None,
             client_capabilities=None,
-            region=None,  # Note: We choose to add this param in this base class,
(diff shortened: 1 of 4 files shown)


msal/region.py
```diff
@@ -0,0 +1,33 @@
+import os
+import json
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+def _detect_region(http_client=None):
+    region = _detect_region_of_azure_function()  # It is cheap, so we do it always
+    if http_client and not region:
```
henrik-me, line 10: >http_client [](http://example.com/codeflow?start=7&length=11)

I recommend a dedicated http client for this call with a timeout set to 20 seconds or so... Neha and Bogdan can advise on the value to use.

msal/application.py
```diff
@@ -220,6 +228,36 @@ def __init__(
...
+
+            4. An app which already onboard to the region's allow-list.
+
+            MSAL's default value is None, which means region behavior remains off.
+            If enabled, some of the MSAL traffic would remain inside that region.
+
+            App developer can opt in to regional endpoint,
+            by provide a region name, such as "westus", "eastus2".
+
+            An app running inside Azure VM can use a special keyword
```
henrik-me, line 255: >Azure VM  [](http://example.com/codeflow?start=34&length=9)

not just azure vm. Azure PaaS are supported afaik, Neha and Bogdan can keep me honest on the details

msal/application.py
```diff
@@ -220,6 +228,36 @@ def __init__(
...
+            ``ClientApplication.ATTEMPT_REGION_DISCOVERY`` to auto-detect region.
+            (Attempting this on a non-VM could hang indefinitely.
+            Make sure you configure a short timeout,
+            or provide a custom http_client which has a short timeout.
```
henrik-me, line 259: >http_client [](http://example.com/codeflow?start=32&length=11)

imo, we should have dedicated http client with a shorter timeout for this.

msal/region.py
rayluo: Actually, the [internal proposal](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=%2FPinAuthToRegion%2FAAD%20SDK%20Proposal%20to%20Pin%20Auth%20to%20region.md&_a=preview&anchor=goals) suggests a 2-second timeout only. But I already had an off-line conversation with Bogdan days ago about the situation that, in Python there is no a system-wide `System.Net.Http.HttpClient`, so, there is no universal way for MSAL Python to configure timeout, especially when the incoming `http_client` is provided by app developer.

Currently, we try to provide clear documentation, also included in this PR.

The conversation did confirm one thing, that the region detection would be turned off by default.

msal/application.py
rayluo: Having 2 different http clients (one dedicated for region detection and one for the rest), would complicate not just MSAL's internal implementation, but also the API model, which would then propagate to our app developer's implementation. Yet the gain is debatable.

When an http_client has short timeout (say, 2 seconds x 2 retries = 4 seconds latency), those seemingly short latency might go unnoticed, and become a perpetual behavior for that app. (If that app happens to be a command-line app such as Azure CLI, it would mean each "az ..." would have 4+ seconds extra latency.)

The current approach would have a longer latency by default. But hopefully this "[fail early, fail loudly](http://oncodingstyle.blogspot.com/2008/10/fail-early-fail-loudly.html)" approach would lead customer to a better solution: to make their "region" setting configurable, so that it could (and should) be completely turned off by an on-site configuration, when their app is deployed outside of Azure VMs.

msal/application.py
```diff
@@ -108,14 +109,21 @@ class ClientApplication(object):
...
     def __init__(
             self, client_id,
             client_credential=None, authority=None, validate_authority=True,
             token_cache=None,
             http_client=None,
             verify=True, proxies=None, timeout=None,
             client_claims=None, app_name=None, app_version=None,
-            client_capabilities=None):
+            client_capabilities=None,
+            region=None,  # Note: We choose to add this param in this base class,
```
bgavrilMS, line 122: Just a note on naming - we called it "AzureRegion" since it only makes sense if the app (service) which uses MSAL runs on from the Azure cloud.

msal/application.py
```diff
@@ -220,6 +228,36 @@ def __init__(
...
+
+            3. An app authenticated by Subject Name/Issuer (SNI).
+
+            4. An app which already onboard to the region's allow-list.
+
+            MSAL's default value is None, which means region behavior remains off.
+            If enabled, some of the MSAL traffic would remain inside that region.
+
+            App developer can opt in to regional endpoint,
+            by provide a region name, such as "westus", "eastus2".
```
bgavrilMS, line 253: Please add a reference to https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.management.resourcemanager.fluent.core.region?view=azure-dotnet - this is the format in which the region should be passed. CC @neha-bhargava

msal/application.py
bgavrilMS: Yes, it's not just Azure VM. I remember @neha-bhargava tested this on other systems. Azure Functions I think supports this, but may be wrong, @neha-bhargava did a lot of testing here and is a great oracle of information :)

msal/region.py
```diff
@@ -0,0 +1,33 @@
...
+        return _detect_region_of_azure_vm(http_client)  # It could hang for minutes
+    return region
+
+
+def _detect_region_of_azure_function():
+    return os.environ.get("REGION_NAME")
+
+
+def _detect_region_of_azure_vm(http_client):
```
bgavrilMS, line 19: I think this is a little more complicated ... see https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/1770654e7a2985dc06849366e5ae8e229b22faa7/src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs#L192

As far as I understood, the algorithm is: 

- try with IDMS api version `2020-06-01`
- if you get a BadRequest (400) error then make a request to `http://169.254.169.254/metadata/instance/compute/location` which will still return a 400 error but the response will have the API version https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/1770654e7a2985dc06849366e5ae8e229b22faa7/src/client/Microsoft.Identity.Client/Instance/Region/LocalImdsErrorResponse.cs#L21
- Use this new api version parameter and retry the call

@neha-bhargava - could you update the API design document with this algorithm please, since you understand it better.

msal/application.py
```diff
@@ -260,6 +301,27 @@ def _build_telemetry_context(
...
+        self._region_detected = self._region_detected or _detect_region(
+            self.http_client if self._region_configured is not None else None)
+        if (is_region_specified and self._region_configured != self._region_detected):
+            logger.warning('Region configured ({}) != region detected ({})'.format(
+                repr(self._region_configured), repr(self._region_detected)))
+        region_to_use = (
+            self._region_configured if is_region_specified else self._region_detected)
+        if region_to_use:
+            logger.info('Region to be used: {}'.format(repr(region_to_use)))
+            regional_host = ("{}.login.microsoft.com".format(region_to_use)
```
bgavrilMS, line 316: I think this does not take into account sovereign and air-gapped clouds. If the original env is the public one (i.e. "login.microsoftonline.com", "login.windows.net", "login.microsoft.com" or "sts.windows.net" ... these are the only ones we know about), then use `<region>.login.microsoft.com`

Otherwise, use `<region>.<original_env>`, e.g. if the original env is "login.microsoftonline.de", then you'd use `<region>.login.microsoftonline.de`. 

This is very important, because app developers write code once and expect it to work on public + sovereign.

msal/region.py
henrik-me: Right, I meant 2 seconds... There must be a way even for Python to ensure this. 

Region detection should be used in the context where region detection is needed, always, not off by default. Azure region is not the default behavior of AquireTokenForClient (though it may be some day) if that is what you mean.

msal/application.py
henrik-me: This is not a http client dedicated for region, but a client dedicated to fetch data from the imds endpoint. Primary reason is the timeout setting from above.

msal/application.py
neha-bhargava: Right, auto detection is definitely available on azure functions (Environment variable) and VMs (IMDS call). Since the EAP servers also use the VMs we were able detect the region through IMDS call there. We know cloud services is not supported by imds and env. variable was also not set there. Other than that we don't know yet.

msal/application.py
neha-bhargava: Yes, usually imds call fails fast but with the default retry logic it was taking longer to fail making the overall response time longer. So it was decided to set timeout to 2 sec for the imds call specifically with no retries.

msal/application.py
```diff
@@ -220,6 +228,36 @@ def __init__(
...
+            2. An app with managed identity, which is formerly known as MSI.
+               (However MSAL Python does not support managed identity,
+               so this one does not apply.)
+
+            3. An app authenticated by Subject Name/Issuer (SNI).
+
+            4. An app which already onboard to the region's allow-list.
+
+            MSAL's default value is None, which means region behavior remains off.
+            If enabled, some of the MSAL traffic would remain inside that region.
```
neha-bhargava, line 250: Does this imply that only traffic for `acquire_token_for_client()` will go to that region?

msal/application.py
```diff
@@ -260,6 +301,27 @@ def _build_telemetry_context(
...
+    def _get_regional_authority(self, central_authority):
+        is_region_specified = bool(self._region_configured
+            and self._region_configured != self.ATTEMPT_REGION_DISCOVERY)
+        self._region_detected = self._region_detected or _detect_region(
+            self.http_client if self._region_configured is not None else None)
+        if (is_region_specified and self._region_configured != self._region_detected):
+            logger.warning('Region configured ({}) != region detected ({})'.format(
+                repr(self._region_configured), repr(self._region_detected)))
+        region_to_use = (
+            self._region_configured if is_region_specified else self._region_detected)
```
neha-bhargava, line 313: In case _region_specified is different from _region_detected what will be the value of region_to_use here?

msal/region.py
neha-bhargava: The api version used here is recent I believe which is fine. But these api versions keep on updating and after sometime the version we mentioned here might no longer be available. So to handle that we check for the BadRequest error and make a request without the api version param which returns the latest api version which we can use to make the api call.

neha-bhargava: Updated the api doc: https://identitydivision.visualstudio.com/DefaultCollection/DevEx/_git/AuthLibrariesApiReview/pullrequest/3291

msal/region.py
```diff
@@ -0,0 +1,33 @@
...
+    url = "http://169.254.169.254/metadata/instance?api-version=2021-01-01"
```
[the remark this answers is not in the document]
rayluo: Thanks for the suggestion. Making this change in the next commit!

msal/application.py
rayluo: It will use `_region_configured`.
PS: That "... if ... else ..." is the Python way of ternary operator. It is supposed to mimic the normal English "outcome = advanced if won else eliminated".

msal/application.py
rayluo: OK. Changed in latest commit.

msal/application.py
neha-bhargava: @rayluo agree that there is a need to set the timeout specifically for the imds call to 2 seconds.

msal/region.py
neha-bhargava: The approach we selected here was to reduce the possibility of failing when they discontinue the version we have. Hardcoding it to make less network calls and because it is true that this would happen far less and far long. Is there any other approach that you would like to suggest for future to avoid failures when the version we mention is no longer supported?

rayluo: Hardcoding an existing known-to-work version is not even about making less network calls. It is because it is the (only?) right way to do versioning among loosely-coupling web services.

And, nope, there will be no programmatical way to avoid failures like this. There has to be out-of-band (OOB) information to trigger such change. Those OOB info would ideally be conveyed as an ahead notice from IMDS team to us, or - if I may be honest here - via an IcM.

msal/application.py
rayluo: Note: The specs above has been changed. The latest specs is in this [Microsoft internal doc](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=/PinAuthToRegion/AAD%20SDK%20Proposal%20to%20Pin%20Auth%20to%20region.md&anchor=algorithm&_a=preview)

PR #363, opened 2021-05-19
Fix NameError in ClientApplication exception handler

chlowell:
To reproduce the error:
```py

class Client:
    def get(*args, **kwargs):
        raise Exception('oops')

PublicClientApplication('...', http_client=Client())
```
```
  File "repro.py", line 8, in <module>
    PublicClientApplication("...", http_client=Client())
  File "...lib\site-packages\msal\application.py", line 1296, in __init__
    super(PublicClientApplication, self).__init__(
  File "...lib\site-packages\msal\application.py", line 302, in __init__
    if validate_authority and region:
NameError: name 'region' is not defined
```

I assume the keyword argument `azure_region` is intended.

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -299,7 +299,7 @@ class ClientApplication(object):
         except ValueError:  # Those are explicit authority validation errors
             raise
         except Exception:  # The rest are typically connection errors
-            if validate_authority and region:
+            if validate_authority and azure_region:
                 # Since caller opts in to use region, here we tolerate connection
                 # errors happened during authority validation at non-region endpoint
                 self.authority = Authority(


msal/application.py
```diff
@@ -299,7 +299,7 @@ def __init__(
         except ValueError:  # Those are explicit authority validation errors
             raise
         except Exception:  # The rest are typically connection errors
-            if validate_authority and region:
+            if validate_authority and azure_region:
```
rayluo, line 302: Argh, our last-minute change from `region` to `azure_region` left this one behind. My bad.

Nice catch! Thanks!

rayluo (approved):

Issue #370, opened 2021-06-21
Session/socket never closed, warnings shown


HansWeltar:
When you run python with -Wdefault (show warnings) and do some work with the msal lib,
the following warning will be displayed on stdout:
sys:1: ResourceWarning: unclosed <ssl.SSLSocket fd=3, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('10.200.33.175', 36376), raddr=('20.190.160.69', 443)>

It seems http_client and it's underlying socket are never closed by the msal lib.
This causes python to print a warning on stdout (when running as a unittest or with -Wdefault)
By default http_client is a requests.Session and these keep the socket open until the Session is explicitely closed.
However we found no place in the code where 'close' is called on http_client.

```python
import msal
import warnings
# APPID="...."
# SECRET="...."
# TENANT="...."
warnings.filterwarnings('always')  # enable warnings, same as python3 -Wdefault
app=msal.application.ConfidentialClientApplication(APPID, client_credential=SECRET, authority="https://login.microsoftonline.com/" + TENANT)
app.acquire_token_for_client(scopes=["https://graph.microsoft.com/.default"])
del(app)  # -> triggers the unclosed socket warning
```

Or, to use one of the samples:
`python3 -Wdefault confidential_client_secret_sample.py config.json`
For us this shows:
```
$ python3 -Wdefault confidential_client_secret_sample.py config.json 
confidential_client_secret_sample.py:42: ResourceWarning: unclosed file <_io.TextIOWrapper name='config.json' mode='r' encoding='UTF-8'>
  config = json.load(open(sys.argv[1]))
ResourceWarning: Enable tracemalloc to get the object allocation traceback
Graph API call result: {
  "@odata.context": "https://graph.microsoft.com/v1.0/$metadata#users",
  "value": [
    {
      ....
    }
  ]
}
sys:1: ResourceWarning: unclosed <ssl.SSLSocket fd=3, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('10.200.33.175', 33236), raddr=('40.126.31.6', 443)>
ResourceWarning: Enable tracemalloc to get the object allocation traceback
```
Note: The ResourceWarnings might not always show at the end of the program (depends on the garbage collector). To reliably reproduce this issue, use the snippet above with the explicit `del(app)`
[...]

PR #371, opened 2021-06-21
Add close method to app

HansWeltar:
Fixes Issue #370 Session/socket never closed, warnings shown

The msal application does not close it's resources like eg: self.http_client.
As a result a 'ResourceWarning unclosed <ssl.SSLSocket' is shown when the
app goes out of scope (when warnings are enabled)

This fix adds and explicit close method so developpers can cleanly close
the app if they so desire.

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -281,7 +281,9 @@ class ClientApplication(object):
 
         if http_client:
             self.http_client = http_client
+            self.internal_client = False
         else:
+            self.internal_client = True
             self.http_client = requests.Session()
             self.http_client.verify = verify
             self.http_client.proxies = proxies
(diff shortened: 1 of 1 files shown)


HansWeltar:
A better fix might be needed. A close() method does not help in case of \_\_init\_\_ errors.
For example when you provide a wrong tenant_id
```python
try:
    app = msal.application.ConfidentialClientApplication(APPID, client_credential=SECRET, 
        authority="https://login.microsoftonline.com/" + FAULTY_TENANT)
except Exception:
    app.close() -> fails because app is not defined....
```

HansWeltar:
adressed
> A better fix might be needed. A close() method does not help in case of __init__ errors.
> For example when you provide a wrong tenant_id
> 
> ```python
> try:
>     app = msal.application.ConfidentialClientApplication(APPID, client_credential=SECRET, 
>         authority="https://login.microsoftonline.com/" + FAULTY_TENANT)
> except Exception:
>     app.close() -> fails because app is not defined....
> ```

bgavrilMS:
Seems like the original issue still exists? Is this PR valid?

PR #379, opened 2021-07-08
Http cache

### LINKED ISSUE #80: Instance metadata caching
This issue is inspired by an improvement made in MSAL .Net 4.1:
* documented in [its release blog post here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/msal-net-4.1#getaccounts-and-acquiretokensilent-are-now-less-network-chatty),
* its original issue [this one](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1174) and [that one](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1040),
* the [internal design review](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/859?_a=files),
* and also [its implementation part1](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1221) and [part 2](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1229).

### LINKED ISSUE #159: [Feature Request] Client throttling is supported in Public Client Application
**Overview**
- Blocking request after receiving 429, 5xx HTTP status.
- Respecting Retry-After header on any HTTP response.
- Cache for responses of interaction required messages.

**Important**
It is expectation that the calling application behavior should not change by this features, as well as it is not expected the increase of number of prompts. 

**Note:** 
This effort is related to Public Client applications only. 

**Related docs:** 
[Epic 328219: Client throttling is supported in MSAL & Brokers](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/823219)
[Design](https://identitydivision.visualstudio.com/devex/_git/AuthLibrariesApiReview?version=GBdev&path=%2FService%20protection%2FIntial%20set%20of%20protection%20measures.md&_a=preview)

rayluo:
It is the cache behavior on http layer. Such a cache is used internally to cache some http responses for performance optimization. In particular, when such an optimization happens on http 429/5xx with Retry-After header, it is a protection also known as "throttling" in MSAL parlance. This behavior resolves #159.

Such an http cache behavior is always enabled. Currently, the http cache is in-memory only.

### DIFF
--- a/msal/individual_cache.py
+++ b/msal/individual_cache.py
@@ -1,0 +1,286 @@
+from functools import wraps
+import time
+try:
+    from collections.abc import MutableMapping  # Python 3.3+
+except ImportError:
+    from collections import MutableMapping  # Python 2.7+
+import heapq
+from threading import Lock
+
+
+class _ExpiringMapping(MutableMapping):
+    _INDEX = "_index_"
+
+    def __init__(self, mapping=None, capacity=None, expires_in=None, lock=None,
+        *args, **kwargs):
+        """Items in this mapping can have individual shelf life,
+        just like food items in your refrigerator have their different shelf life
+        determined by each food, not by the refrigerator.
+
+        Expired items will be automatically evicted.
+        The clean-up will be done at each time when adding a new item,
+        or when looping or counting the entire mapping.
+        (This is better than being done indecisively by a background thread,
+        which might not always happen before your accessing the mapping.)
+
+        This implementation uses no dependency other than Python standard library.
+
+        :param MutableMapping mapping:
+            A dict-like key-value mapping, which needs to support __setitem__(),
+            __getitem__(), __delitem__(), get(), pop().
+
+            The default mapping is an in-memory dict.
+
+            You could potentially supply a file-based dict-like object, too.
+            This implementation deliberately avoid mapping.__iter__(),
+            which could be slow on a file-based mapping.
+
+        :param int capacity:
+            How many items this mapping will hold.
+            When you attempt to add new item into a full mapping,
+            it will automatically delete the item that is expiring soonest.
+
+            The default value is None, which means there is no capacity limit.
+
+        :param int expires_in:
+            How many seconds an item would expire and be purged from this mapping.
+            Also known as time-to-live (TTL).
+            You can also use :func:`~set()` to provide per-item expires_in value.
+
+        :param Lock lock:
+            A locking mechanism with context manager interface.
+            If no lock is provided, a threading.Lock will be used.
+            But you may want to supply a different lock,
+            if your customized mapping is being shared differently.
+        """
+        super(_ExpiringMapping, self).__init__(*args, **kwargs)
+        self._mapping = mapping if mapping is not None else {}
+        self._capacity = capacity
+        self._expires_in = expires_in
+        self._lock = Lock() if lock is None else lock
+
+    def _validate_key(self, key):
+        if key == self._INDEX:
+            raise ValueError("key {} is a reserved keyword in {}".format(
+                key, self.__class__.__name__))
+
+    def set(self, key, value, expires_in):
+        # This method's name was chosen so that it matches its cousin __setitem__(),
+        # and it also complements the counterpart get().
+        # The downside is such a name shadows the built-in type set in this file,
+        # but you can overcome that by defining a global alias for set.
+        """It sets the key-value pair into this mapping, with its per-item expires_in.
+
+        It will take O(logN) time, because it will run some maintenance.
+        This worse-than-constant time is acceptable, because in a cache scenario,
+        __setitem__() would only be called during a cache miss,
+        which would already incur an expensive target function call anyway.
+
+        By the way, most other methods of this mapping still have O(1) constant time.
+        """
+        with self._lock:
+            self._set(key, value, expires_in)
+
+    def _set(self, key, value, expires_in):
+        # This internal implementation powers both set() and __setitem__(),
+        # so that they don't depend on each other.
+        self._validate_key(key)
+        sequence, timestamps = self._mapping.get(self._INDEX, ([], {}))
+        self._maintenance(sequence, timestamps)  # O(logN)
+        now = int(time.time())
+        expires_at = now + expires_in
+        entry = [expires_at, now, key]
+        is_new_item = key not in timestamps
+        is_beyond_capacity = self._capacity and len(timestamps) >= self._capacity
+        if is_new_item and is_beyond_capacity:
+            self._drop_indexed_entry(timestamps, heapq.heappushpop(sequence, entry))
+        else:  # Simply add new entry. The old one would become a harmless orphan.
+            heapq.heappush(sequence, entry)
+        timestamps[key] = [expires_at, now]  # It overwrites existing key, if any
+        self._mapping[key] = value
+        self._mapping[self._INDEX] = sequence, timestamps
+
+    def _maintenance(self, sequence, timestamps):  # O(logN)
+        """It will modify input sequence and timestamps in-place"""
+        now = int(time.time())
+        while sequence:  # Clean up expired items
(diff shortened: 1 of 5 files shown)


msal/http_decorate.py
```diff
@@ -0,0 +1,114 @@
...
+# https://datatracker.ietf.org/doc/html/rfc8628#section-3.4
+DEVICE_AUTH_GRANT = "urn:ietf:params:oauth:grant-type:device_code"
+
+
+def _hash(raw):
+    return sha256(repr(raw).encode("utf-8")).hexdigest()
+
+
+def _handle_http_429_5xx_retry_after(result=None, **ignored):
+    assert result
```
chlowell, line 17: I would be surprised by the `AssertionError` because the default value in the signature tells me `None` is an acceptable value

rayluo: Thanks for catching this kind of details!

Admittedly, it looks weird at first glance. The signature here seemingly contains `None` as the default value, only because the actual callback function's shape is decided by the `_IndividualCache`'s `__call__()`. In that shape, the signature is defined as a bunch of position-less parameters, and `result` is not even the first parameter. So, here in this layer, we use `def _handle_http_429_5xx_retry_after(result=None, **ignored):` to match that shape.

The `assert` here would never be triggered in real run, however, I deliberately add it to remind maintainers that the `result` should/would not be `None`.

Not sure how much of those info I can squeeze in a short sentence in my next commit. :-)

msal/http_decorate.py
```diff
@@ -0,0 +1,117 @@
...
+    response = result
+    lowercase_headers = {  # MSAL's HttpClient may not have headers
+        k.lower(): v for k, v in getattr(response, "headers", {}).items()}
+    default = 600
+    try:
+        # AAD's retry_after uses integer format only
+        # https://stackoverflow.microsoft.com/questions/264931/264932
+        retry_after = int(lowercase_headers.get("retry-after", default))
+    except ValueError:
+        retry_after = default
```
chlowell, line 32: If I'm reading this correctly, when we get a response whose Retry-After is not a number, we cache that response for 10 minutes? Probably that never happens but it would be an interesting bug if it ever did. Do you think it's reasonable to just forego the default and assume all responses are from AAD?

rayluo: The reasoning behind the current implementation was that the presence of `Retry-After` would mean the service wants the client to cool down for *some time*. When the current implementation does not (yet?) support all types of input, it seems a moderate default value would be close enough to honor the "spirit of the `Retry-After` header". It should be better of an uncaught exception, which could nullify the throttling effect.

We might, however, consider to also add support to the other time format of `Retry-After`. But, even after that, this default fallback branch would likely stay.

rayluo: UPDATE: In the latest commit, I added support to another popular date format, even though it is not (currently?) used by AAD; I also changed the default value from 600 to 60, which is hopefully less aggressive now.

Closing this conversation.

msal/individual_cache.py
```diff
@@ -0,0 +1,283 @@
...
+        with self._lock:
+            # Skip self._maintenance(), because it would need O(logN) time
+            sequence, timestamps = self._mapping.get(self._INDEX, ([], {}))
+            expires_at, created_at = timestamps[key]  # Would raise KeyError accordingly
+            now = int(time.time())
+            if not created_at <= now < expires_at:
+                self._mapping.pop(key, None)
+                timestamps.pop(key, None)
+                self._mapping[self._INDEX] = sequence, timestamps
+                raise KeyError("{} expired".format(key))
```
chlowell, line 146: Or was created in the future 😉

msal/individual_cache.py
```diff
@@ -0,0 +1,283 @@
...
+        By the way, most other methods of this mapping still have O(1) constant time.
+        """
+        with self._lock:
+            self._set(key, value, expires_in)
+
+    def _set(self, key, value, expires_in):
+        # This internal implementation powers both set() and __setitem__(),
+        # so that they don't depend on each other.
+        self._validate_key(key)
+        sequence, timestamps = self._mapping.get(self._INDEX, ([], {}))
```
chlowell, line 88: Looking this over left me thinking the class is doing redundant bookkeeping. `sequence` is a list of `[expires_at, created_at, key]` entries, while `timestamps` is `{key: [expires_at, created_at]}`, i.e. the same data structured as a list and a dict. `sequence` records the history of a key's value changes (but not the values themselves) while `timestamps` records only the most recent time the value was set. `timestamps` seems more appropriate because there's only one value per key. Why maintain `sequence`, the history of value changes?

rayluo: Good question! The answer is, while the two bookkeeping data structures happen to contain similar data, their different data structure are chosen to suit two distinct purposes. The key-value pairs in `timestamps` would be used to perform O(1) check to see whether a given key's value already expired. The `sequence`, on the other hand, is used to perform regular maintenance to weep out expired values.

msal/individual_cache.py
```diff
@@ -0,0 +1,283 @@
...
+            timestamps.pop(key, None)  # This would probably always succeed
+
+    def __setitem__(self, key, value):
+        """Implements the __setitem__().
+
+        Same characteristic as :func:`~set()`,
+        but use class-wide expires_in which was specified by :func:`~__init__()`.
+        """
+        if self._expires_in is None:
+            raise ValueError("Need a numeric value for expires_in during __init__()")
```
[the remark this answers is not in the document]
rayluo: The `expires_in` has to be an optional parameter in `__init__()`, because this class is designed to be flexible enough to support per-item `set(key, value, expires_in)` (where `expires_in` is a required parameter there). But if the caller did not provide an `expires_in` during `__init__()`, yet ended up calling this `__setitem__()` code path, the safety check has to occur here and that error message is appropriate.

msal/individual_cache.py
rayluo: Good catch! I'll update that in next commit.

tests/test_throttled_http_client.py
```diff
@@ -0,0 +1,165 @@
...
+        self.assertNotEqual(resp1.text, resp2.text, "Should return a new response")
+        resp2_again = http_client.post("https://example.com", data={"claims": "bar"})
+        self.assertEqual(resp2.text, resp2_again.text, "Should return a cached response")
+
+    def test_one_foci_app_recovering_from_invalid_grant_should_also_unblock_another(self):
+        """
+        Need not test multiple FOCI app's acquire_token_silent() here. By design,
+        one FOCI app's successful populating token cache would result in another
+        FOCI app's acquire_token_silent() to hit a token without invoking http request.
+        """
```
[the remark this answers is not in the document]
rayluo: The missing implementation was a feature. (TM)  :stuck_out_tongue: 

The internal high level design suggestted a bunch of acceptance test cases. This one (and the next one) was one of them. So, in case a future maintainer would be auditing/comparing this implementation to the design, this test case and its docstring exist only to tell him/her that "we did not forget this acceptance test criteria, but this is the reason why this particular implementation is insusceptible to this test requirement".

tests/test_throttled_http_client.py
```diff
@@ -0,0 +1,165 @@
...
+        This was discussed in https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/3618?_a=files
+        """
+
+    def test_http_get_200_should_be_cached(self):
+        http_cache = {}
+        http_client = DummyHttpClient(
+            status_code=200)  # It covers UserRealm discovery and OIDC discovery
+        http_client = ThrottledHttpClient(http_client, http_cache)
+        resp1 = http_client.get("https://example.com")
+        resp2 = http_client.get("https://example.com")
```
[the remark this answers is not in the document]
rayluo: OK. It will be addressed in next commit.

PR #388, opened 2021-07-30
Prefer Edge when running on Linux

rayluo:
Edge will soon become the first browser on Linux to support device-based conditional access policy. This PR has MSAL to prefer using Edge for `acquire_token_interactive()`.

How to test this feature?

1. Install this feature branch by `pip install git+https://github.com/AzureAD/microsoft-authentication-library-for-python.git@prefer-edge-on-linux`.
2. Download the out-of-the-box [interactive sample](https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/1.13.0/sample/interactive_sample.py)
3. Prepare a config.json file with content like this, and then run `python interactive_sample.py config.json`:

```json
{
    "authority": "https://login.microsoftonline.com/common",
    "client_id": "289a413d-284b-4303-9c79-94380abe5d22",
    "scope": ["User.Read"],
    "endpoint": "https://graph.microsoft.com/v1.0/me",
    "placeholder": null
}
```

@chlowell @jiasli Please help review this feature.

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -9,6 +9,7 @@ import logging
 import sys
 import warnings
 from threading import Lock
+import os
 
 import requests
 
@@ -67,6 +68,46 @@ def _clean_up(result):
     if isinstance(result, dict):
         result.pop("refresh_in", None)  # MSAL handled refresh_in, customers need not
     return result
+
+
+def _preferred_browser():
+    """Register Edge and return a name suitable for subsequent webbrowser.get(...)
+    when appropriate. Otherwise return None.
+    """
+    # On Linux, only Edge will provide device-based Conditional Access support
+    if sys.platform != "linux":  # On other platforms, we have no browser preference
+        return None
+    browser_path = "/usr/bin/microsoft-edge"  # Use a full path owned by sys admin
+    user_has_no_preference = "BROWSER" not in os.environ
+    user_wont_mind_edge = "microsoft-edge" in os.environ.get("BROWSER", "")  # Note:
+        # BROWSER could contain "microsoft-edge" or "/path/to/microsoft-edge".
+        # Python documentation (https://docs.python.org/3/library/webbrowser.html)
+        # does not document the name being implicitly register,
+        # so there is no public API to know whether the ENV VAR browser would work.
+        # Therefore, we would not bother examine the env var browser's type.
+        # We would just register our own Edge instance.
+    if (user_has_no_preference or user_wont_mind_edge) and os.path.exists(browser_path):
+        try:
+            import webbrowser  # Lazy import. Some distro may not have this.
+            browser_name = "msal-edge"  # Avoid popular name "microsoft-edge"
+                # otherwise `BROWSER="microsoft-edge"; webbrowser.get("microsoft-edge")`
+                # would return a GenericBrowser instance which won't work.
+            try:
+                registration_available = isinstance(
+                    webbrowser.get(browser_name), webbrowser.BackgroundBrowser)
+            except webbrowser.Error:
+                registration_available = False
+            if not registration_available:
+                logger.debug("Register %s with %s", browser_name, browser_path)
+                # By registering our own browser instance with our own name,
+                # rather than populating a process-wide BROWSER enn var,
+                # this approach does not have side effect on non-MSAL code path.
+                webbrowser.register(  # Even double-register happens to work fine
+                    browser_name, None, webbrowser.BackgroundBrowser(browser_path))
+            return browser_name
+        except ImportError:
+            pass  # We may still proceed
+    return None
 
 
 class ClientApplication(object):
@@ -1393,6 +1434,7 @@ class PublicClientApplication(ClientApplication):  # browser app or mobile app
                 },
             data=dict(kwargs.pop("data", {}), claims=claims),
             headers=telemetry_context.generate_headers(),
+            browser_name=_preferred_browser(),
             **kwargs))
         telemetry_context.update_telemetry(response)
         return response


msal/application.py
```diff
@@ -69,6 +70,36 @@ def _clean_up(result):
...
+                # would return a GenericBrowser instance which won't work.
+            try:
+                registration_available = isinstance(
+                    webbrowser.get(browser_name), webbrowser.BackgroundBrowser)
+            except webbrowser.Error:
+                registration_available = False
+            if not registration_available:
+                logger.debug("Register %s with %s", browser_name, browser_path)
+                webbrowser.register(  # Even double-register happens to work fine
+                    browser_name, None, webbrowser.BackgroundBrowser(browser_path))
```
jiasli, line 96: Off topic: I actually find the implementation of `webbrowser.BackgroundBrowser` problematic.

https://github.com/python/cpython/blob/2b496e79293a8b80e8ba0e514e186b3b1467b64b/Lib/webbrowser.py#L190-L202

```py
class BackgroundBrowser(GenericBrowser):
    def open(self, url, new=0, autoraise=True):
        ...
            if sys.platform[:3] == 'win':
                p = subprocess.Popen(cmdline)
            ...
            return (p.poll() is None)
```

Per https://docs.python.org/3/library/subprocess.html#subprocess.Popen.poll:

> `Popen.poll()`
> Check if child process has terminated. Set and return returncode attribute. Otherwise, returns None.

Say my computer is super fast and `p.poll()` immediately returns `0`, this `open` method will return `False`. Then MSAL will think the browser is not opened, showing the warning 

https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/c687d5b0a198c11c2ff99fd0ca9feaf73a7376f0/msal/oauth2cli/authcode.py#L210-L221

We can mimic this behavior by putting a `time.sleep(2)` before `return (p.poll() is None)` to slow down `webbrowser.BackgroundBrowser.open`.

Therefore, it should actually be 

```py
poll_result = p.poll()
return poll_result is None or poll_result == 0
```

Perhaps we can submit a PR to https://github.com/python/cpython

jiasli: I tested on Windows. We need to explicitly request a web browser, like `firefox` (it should be on `PATH`).

rayluo: > I tested on Windows. We need to explicitly request a web browser, like `firefox` (it should be on `PATH`).

@jiasli Did you actually see any abnormal behavior when you were testing it on Windows? [Line 81 in the current PR](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/388/files#diff-4d32d284fa6acf43894719eff11cb366e5969e0b61593bfd00c68a4cf9c441caR81) would at least make this PR's behavior would only occur on Linux. I tested it on Windows just now, and its behavior is same as before, just opening my default browser.

> Off topic: I actually find the implementation of `webbrowser.BackgroundBrowser` problematic.
> 
> https://github.com/python/cpython/blob/2b496e79293a8b80e8ba0e514e186b3b1467b64b/Lib/webbrowser.py#L190-L202
> 
> ```python
> class BackgroundBrowser(GenericBrowser):
>     def open(self, url, new=0, autoraise=True):
>         ...
>             if sys.platform[:3] == 'win':
>                 p = subprocess.Popen(cmdline)
>             ...
>             return (p.poll() is None)
> ```
> 
> Per https://docs.python.org/3/library/subprocess.html#subprocess.Popen.poll:
> 
> > `Popen.poll()`
> > Check if child process has terminated. Set and return returncode attribute. Otherwise, returns None.
> 
> Say my computer is super fast and `p.poll()` immediately returns `0`, this `open` method will return `False`. Then MSAL will think the browser is not opened, showing the warning

In theory, yes, `subprocess.Popen(cmdline).poll()` could have finished very fast. (Actually I was not aware of this detail. Thanks for teaching me this new knowledge.)

But in our case, the browser is meant to open for user interaction. Even if the user would have a super fast computer, the opened browser window would not be immediately terminated unless the user closes it immediately. Even if the user does so, that subsequent MSAL warning message would seem quite reasonably in such a corner case.

Unless I haven't understood the problem thoroughly. Please let me know if I miss something.

jiasli: > Did you actually see any abnormal behavior when you were testing it on Windows?

In reality, no. This issue should only be a corner case.

> the opened browser window would not be immediately terminated unless the user closes it immediately.

This is not true per my observation. On Windows, even if the browser is not closed, `p.poll()` will return `0` very quickly. Putting a `time.sleep(2)` before `p.poll()` can easily repro this. This delay can also happen when the execution of `webbrowser` is suspended for some reason, for example, when garbage collection (GC) kicks in, CPU goes to 100%, etc.

rayluo: > > > I tested on Windows. We need to explicitly request a web browser, like `firefox` (it should be on `PATH`).
> >
> > @jiasli Did you actually see any abnormal behavior when you were testing it on Windows?
>
> In reality, no. This issue should only be a corner case.

I thought you were seeing some symptom that could be fix by your very specific suggestion ("need to explicitly request a web browser, like `firefox`"). If that is not the case, then it is not actionable here.

You did raise some good points on Windows. I think you can create yet another issue for the "acquire_token_interactive() on Windows" topic, and then we can keep track there. The current PR here contains new behaviors which are [only invoked when running on Linux](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/388/files#diff-4d32d284fa6acf43894719eff11cb366e5969e0b61593bfd00c68a4cf9c441caR81).

jiasli: I know, that's why I marked it as "off topic" at the very beginning. 😆

msal/application.py
```diff
@@ -69,6 +70,36 @@ def _clean_up(result):
...
+            browser_name = "msal-edge"  # A unique name rather than "microsoft-edge"
+                # otherwise `BROWSER="microsoft-edge"; webbrowser.get("microsoft-edge")`
+                # would return a GenericBrowser instance which won't work.
+            try:
+                registration_available = isinstance(
+                    webbrowser.get(browser_name), webbrowser.BackgroundBrowser)
+            except webbrowser.Error:
+                registration_available = False
+            if not registration_available:
+                logger.debug("Register %s with %s", browser_name, browser_path)
```
chlowell, line 94: You don't need `registration_available`, because the exception tells you whether the browser is registered. And do you need to check the registered type? Seems reasonable to assume applications won't register their own "msal-edge".

```suggestion
            try:
                webbrowser.get(browser_name)
                return browser_name
            except webbrowser.Error:
                logger.debug("Register %s with %s", browser_name, browser_path)
```

rayluo:
Before this PR, MSAL would always launch the default browser. After this PR, MSAL would launch Edge browser when running on Linux and when Edge is available. Here we document the different user experience of the two approaches when end user is on Linux and potentially accessing resources that are protected by device-based Conditional Access (CA) policy.

* If we launch default browser, which is typically not Edge.

1. For resources require no device-based CA, it would just work. And, typically user would already have a valid signed-in session (i.e. cookie) in their default browser, so, they even get SSO. This is the most ideal case.

2. For resources require device-based CA, user will be prompted during the sign-in experience. The suggestion of using Edge is currently in the fine print at the end of slide No.3 in this [internal document: Enrollment - high level experience .pptx](https://microsoft.sharepoint-df.com/:p:/t/LinuxEndpointManagement/ERdMk1FnROdApNG5L1IDdIwBioi5AtX93dleseI3M_VLcA?e=bCLoCj&wdOrigin=TEAMS-ELECTRON.p2p.bim&wdExp=TEAMS-TREATMENT&wdhostclicktime=1628956191851)

   > If you have already enrolled with Intune, *use Microsoft Edge* with your work or school profile to access this resource.​

3. Note that the error page above only mentions a vague "use Edge", it does NOT cover how to enable Edge for a specific app. And I doubt any MSAL-powered app would bother convey that instruction beforehand.

4. Even if the end user somehow figure out 2 & 3, he would repeatedly run into the same issue each time he is going to access that resource, which is annoying. To make the things worse, on Linux there is no universal way to specify a browser as default browser. The `BROWSER=microsoft-edge` way happens to work in Python, but end user may not know that. There would be some trial-and-error for the end user here.


* If we always launch Edge when it is available and when MSAL is running on Linux

1. For resources require no device-based CA, the very first time after Edge being installed into this desktop, the Edge will pop up, it contains no signed-in session, so that end user would need to sign in again. That is suboptimal, but at least it would be much easier than the 3 & 4 above.

2. After #1 above, #2 would just work. The end users don't have to read those fine print.

3. Not applicable.

4. Also not applicable. However, there is currently a known issue: currently the sign-in experience does not (yet?) present a "Stay signed in?" option to end user from Linux. There is an internal issue kept track [here](https://stackoverflow.microsoft.com/questions/270240).

msal/application.py
```diff
@@ -69,6 +70,36 @@ def _clean_up(result):
...
 
 
+def _preferred_browser():
+    """Register Edge and return a name suitable for webbrowser.get(...),
+    if running on Linux and there is no BROWSER env var, otherwise return None.
+    """
+    browser_path = "/usr/bin/microsoft-edge"  # Use a full path owned by sys admin
+    user_has_no_preference = "BROWSER" not in os.environ
+    user_opted_in_for_edge = "microsoft-edge" in os.environ.get("BROWSER", "")
+    if ((user_opted_in_for_edge or user_has_no_preference)
```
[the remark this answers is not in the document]
rayluo: To my surprise, based on my hack-and-learn-webbrowser experience, a `BROWSER=microsoft-edge` would not work, due to [this issue I reported to Python](https://bugs.python.org/issue44844) and [reported to Edge](https://microsoft.visualstudio.com/Edge/_workitems/edit/35081718?src=WorkItemMention&src-action=artifact_link). The workaround is to launch Edge by `BackgroundBrowser`, and the `user_opted_in_for_edge` logic here is to shield those users who opted in for Edge from this issue.

Perhaps I should also document this history into the source code?

msal/application.py
rayluo: Due to the significance of the need for `BackgroundBrowser`, as explained in the topic above, such an `isinstance(...)` is used here. Perhaps we could rename `registration_available` to something like `suitable_registration_available`, to emphasize our intention here.

msal/application.py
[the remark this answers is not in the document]
chlowell: That assumes users who registered `microsoft-edge` and set `BROWSER=microsoft-edge` didn't use `BackgroundBrowser`. What if they did, and made their own registration because they needed to specify different path?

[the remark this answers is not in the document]
rayluo: > That assumes users who registered `microsoft-edge` and set `BROWSER=microsoft-edge` didn't use `BackgroundBrowser`. What if they did, and made their own registration because they needed to specify different path?

By "users who registered `microsoft-edge`", did you mean they did `webbrowser.register("microsoft-edge", ...)`? Those users would not be affected by this PR, because MSAL registers its own with the name `msal-edge`.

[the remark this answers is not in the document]
chlowell: Those users are affected in that MSAL would override their configuration. If the user does `webbrowser.register("microsoft-edge", ...)` and `BROWSER=microsoft-edge`, MSAL registers msal-edge and launches that instead.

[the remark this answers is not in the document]
rayluo: Concluding this conversation. The next commit will contain more comments based on the recent conversation.

lulujrush:
I know this is long done and gone but my issue was the last one linked so I wanted to comment. 

IMHO I found this unfortunate that you took away the Linux Users choice of default browser. This has led me to remove Edge from my system and will not be re-introduced.

rayluo:
> IMHO I found this unfortunate that you took away the Linux Users choice of default browser. This has led me to remove Edge from my system and will not be re-introduced.

Fair enough. In case you want to try Edge again but want to have MSAL Python (thus Azure CLI) honor your default browser, you can set an env var `BROWSER=firefox_excutable_file`.

julian-alarcon:
I see this documented here:
* https://learn.microsoft.com/en-us/entra/identity/conditional-access/concept-conditional-access-conditions#supported-browsers
But I will prefer a more direct/honest error or warning than just ignoring the default browser. At least it should indicate why is it doing that

I already needed to remove MS Edge as default using:

```sh
sudo update-alternatives --remove x-www-browser /usr/bin/microsoft-edge-stable
sudo update-alternatives --remove gnome-www-browser /usr/bin/microsoft-edge-stable
```

PR #395, opened 2021-08-15
Implementing CCS Routing info

### LINKED ISSUE #376: MSAL libraries provide routing information to CCS
Internal workitem [here](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1333551?src=WorkItemMention&src-action=artifact_link).

rayluo:
The implementation is based on a follow-up conversation in this [internal design PR](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/3304). This PR will resolve #376.

### DIFF
--- a/tests/test_ccs.py
+++ b/tests/test_ccs.py
@@ -1,0 +1,73 @@
+import unittest
+try:
+    from unittest.mock import patch, ANY
+except:
+    from mock import patch, ANY
+
+from tests.http_client import MinimalResponse
+from tests.test_token_cache import build_response
+
+import msal
+
+
+class TestCcsRoutingInfoTestCase(unittest.TestCase):
+
+    def test_acquire_token_by_auth_code_flow(self):
+        app = msal.ClientApplication("client_id")
+        state = "foo"
+        flow = app.initiate_auth_code_flow(
+            ["some", "scope"], login_hint="johndoe@contoso.com", state=state)
+        with patch.object(app.http_client, "post", return_value=MinimalResponse(
+                status_code=400, text='{"error": "mock"}')) as mocked_method:
+            app.acquire_token_by_auth_code_flow(flow, {
+                "state": state,
+                "code": "bar",
+                "client_info":  # MSAL asks for client_info, so it would be available
+                    "eyJ1aWQiOiJhYTkwNTk0OS1hMmI4LTRlMGEtOGFlYS1iMzJlNTNjY2RiNDEiLCJ1dGlkIjoiNzJmOTg4YmYtODZmMS00MWFmLTkxYWItMmQ3Y2QwMTFkYjQ3In0",
+                })
+            self.assertEqual(
+                "Oid:aa905949-a2b8-4e0a-8aea-b32e53ccdb41@72f988bf-86f1-41af-91ab-2d7cd011db47",
+                mocked_method.call_args[1].get("headers", {}).get('X-AnchorMailbox'),
+                "CSS routing info should be derived from client_info")
+
+    # I've manually tested acquire_token_interactive. No need to automate it,
+    # because it and acquire_token_by_auth_code_flow() share same code path.
+
+    def test_acquire_token_silent(self):
+        uid = "foo"
+        utid = "bar"
+        client_id = "my_client_id"
+        scopes = ["some", "scope"]
+        authority_url = "https://login.microsoftonline.com/common"
+        token_cache = msal.TokenCache()
+        token_cache.add({  # Pre-populate the cache
+            "client_id": client_id,
+            "scope": scopes,
+            "token_endpoint": "{}/oauth2/v2.0/token".format(authority_url),
+            "response": build_response(
+                access_token="an expired AT to trigger refresh", expires_in=-99,
+                uid=uid, utid=utid, refresh_token="this is a RT"),
+            })  # The add(...) helper populates correct home_account_id for future searching
+        app = msal.ClientApplication(
+            client_id, authority=authority_url, token_cache=token_cache)
+        with patch.object(app.http_client, "post", return_value=MinimalResponse(
+                status_code=400, text='{"error": "mock"}')) as mocked_method:
+            account = {"home_account_id": "{}.{}".format(uid, utid)}
+            app.acquire_token_silent(["scope"], account)
+            self.assertEqual(
+                "Oid:{}@{}".format(  # Server accepts case-insensitive value
+                    uid, utid),  # It would look like "Oid:foo@bar"
+                mocked_method.call_args[1].get("headers", {}).get('X-AnchorMailbox'),
+                "CSS routing info should be derived from home_account_id")
+
+    def test_acquire_token_by_username_password(self):
+        app = msal.ClientApplication("client_id")
+        username = "johndoe@contoso.com"
+        with patch.object(app.http_client, "post", return_value=MinimalResponse(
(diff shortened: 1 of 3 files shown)


tests/test_ccs.py
```diff
@@ -0,0 +1,53 @@
...
+        state = "foo"
+        flow = app.initiate_auth_code_flow(
+            ["some", "scope"], login_hint="johndoe@contoso.com", state=state)
+        with patch.object(app.http_client, "post", return_value=MinimalResponse(
+                status_code=400, text='{"error": "mock"}')) as mocked_method:
+            app.acquire_token_by_auth_code_flow(flow, {"state": state, "code": "bar"})
+            self.assertEqual(
+                "UPN:johndoe@contoso.com",
+                mocked_method.call_args[1].get("headers", {}).get('X-AnchorMailbox'),
+                "CSS routing info should be derived from login_hint")
```
rayluo, line 26: Hi @trwalke , you have been one of the early adopters who implements the CCS Routing info workitem. Would you please help reviewing this PR? You do not have to review all those implementation code written in Python, but please give a high level confirmation on the behaviors demonstrated by the test cases.

This test case here indicates that, an auth code flow will add an `X-AnchorMailbox: UPN:johndoe@contoso.com` header.

tests/test_ccs.py
```diff
@@ -0,0 +1,53 @@
...
+        app = msal.ClientApplication(
+            client_id, authority=authority_url, token_cache=token_cache)
+        with patch.object(app.http_client, "post", return_value=MinimalResponse(
+                status_code=400, text='{"error": "mock"}')) as mocked_method:
+            account = {"home_account_id": "{}.{}".format(uid, utid)}
+            app.acquire_token_silent(["scope"], account)
+            self.assertEqual(
+                "Oid:{}@{}".format(uid, utid),  # It would look like "Oid:foo@bar"
+                mocked_method.call_args[1].get("headers", {}).get('X-AnchorMailbox'),
+                "CSS routing info should be derived from home_account_id")
```
rayluo, line 52: And this test case here indicates that the acquire_token_silent() would add an `X-AnchorMailbox: Oid:foo@bar` header.

No other test case is added. Because in a follow-up conversation in the [internal design doc](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/3304?_a=files&path=%2FCCS%20Routing%20Headers%2FCCSRoutingHeaders.md&discussionId=31287), we clarified that the native `login_hint` and `username` would already be recognized by CCS.

tests/test_ccs.py
```diff
@@ -0,0 +1,53 @@
...
+    def test_acquire_token_by_auth_code_flow(self):
+        app = msal.ClientApplication("client_id")
+        state = "foo"
+        flow = app.initiate_auth_code_flow(
+            ["some", "scope"], login_hint="johndoe@contoso.com", state=state)
+        with patch.object(app.http_client, "post", return_value=MinimalResponse(
+                status_code=400, text='{"error": "mock"}')) as mocked_method:
+            app.acquire_token_by_auth_code_flow(flow, {"state": state, "code": "bar"})
+            self.assertEqual(
+                "UPN:johndoe@contoso.com",
```
trwalke, line 24: Characters here should be lowercase

tests/test_ccs.py
```diff
@@ -0,0 +1,53 @@
...
+                uid=uid, utid=utid, refresh_token="this is a RT"),
+            })  # The add(...) helper populates correct home_account_id for future searching
+        app = msal.ClientApplication(
+            client_id, authority=authority_url, token_cache=token_cache)
+        with patch.object(app.http_client, "post", return_value=MinimalResponse(
+                status_code=400, text='{"error": "mock"}')) as mocked_method:
+            account = {"home_account_id": "{}.{}".format(uid, utid)}
+            app.acquire_token_silent(["scope"], account)
+            self.assertEqual(
+                "Oid:{}@{}".format(uid, utid),  # It would look like "Oid:foo@bar"
```
trwalke, line 50: Also, lower case here

tests/test_ccs.py
```diff
@@ -0,0 +1,53 @@
...
+import msal
+
+
+class TestCcsRoutingInfoTestCase(unittest.TestCase):
+
+    def test_acquire_token_by_auth_code_flow(self):
+        app = msal.ClientApplication("client_id")
+        state = "foo"
+        flow = app.initiate_auth_code_flow(
+            ["some", "scope"], login_hint="johndoe@contoso.com", state=state)
```
trwalke, line 19: This should also be sent to the /authorize endpoint

tests/test_ccs.py
```diff
@@ -0,0 +1,53 @@
...
+        with patch.object(app.http_client, "post", return_value=MinimalResponse(
+                status_code=400, text='{"error": "mock"}')) as mocked_method:
+            app.acquire_token_by_auth_code_flow(flow, {"state": state, "code": "bar"})
```
trwalke, line 22: You can also request the client info from the /authorize endpoint by adding adding "client_info=1" and use that to create the "Oid:foo@bar" header that can be sent to the token endpoint so there should probably be a test for that too.

tests/test_ccs.py
trwalke: oh ok, i didnt see your comments on the PR. makes sense

tests/test_ccs.py
trwalke: Not needed

tests/test_ccs.py
rayluo: Thanks for spotting this topic. We may want to double check with Ryan on this (I'll ping him offline). Currently, the ["Scenarios" section of the internal design doc](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?version=GBdev&path=%2FCCS%20Routing%20Headers%2FCCSRoutingHeaders.md&_a=preview&anchor=scenarios) contains examples with `UPN:...` and `Oid:...`. I'd assume many readers would follow that sample.

Either Ryan would modify the service-side implementation to accept different cases, or we'd have to: (1) update that design doc, and (2) have different MSAL teams to examine their implementation and make changes.

trwalke: Yes. Actually, casing does not really matter but it is better to keep everything lowercase since headers are usually lowercase anyway. I confirmed this with CCS already

rayluo: To summarize our offline conversation: Travis forwarded me an email which confirmed with CCS team that the *values* of the `X-AnchorMailbox: ValueS` are also treated as case-insensitive. Glad that we concluded this topic. I'll still add a comment in this implementation to call it out.

tests/test_ccs.py
rayluo: We clarified that the value here is case-insensitive. So, no need to change the implementation here.

tests/test_ccs.py
rayluo: Thanks for the hint! The `client_info=1` trick was not mentioned in the body of [internal design](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/3304), but only in a conversation which I overlooked. Now I refactor this PR to request and utilize `client_info`. My test cases are also updated.

rayluo: FYI: I documented that usage into our internal specs here. https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/3964

PR #402, opened 2021-09-01
Fix typos

jiasli:

PR #403, opened 2021-09-07
Fix typo spotted by Lintian

bluca:


bluca:
nvm #402 was there already

PR #411, opened 2021-09-30
mbeliaev/fix_setup

beliaev-maksim:
need to fix the version due to change in cryptography release. See ht…tps://github.com/pyca/cryptography/issues/6344 and https://github.com/pyca/cryptography/blob/main/CHANGELOG.rst

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -75,14 +75,13 @@ setup(
         'requests>=2.0.0,<3',
         'PyJWT[crypto]>=1.0.0,<3',
 
-        'cryptography>=0.6,<4',
+        'cryptography>=0.6,<50',
             # load_pem_private_key() is available since 0.6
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -75,14 +75,13 @@
         'requests>=2.0.0,<3',
         'PyJWT[crypto]>=1.0.0,<3',
 
-        'cryptography>=0.6,<4',
+        'cryptography>=0.6,<50',
```
graingert, line 78:
```suggestion
        'cryptography>=0.6',
```
The recommendation is to not set an upper bound:
https://github.com/pyca/cryptography/issues/5532#issuecomment-720252157

graingert (changes requested):

beliaev-maksim:
closed in favor of #414

PR #414, opened 2021-09-30
Bumping cryptography upper bound to X+3

rayluo:
Even after this change, our CI/CD pipeline is still not yet picking up the latest cryptography 35.0.0, [not until our upstream dependency PyJWT also makes similar change](https://github.com/jpadilla/pyjwt/pull/691).

Thanks @beliaev-maksim and @graingert for bringing this to our attention.

This would close #411, close #413.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -75,14 +75,13 @@ setup(
         'requests>=2.0.0,<3',
         'PyJWT[crypto]>=1.0.0,<3',
 
-        'cryptography>=0.6,<4',
+        'cryptography>=0.6,<38',
             # load_pem_private_key() is available since 0.6
             # https://github.com/pyca/cryptography/blob/master/CHANGELOG.rst#06---2014-09-29
             #
-            # Not sure what should be used as an upper bound here
-            # https://github.com/pyca/cryptography/issues/5532
-            # We will go with "<4" for now, which is also what our another dependency,
-            # pyjwt, currently use.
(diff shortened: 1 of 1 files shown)


beliaev-maksim:
@rayluo 
I am fine to close my PR, just want to mention that it is not the way, how this is managed on GitHub. If you want a change in PR, you explicitly request change, eg I want up bound to be +3, but not opening another PR with same change

rayluo:
> @rayluo I am fine to close my PR, just want to mention that it is not the way, how this is managed on GitHub. If you want a change in PR, you explicitly request change, eg I want up bound to be +3, but not opening another PR with same change

Thanks Maksim for your remind. We should have properly acknowledge your (and Thomas's) contribution, no matter what. My bad. Now we added the attribution in this PR's description.

We understand, and we do usually follow that convention of notifying the PR author to make some *small* change, because it would be convenient for both sides. But in this particular case, our proposed change AND its (imho equally important) inline comments would literally overwrite your entire existing PR. In such a case, we feel like it would seem bureaucratic to say "hey please copy and paste these 4 lines as-is into your existing PR: line 1... line 2... line 3... line 4...". We thought we would better use the same amount of typing to create a new PR, and saving you from a perhaps-boring copy&paste.

Thanks again. Meanwhile, we would do some more testing on our side, while we are also waiting to see how our upstream package PyJWT addresses this.

PR #415, opened 2021-10-01
Broker integration

### LINKED ISSUE #451: Add WAM Support to MSAL for Python, for cross-language consistency
Today WAM support is only in MSAL for .NET.  In order to support WAM in Azure Identity Client Library and meet Azure SDK Design Guidelines (of maintaining feature parity across languages), the support for WAM must exist in all Tier-1 languages (C#, Java, JS, Python) for Azure Identity SDK to GA the support at its layer of the stack.

This GitHub issue tracks the ask for adding WAM support to MSAL in Java, JS, Python, and in Go.  I'll file GitHub issues in each respective repo.

Related: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/643
Related: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/wam
Related: https://github.com/Azure/azure-sdk/issues/3724

/cc @dingmeng-xue @scottaddie @schaabs @xiangyan99 @mccoyp

## Related MSAL Feature Request Issues

MSAL for Python: https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/451
MSAL for JS: https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/4346
MSAL for Java: https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/455
MSAL for Go: https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/284

rayluo:
## What is this?

A broker is a component installed on your device. Broker implicitly gives your device an identity. By using a broker, your device becomes a factor that can satisfy MFA (Multi-factor authentication). This factor would become mandatory if/when a tenant's admin enables a corresponding Conditional Access (CA) policy. The broker's presence allows Microsoft identity platform to have higher confidence that the tokens are being issued to your device, and that is more secure. 

An additional benefit of broker is, it runs as a long-lived process with your device's OS, and maintains its own cache, so that your broker-enabled apps (even CLI) could automatically SSO from a previously established signed-in session.

Currently, broker is available on recent Windows platforms.

In this PR, MSAL Python utilizes the broker to acquire tokens. When enabled, broker behaviors will kick in when your `PublicClientApplication` app calls MSAL Python's `acquire_token_interactive()`, `acquire_token_by_username_password()`, `acquire_token_silent()`, `acquire_token_silent_with_error()`. Most noticeably, the `acquire_token_interactive([...], prompt="select_account")` will trigger a pop-up window, rather than a browser.

API reference docs has also been updated and staged at [here](https://msal-python.readthedocs.io/en/docs-staging/#msal.PublicClientApplication.acquire_token_by_auth_code_flow) (you would need to scroll up a little bit).

## Prerequisite for an app

0. Because this feature branch is not yet officially released with a version number, you will need to clean up your test environment by `pip uninstall msal pymsalruntime -y`, especially when you want to reset your test environment to test the latest changes in this PR.

1. Use a recent version of MSAL Python.
    Currently, this can be achieved by installing the MSAL Python from this feature branch:
    `pip install "git+https://github.com/AzureAD/microsoft-authentication-library-for-python.git@wam"`

2. Install the broker package.
   The [broker package `pymsalruntime`](https://pypi.org/project/pymsalruntime/) needs to be available in your Python environment, ~so that MSAL Python will automatically utilize it~. This can be satisfied by `pip install "msal[broker] @ git+https://github.com/AzureAD/microsoft-authentication-library-for-python.git@wam"`. <sub>In the future (after this PR is merged and shipped), your app can meet the two prerequisites above by a simpler `pip install "msal[broker]>=1.19,<2"` (The actual release version is not yet determined.)</sub>.
   (We do NOT recommend directly install `pymsalruntime`, because its latest version may not have been tested with MSAL Python. Stick with the installation command above so that you will always get the latest version of `pymsalruntime` which has been tested with MSAL Python.)

3. Register one more redirect URI.
   Your app would need to register this one more `redirect_uri` beforehand, in this form: `ms-appx-web://Microsoft.AAD.BrokerPlugin/your_client_id`.

4. Opt-in.
   ~Currently, MSAL Python does not provide an API-level opt-in flag. App developer opts in by declaring needed dependencies AND registering a new redirect URL. Once all prerequisites are met, broker behavior will kick in, otherwise, it will gracefully be fallback to use non-broker behavior. The idea is your app does not need to hardcode your opt-in/opt-out decision, or to implement an opt-in or opt-out setting. The broker functionality can be toggled flexibly, without any source code changes to your app. This approach would make the adoption easier.~
   Opt-in by `PublicClientApplication(..., allow_broker=True)`.

## Action items for a downstream library

* As a library, the prerequisite No. 3 is not applicable to you
* ~You do not need any source code changes to your library~. You would need to somehow expose prerequisite No.4 `allow_broker=True or False` to your app developers.
* There is also one new optional parameter `acquire_token_interactive(..., window=...)`. It is [documented here](https://msal-python.readthedocs.io/en/docs-staging/#msal.PublicClientApplication.acquire_token_interactive).
* Optionally, you may consider also expose a `[broker]` option which internally makes dependency on `msal[broker]`, so that your downstream apps could use `pip install YourLibrary[broker]` to pull in the dependencies for Microsoft identity platform's broker behavior. Or you simply declare `msal[broker]...` as a required dependency.

Regardless, you would still want to use your own test app to test the broker behavior to understand its impact on your library's experience.

## What/How to test

* Test script is available for [download here](https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/wam/tests/msaltest.py) and then run it by `python msaltest.py`.
  (Alternatively, you can use this MSAL Python's interactive sample, add the `allow_broker=True` flag, and then uncomment [this line](https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/1.18.0/sample/interactive_sample.py#L65), to force a broker-powered pop-up window for authentication. Without changing that line, broker would still work, but it would silently use your already-signed-in account in your subsequent test runs, so it will be less tangible.)

* Test this statement: "When an app does not yet ~register the broker-specific redirect_uri~ set `allow_broker=True`, installing the new MSAL Python and the broker package `pymsalruntime` should *not* cause authentication failure".

  | 1. Use recent MSAL Python | 2. Install dependency (The environment contains `pymsalruntime`) | 3. App registers a new `redirect_uri` | 4. Opt-in | Expected behavior (The behaviors implemented in this PR) |
  | --- | --- | --- | --- | --- |
  | No (i.e. using msal<=1.18) | * | * | * | This is the status quo without broker. We use this as a baseline. |
  | Yes (i.e. using this PR) | * | * | False | App should behave the same as before (i.e. above). |
  | Yes (i.e. using this PR) | No | * | True | ~App should behave the same as before (i.e. above).~ Exception will be raised |
  | Yes (i.e. using this PR) | Yes | No | True | MSAL would attempt broker, detect the error, ~emit a warning currently (but we could choose to remove that warning in next commit), and then fall back to use browser/AAD directly~ and raise an exception |
  | Yes (i.e. using this PR) | Yes | Yes | True | MSAL would use broker. Token or error would be returned to the app. More on this below. |

* In particular, when using `acquire_token_interactive(..., prompt="select_account")`, see if you would find any behavior difference between enabling and disabling broker.

  | `allow_broker=False` | `allow_broker=True` | Comment |
  | --- | --- | --- |
  | Successfully obtained an AT | Successfully obtained an AT | This is the happy path |
  | Successfully obtained an AT | Error out | If your app has not met the prerequisite, some errors are expected. That being said, if you encounter some misleading error message, please report back to us. |
  | Error out | Error out | Generally we recommend you to get your app working without broker first. That being said, if you encounter some misleading error message in either case, please report back to us. |

To report an issue, please share with us your test configuration. For example, you can do that by copy and paste the console history of your interaction with our test script `python msaltest.py`.

* Test your app/library can work without a Refresh Token (RT)

  Another subtle change is the absence of refresh token (RT). When an acquire_token request is served by broker, the RT will not be available from the return value. So, if your MSAL-Python-powered app or sdk used to rely on RT to work, you would need to redesign your app/sdk accordingly.

## Roadmap

The following features are not yet supported, but they are expected to be available in the near future. But you should not wait. Please start testing your app with this PR asap, and report back your findings or concerns.

* ~Logging will include broker's debug logs~ Done
* ~Provide support for `max_age` in broker code path~ Done.
* ~Other adjustments in `select_account` and `remove_account()` behavior. But their MSAL Python API are expected to remain unchanged.~ Done.

P.S.:
* Once merged, this PR will resolve #451.
* This is the [internal workitem](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1532789).

CC: @jiasli , @xiangyan99

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -1766,6 +1766,17 @@ class PublicClientApplication(ClientApplication):  # browser app or mobile app
             - A dict containing an "error" key, when token refresh failed.
         """
         data = kwargs.pop("data", {})
+        enable_msa_passthrough = kwargs.pop(  # MUST remove it from kwargs
+            "enable_msa_passthrough",  # Keep it as a hidden param, for now.
+                # OPTIONAL. MSA-Passthrough is a legacy configuration,
+                # needed by a small amount of Microsoft first-party apps,
+                # which would login MSA accounts via ".../organizations" authority.
+                # If you app belongs to this category, AND you are enabling broker,
+                # you would want to enable this flag. Default value is False.
+                # More background of MSA-PT is available from this internal docs:
+                # https://microsoft.sharepoint.com/:w:/t/Identity-DevEx/EatIUauX3c9Ctw1l7AQ6iM8B5CeBZxc58eoQCE0IuZ0VFw?e=tgc3jP&CID=39c853be-76ea-79d7-ee73-f1b2706ede05
+            False
+            ) and data.get("token_type") != "ssh-cert"  # Work around a known issue as of PyMsalRuntime 0.8
         self._validate_ssh_cert_input_data(data)
         if not on_before_launching_ui:
             on_before_launching_ui = lambda **kwargs: None
@@ -1786,21 +1797,6 @@ class PublicClientApplication(ClientApplication):  # browser app or mobile app
                 logger.warning(
                     "Ignoring parameter extra_scopes_to_consent, "
                     "which is not supported by broker")
-            enable_msa_passthrough = kwargs.pop(
-                "enable_msa_passthrough",  # Keep it as a hidden param, for now.
-                    # OPTIONAL. MSA-Passthrough is a legacy configuration,
-                    # needed by a small amount of Microsoft first-party apps,
-                    # which would login MSA accounts via ".../organizations" authority.
-                    # If you app belongs to this category, AND you are enabling broker,
-                    # you would want to enable this flag. Default value is equivalent to False.
-                self.client_id in [
-                    # Experimental: Automatically enable MSA-PT mode for known MSA-PT apps
-                    # More background of MSA-PT is available from this internal docs:
-                    # https://microsoft.sharepoint.com/:w:/t/Identity-DevEx/EatIUauX3c9Ctw1l7AQ6iM8B5CeBZxc58eoQCE0IuZ0VFw?e=tgc3jP&CID=39c853be-76ea-79d7-ee73-f1b2706ede05
-                    "04b07795-8ddb-461a-bbee-02f9e1bf7b46",  # Azure CLI
-                    "04f0c124-f2bc-4f59-8241-bf6df9866bbd",  # Visual Studio
-                    ] and data.get("token_type") != "ssh-cert"  # Work around a known issue as of PyMsalRuntime 0.8
-                )
             return self._acquire_token_interactive_via_broker(
                 scopes,
                 parent_window_handle,
--- a/msal/application.py
+++ b/msal/application.py
@@ -444,9 +444,22 @@ class ClientApplication(object):
             New in version 1.19.0.
 
         :param boolean allow_broker:
-            Brokers provide Single-Sign-On, device identification,
-            and application identification verification.
-            This flag defaults to None, which means MSAL will not utilize broker.
+            A broker is a component installed on your device.
+            Broker implicitly gives your device an identity. By using a broker,
+            your device becomes a factor that can satisfy MFA (Multi-factor authentication).
+            This factor would become mandatory
+            if a tenant's admin enables a corresponding Conditional Access (CA) policy.
+            The broker's presence allows Microsoft identity platform
+            to have higher confidence that the tokens are being issued to your device,
+            and that is more secure.
+
+            An additional benefit of broker is,
+            it runs as a long-lived process with your device's OS,
+            and maintains its own cache,
+            so that your broker-enabled apps (even a CLI)
+            could automatically SSO from a previously established signed-in session.
+
+            This parameter defaults to None, which means MSAL will not utilize a broker.
             If this parameter is set to True,
             MSAL will use the broker whenever possible,
             and automatically fall back to non-broker behavior.
--- a/msal/application.py
+++ b/msal/application.py
@@ -505,10 +505,17 @@ class ClientApplication(object):
             isinstance(self, ConfidentialClientApplication) or self.client_credential)
         if is_confidential_app and allow_broker:
             raise ValueError("allow_broker=True is only supported in PublicClientApplication")
-        self._enable_broker = bool(
-            allow_broker and not is_confidential_app
-            and sys.platform == "win32"
-            and not self.authority.is_adfs and not self.authority._is_b2c)
+        self._enable_broker = False
+        if (allow_broker and not is_confidential_app
+                and sys.platform == "win32"
+                and not self.authority.is_adfs and not self.authority._is_b2c):
+            try:
+                from . import broker  # Trigger Broker's initialization
+                self._enable_broker = True
+            except RuntimeError:
+                logger.exception(
+                    "Broker is unavailable on this platform. "
+                    "We will fallback to non-broker.")
         logger.debug("Broker enabled? %s", self._enable_broker)
 
         self.token_cache = token_cache or TokenCache()
@@ -1072,14 +1079,10 @@ class ClientApplication(object):
         """Sign me out and forget me from token cache"""
         self._forget_me(account)
         if self._enable_broker:
-            try:
-                from .broker import _signout_silently
-            except RuntimeError:  # TODO: TBD
-                logger.debug("Broker is unavailable on this platform. Fallback to non-broker.")
-            else:
-                error = _signout_silently(self.client_id, account["local_account_id"])
-                if error:
-                    logger.debug("_signout_silently() returns error: %s", error)
+            from .broker import _signout_silently
+            error = _signout_silently(self.client_id, account["local_account_id"])
+            if error:
+                logger.debug("_signout_silently() returns error: %s", error)
 
     def _sign_out(self, home_account):
         # Remove all relevant RTs and ATs from token cache
@@ -1312,22 +1315,18 @@ class ClientApplication(object):
                 return self._acquire_token_by_cloud_shell(scopes, data=data)
 
             if self._enable_broker and account is not None and data.get("token_type") != "ssh-cert":
-                try:
-                    from .broker import _acquire_token_silently
-                except RuntimeError:  # TODO: TBD
-                    logger.debug("Broker is unavailable on this platform. Fallback to non-broker.")
-                else:
-                    response = _acquire_token_silently(
-                        "https://{}/{}".format(self.authority.instance, self.authority.tenant),
-                        self.client_id,
-                        account["local_account_id"],
-                        scopes,
-                        claims=_merge_claims_challenge_and_capabilities(
-                            self._client_capabilities, claims_challenge),
-                        correlation_id=correlation_id,
-                        **data)
-                    if response:  # The broker provided a decisive outcome, so we use it
-                        return self._process_broker_response(response, scopes, data)
+                from .broker import _acquire_token_silently
+                response = _acquire_token_silently(
+                    "https://{}/{}".format(self.authority.instance, self.authority.tenant),
+                    self.client_id,
+                    account["local_account_id"],
+                    scopes,
+                    claims=_merge_claims_challenge_and_capabilities(
+                        self._client_capabilities, claims_challenge),
+                    correlation_id=correlation_id,
+                    **data)
+                if response:  # The broker provided a decisive outcome, so we use it
+                    return self._process_broker_response(response, scopes, data)
 
             result = _clean_up(self._acquire_token_silent_by_finding_rt_belongs_to_me_or_my_family(
                 authority, self._decorate_scope(scopes), account,
@@ -1533,24 +1532,20 @@ class ClientApplication(object):
         claims = _merge_claims_challenge_and_capabilities(
                 self._client_capabilities, claims_challenge)
         if self._enable_broker:
-            try:
-                from .broker import _signin_silently
-            except RuntimeError:  # TODO: TBD
-                logger.debug("Broker is unavailable on this platform. Fallback to non-broker.")
-            else:
-                response = _signin_silently(
-                    "https://{}/{}".format(self.authority.instance, self.authority.tenant),
-                    self.client_id,
-                    scopes,  # Decorated scopes won't work due to offline_access
-                    MSALRuntime_Username=username,
-                    MSALRuntime_Password=password,
-                    validateAuthority="no"
-                        if self.authority._validate_authority is False
-                        or self.authority.is_adfs or self.authority._is_b2c
-                        else None,
-                    claims=claims,
-                    )
-                return self._process_broker_response(response, scopes, kwargs.get("data", {}))
+            from .broker import _signin_silently
+            response = _signin_silently(
+                "https://{}/{}".format(self.authority.instance, self.authority.tenant),
+                self.client_id,
+                scopes,  # Decorated scopes won't work due to offline_access
+                MSALRuntime_Username=username,
+                MSALRuntime_Password=password,
+                validateAuthority="no"
+                    if self.authority._validate_authority is False
+                    or self.authority.is_adfs or self.authority._is_b2c
+                    else None,
+                claims=claims,
+                )
+            return self._process_broker_response(response, scopes, kwargs.get("data", {}))
 
         scopes = self._decorate_scope(scopes)
         telemetry_context = self._build_telemetry_context(
@@ -1759,20 +1754,17 @@ class PublicClientApplication(ClientApplication):  # browser app or mobile app
                     "04f0c124-f2bc-4f59-8241-bf6df9866bbd",  # Visual Studio
                     ] and data.get("token_type") != "ssh-cert"  # Work around a known issue as of PyMsalRuntime 0.8
                 )
-            try:
-                return self._acquire_token_interactive_via_broker(
-                    scopes,
-                    parent_window_handle,
-                    enable_msa_passthrough,
-                    claims,
-                    data,
-                    on_before_launching_ui,
-                    prompt=prompt,
-                    login_hint=login_hint,
-                    max_age=max_age,
-                    )
-            except RuntimeError:  # TODO: TBD
-                logger.debug("Broker is unavailable on this platform. Fallback to non-broker.")
+            return self._acquire_token_interactive_via_broker(
+                scopes,
+                parent_window_handle,
+                enable_msa_passthrough,
+                claims,
+                data,
+                on_before_launching_ui,
+                prompt=prompt,
+                login_hint=login_hint,
+                max_age=max_age,
+                )
 
         on_before_launching_ui(ui="browser")
         telemetry_context = self._build_telemetry_context(
--- a/msal/application.py
+++ b/msal/application.py
@@ -1857,6 +1857,7 @@ class PublicClientApplication(ClientApplication):  # browser app or mobile app
         if login_hint and prompt != "select_account":  # OIDC prompts when the user did not sign in
             accounts = self.get_accounts(username=login_hint)
             if len(accounts) == 1:  # Unambiguously proceed with this account
+                logger.debug("Calling broker._acquire_token_silently()")
                 response = _acquire_token_silently(  # When it works, it bypasses prompt
                     authority,
                     self.client_id,
@@ -1868,6 +1869,7 @@ class PublicClientApplication(ClientApplication):  # browser app or mobile app
                     return self._process_broker_response(response, scopes, data)
         # login_hint undecisive or not exists
         if prompt == "none" or not prompt:  # Must/Can attempt _signin_silently()
+            logger.debug("Calling broker._signin_silently()")
             response = _signin_silently(  # Unlike OIDC, it doesn't honor login_hint
                 authority, self.client_id, scopes,
                 validateAuthority=validate_authority,
@@ -1903,7 +1905,8 @@ class PublicClientApplication(ClientApplication):  # browser app or mobile app
                     pass  # It will fall back to the _signin_interactively()
                 else:
                     return self._process_broker_response(response, scopes, data)
-        # Falls back to _signin_interactively()
+
+        logger.debug("Falls back to broker._signin_interactively()")
         on_before_launching_ui(ui="broker")
         response = _signin_interactively(
             authority, self.client_id, scopes,
--- a/msal/application.py
+++ b/msal/application.py
@@ -17,7 +17,7 @@ from .authority import Authority
 from .mex import send_request as mex_send_request
 from .wstrust_request import send_request as wst_send_request
 from .wstrust_response import *
-from .token_cache import TokenCache
+from .token_cache import TokenCache, _get_username
 import msal.telemetry
 from .region import _detect_region
 from .throttled_http_client import ThrottledHttpClient
@@ -25,7 +25,7 @@ from .cloudshell import _is_running_in_cloud_shell
 
 
 # The __init__.py will import this. Not the other way around.
-__version__ = "1.18.0"  # When releasing, also check and bump our dependencies's versions if needed
+__version__ = "1.20.0b1"  # When releasing, also check and bump our dependencies's versions if needed
 
 logger = logging.getLogger(__name__)
 _AUTHORITY_TYPE_CLOUDSHELL = "CLOUDSHELL"
@@ -67,8 +67,12 @@ def _str2bytes(raw):
 
 def _clean_up(result):
     if isinstance(result, dict):
-        result.pop("refresh_in", None)  # MSAL handled refresh_in, customers need not
-    return result
+        return {
+            k: result[k] for k in result
+            if k != "refresh_in"  # MSAL handled refresh_in, customers need not
+            and not k.startswith('_')  # Skim internal properties
+            }
+    return result  # It could be None
 
 
 def _preferred_browser():
@@ -174,6 +178,7 @@ class ClientApplication(object):
                 # when we would eventually want to add this feature to PCA in future.
             exclude_scopes=None,
             http_cache=None,
+            allow_broker=None,
             ):
         """Create an instance of application.
 
@@ -409,6 +414,34 @@ class ClientApplication(object):
             Personally Identifiable Information (PII). Encryption is unnecessary.
 
             New in version 1.16.0.
+
+        :param boolean allow_broker:
+            Brokers provide Single-Sign-On, device identification,
+            and application identification verification.
+            If this parameter is set to True,
+            MSAL will use the broker and return either a token or an error,
+            when your scenario is supported by a broker,
+            otherwise it will automatically fall back to non-broker behavior.
+            This also means you could set this flag as True universally,
+            as long as your app meets the following prerequisite:
+
+            * Installed optional dependency, e.g. ``pip install msal[broker]>=1.20,<2``.
+              (Note that broker is currently only available on Windows 10+)
+
+            * Register a new redirect_uri for your desktop app as:
+              ``ms-appx-web://Microsoft.AAD.BrokerPlugin/your_client_id``
+
+            * Tested your app in following scenarios:
+
+              * Windows 10+
+
+              * PublicClientApplication's following methods::
+                acquire_token_interactive(), acquire_token_by_username_password(),
+                acquire_token_silent() (or acquire_token_silent_with_error()).
+
+              * AAD and MSA accounts (i.e. Non-ADFS, non-B2C)
+
+            New in version 1.20.0.
         """
         self.client_id = client_id
         self.client_credential = client_credential
@@ -467,6 +500,15 @@ class ClientApplication(object):
                     self.http_client, validate_authority=False)
             else:
                 raise
+        is_confidential_app = bool(
+            isinstance(self, ConfidentialClientApplication) or self.client_credential)
+        if is_confidential_app and allow_broker:
+            raise ValueError("allow_broker=True is only supported in PublicClientApplication")
+        self._enable_broker = bool(
+            allow_broker and not is_confidential_app
+            and sys.platform == "win32"
+            and not self.authority.is_adfs and not self.authority._is_b2c)
+        logger.debug("Broker enabled? %s", self._enable_broker)
 
         self.token_cache = token_cache or TokenCache()
         self._region_configured = azure_region
@@ -1028,6 +1070,15 @@ class ClientApplication(object):
     def remove_account(self, account):
         """Sign me out and forget me from token cache"""
         self._forget_me(account)
+        if self._enable_broker:
+            try:
+                from .broker import _signout_silently
+            except RuntimeError:  # TODO: TBD
+                logger.debug("Broker is unavailable on this platform. Fallback to non-broker.")
+            else:
+                error = _signout_silently(self.client_id, account["local_account_id"])
+                if error:
+                    logger.debug("_signout_silently() returns error: %s", error)
 
     def _sign_out(self, home_account):
         # Remove all relevant RTs and ATs from token cache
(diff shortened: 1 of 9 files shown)


msal/wam.py
```diff
@@ -0,0 +1,123 @@
...
+def _signin_interactive():
+    callback_data = _CallbackData()
+    pymsalruntime.signin_interactive(
+        # TODO: Add other input parameters
+        lambda result, callback_data=callback_data: callback_data.complete(result))
+    callback_data.signal.wait()
+    return callback_data.auth_result
+
+def _acquire_token_silently(authority, client_id, account):
+    params = pymsalruntime.MSALRuntimeAuthParameters(client_id, authority)
```
MSamWils, line 45: >params [](http://example.com/codeflow?start=4&length=6)

I think you also need to add the scope.

msal/wam.py
```diff
@@ -0,0 +1,123 @@
...
+    callback_data = _CallbackData()
+    pymsalruntime.signin_interactive(
```
MSamWils, line 47: >signin_interactive [](http://example.com/codeflow?start=18&length=18)

nit: acquire_token_silently

msal/wam.py
```diff
@@ -0,0 +1,123 @@
...
+        authority,  # type: str
+        client_id,  # type: str
+        scopes,  # type: list[str]
+        **kwargs):
+    """MSAL Python's acquire_token_interactive() will call this"""
+    result = _signin_silent(authority, client_id)
+    if not result.get_account():
+        result = _signin_interactive(authority, client_id)
+    if not result.get_account():
+        return {"error": result.get_error()}  # TODO
```
MSamWils, line 95: nit: I am working on a sign_in API which you will be able to use that instead of silently followed by interactive. Thanks

rayluo: Good to know.

As you can see here, the boilerplate code here does not seem too overwhelming. But the bigger topic is that the draft here does not indicate a need to have 2 underlying functions: signin_silently() and signin_interactively(). That seems to be a mismatch of the different programming model. If our further experiments could indeed succeed with only one SignIn function, that would be very helpful in terms of simplifying the programming model.

Look forward to your new signin API.

rayluo: In my latest prototyping, I ended up attempting `signin_interactively()` *only* here. So, @MSamWils you might not need to work out a combined signin API at all. I'll let you know if our future iteration changes course.

msal/wam.py
```diff
@@ -0,0 +1,159 @@
...
+                error.get_context(),  # Available since pymsalruntime 0.0.4
+                error.get_status(), error.get_error_code(), error.get_tag()),
+            }
+    id_token_claims = json.loads(result.get_id_token()) if result.get_id_token() else {}
+    account = result.get_account()
+    assert account.get_account_id() == id_token_claims.get("oid"), "Emperical observation"  # TBD
+    return {k: v for k, v in {
+        "access_token": result.get_access_token(),
+        "expires_in": result.get_access_token_expiry_time(),
+        #"scope": result.get_granted_scopes(),  # TODO
```
MSamWils, line 52: >result.get_granted_scopes() [](http://example.com/codeflow?start=18&length=27)

@rayluo , in MSAL CPP, this is a map, what is the string format you want me to return?

msal/wam.py
```diff
@@ -0,0 +1,159 @@
...
+
+def acquire_token_silent(
+        authority,  # type: str
+        client_id,  # type: str
+        scopes,  # type: list[str]
+        account=None,  # TBD
+        ):
+    scope = " ".join(scopes)
+    if account:
+        wam_account = _read_account_by_id(account["some_sort_of_id"])  # TODO
```
MSamWils, line 154: >some_sort_of_id [](http://example.com/codeflow?start=51&length=15)

This is the localAccountId which is returned from the mid-tier get_accountID API.

msal/wam.py
rayluo: Interesting. ~Why is it a map?~ Anyway. Conceptually, the scopes defined in OAuth2 RFC is a list of strings, but can be represented as one long space-delimited string.

In our case, my understanding is mid-tier data types do not include list. So, let's go with the space-delimited string, for example `"scope1 scope2 scope3"`. (Note that, the scopes are supposed to be orderless, but it does not harm if mid-tier would return a pre-sorted scope list, represented as space-delimited string.)

msal/wam.py
```diff
@@ -0,0 +1,189 @@
...
+        prompt=None,
+        login_hint=None,
+        claims=None,
+        **kwargs):
+    params = pymsalruntime.MSALRuntimeAuthParameters(client_id, authority)
+    params.set_requested_scopes(scopes)
+    params.set_redirect_uri("placeholder")  # pymsalruntime 0.1 requires non-empty str,
+        # the actual redirect_uri will be a value hardcoded by the underlying WAM
+    if prompt:
+        if prompt == "select_account":
```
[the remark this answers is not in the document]
rayluo: Are you referring to the adjustment of the `params.set_select_account_option(pymsalruntime.SelectAccountOption.SHOWLOCALACCOUNTSCONTROL)` mechanism? Yes, MSAL Python would need to change this part, as soon as the underlying changes is shipped in a newer version of MsalRuntime.

msal/wam.py
MSamWils: FYI, the latest pymsalruntime should have the signin API for you to integrate.

msal/wam.py
```diff
@@ -0,0 +1,189 @@
...
+        self.auth_result = None
+
+    def complete(self, auth_result):
+        self.signal.set()
+        self.auth_result = auth_result
+
+
+def _convert_error(error, client_id):
+    context = error.get_context()  # Available since pymsalruntime 0.0.4
+    if "AADSTS50011" in context:  # In WAM, this could happen on both interactive and silent flows
```
MSamWils, line 32: FYI, this is error message that server should return based on their test: https://msazure.visualstudio.com/DefaultCollection/One/_git/226286d5-a05d-446c-a950-4c323c15e588?path=/src/Test/BvtFvt.Microsoft.AzureAD.Sts/OAuth2/OAuth2EndToEndTests.cs&version=GBmaster&line=3629&lineEnd=3632&lineStartColumn=41&lineEndColumn=113&lineStyle=plain&_a=contents

msal/wam.py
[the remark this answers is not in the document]
MSamWils: Yes, I meant pymsalruntime.SelectAccountOption.SHOWLOCALACCOUNTSCONTROL. 
Is that possible for you to remove this first so that when a new version of msalruntime is published, you don't have a build break?
Just to make sure since you may already send this topic branch to Azure CLI team for testing before MSALRuntime publish a new version.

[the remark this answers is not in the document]
rayluo: If we remove it here right now, it will mean one more feature gap in our partner teams' integration test.

But your point of the upcoming breaking change is legit. The right way to solve this will be for MSAL Python to pin the upper bound of MsalRuntime's version to `<0.6`. I'll make that change in my next commit.

msal/broker.py
```diff
@@ -0,0 +1,198 @@
...
+            error.get_status(), error.get_error_code(), error.get_tag()),
+        }
+
+
+def _read_account_by_id(account_id):
+    """Return the callback result which contains the account or error"""
+    callback_data = _CallbackData()
+    pymsalruntime.read_account_by_id(
+        account_id,
+        "correlation_id",
```
MSamWils, line 53: >correlation_id [](http://example.com/codeflow?start=9&length=14)

nit, this should be a guid format and passed from the app.

msal/broker.py
```diff
@@ -0,0 +1,198 @@
...
+        lambda result, callback_data=callback_data: callback_data.complete(result)
+        )
+    callback_data.signal.wait()
+    return callback_data.auth_result
```
MSamWils, line 57: >return callback_data.auth_result [](http://example.com/codeflow?start=4&length=32)

Instead of the auth_result, this can be similar as the _convert_result where we can check the error object first, and if it's null, return the callback_data.auth_result.get_account() if it's not null.

msal/broker.py
```diff
@@ -0,0 +1,198 @@
...
+
+def _convert_result(result, client_id):  # Mimic an on-the-wire response from AAD
+    error = result.get_error()
+    if error:
+        return _convert_error(error, client_id)
+    id_token_claims = json.loads(result.get_id_token()) if result.get_id_token() else {}
+    account = result.get_account()
+    assert account, "Account is expected to be always available"
+    ## Note: As of pymsalruntime 0.1.0, only wam_account_ids property is available
+    #account.get_account_property("wam_account_ids")
```
MSamWils, line 68: Is this still needed?

msal/broker.py
MSamWils: Which is similar as what you did in the other operation functions:
correlation_id or _get_new_correlation_id(),

msal/broker.py
```diff
@@ -0,0 +1,198 @@
...
+    callback_data = _CallbackData()
+    for k, v in kwargs.items():  # This can be used to support domain_hint, max_age, etc.
+        if v is not None:
+            params.set_additional_parameter(k, str(v))
+    pymsalruntime.signin_silently(
+        params,
+        correlation_id or _get_new_correlation_id(),
+        lambda result, callback_data=callback_data: callback_data.complete(result))
+    callback_data.signal.wait()
+    return _convert_result(callback_data.auth_result, client_id)
```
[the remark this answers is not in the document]
MSamWils: Or are you just returning the account Id without caching the account object on your end?

msal/broker.py
```diff
@@ -0,0 +1,198 @@
...
+        if v is not None:
+            params.set_additional_parameter(k, str(v))  # TODO: End-to-end test
+    if claims:
+        params.set_decoded_claims(claims)
+    callback_data = _CallbackData()
+    pymsalruntime.signin_interactively(
+        window or pymsalruntime.get_console_window() or pymsalruntime.get_desktop_window(),  # Since pymsalruntime 0.2+
+        params,
+        correlation_id or _get_new_correlation_id(),
+        login_hint or "",
```
MSamWils, line 137: > or "" [](http://example.com/codeflow?start=18&length=6)

nit: remove the or "" as MSAL CPP will check the login_hint or ignore it if necessary.

msal/broker.py
```diff
@@ -0,0 +1,198 @@
...
+        lambda result, callback_data=callback_data: callback_data.complete(result))
+    callback_data.signal.wait()
+    return _convert_result(callback_data.auth_result, client_id)
+
+
+def _acquire_token_silently(
+        authority, client_id, account_id, scopes, claims=None, correlation_id=None):
+    account = _read_account_by_id(account_id)
```
MSamWils, line 145: >_read_account_by_id(account_id) [](http://example.com/codeflow?start=14&length=31)

As commented above, should we return the account for _read_account_by_id? Instead of checking the error from the caller?

msal/broker.py
```diff
@@ -0,0 +1,198 @@
...
+        account_id,
+        scopes,
+        prompt=None,  # TODO
+        claims=None,
+        extra_scopes_to_consent=None,  # TODO
+        max_age=None,  # TODO
+        correlation_id=None,
+        window=None,
+        **kwargs):
+    raise NotImplementedError("We ended up not currently using this function")
```
MSamWils, line 177: >raise NotImplementedError("We ended up not currently using this function") [](http://example.com/codeflow?start=4&length=74)

May I ask why? what's the blocking issue so that we can resolve it together?

setup.cfg
```diff
@@ -4,3 +4,9 @@ universal=1
 [metadata]
 project_urls =
     Changelog = https://github.com/AzureAD/microsoft-authentication-library-for-python/releases
+
+[options.extras_require]
+# The broker is defined as optional dependency, so that downstream apps can opt in.
+# The opt-in is needed because historically most MSAL Python apps do not have the redirect_uri needed by broker
+#broker =
+#  pymsalruntime>=0.3,<0.6;python_version>='3' and platform_system=='Windows'
```
MSamWils, line 12: >>=0.3,<0.6 [](http://example.com/codeflow?start=16&length=10)

Currently we are at 0.5 and we will rev the version once the removing select_account work is done. So does it make sense to set it <= 0.6?
And does 0.6 includes 0.6.X ?

msal/broker.py
```diff
@@ -0,0 +1,198 @@
+"""This module is an adaptor to the underlying broker.
```
MSamWils, line 1: >This module is an adaptor to the underlying broker. [](http://example.com/codeflow?start=3&length=51)

How about signin API? You should able to call that API with the 0.5 version.

msal/broker.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+def _signin_silently(authority, client_id, scopes, correlation_id=None, **kwargs):
+    params = pymsalruntime.MSALRuntimeAuthParameters(client_id, authority)
+    params.set_requested_scopes(scopes)
+    callback_data = _CallbackData()
+    for k, v in kwargs.items():  # This can be used to support domain_hint, max_age, etc.
+        if v is not None:
+            params.set_additional_parameter(k, str(v))
+    pymsalruntime.signin_silently(
```
MSamWils, line 93: >pymsalruntime.signin_silently [](http://example.com/codeflow?start=4&length=29)

All operation functions will return a handle so that you can use that handle to call cancel.

msal/broker.py
rayluo: Nice catch! Pushing in a new commit. Resolving here.

msal/broker.py
rayluo: Either way would work. The current implementation postpones the `auth_result` check to the caller, so that the caller has more flexibility with its error handling logic. Bottom line: We can always refactor this later, when/if the need arises, because these are not public APIs. #Resolve

msal/broker.py
rayluo: Not strictly necessary but, if you still remember, back then it took you and I quite a while to reverse-engineer the available account properties. Leaving this inline comment here is my way to keep track of this piece of potentially useful information at the place it would likely be used. #Resolve

msal/broker.py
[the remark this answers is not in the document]
rayluo: Yes, this implementation returns the account id only.

msal/broker.py
```diff
@@ -0,0 +1,198 @@
...
+        authority, client_id, scopes,
+        window=None,
+        prompt=None,
+        login_hint=None,
+        claims=None,
+        correlation_id=None,
+        **kwargs):
+    params = pymsalruntime.MSALRuntimeAuthParameters(client_id, authority)
+    params.set_requested_scopes(scopes)
+    params.set_redirect_uri("placeholder")  # pymsalruntime 0.1 requires non-empty str,
```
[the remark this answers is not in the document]
rayluo: Because the underlying WAM behavior happens to be throw away any redirect_uri value the app passes it, I chose to literally use some irrelevant "placeholder" here to make a point.

If we pass in the app's redirect_uri here, it will give future maintainer an illusion that the underlying broker would use that redirect_uri value.

#Resolve

msal/broker.py
```diff
@@ -0,0 +1,198 @@
...
+            if login_hint:
+                # FWIW, AAD's browser interactive flow would honor select_account
+                # and ignore login_hint in such a case.
+                # But pymsalruntime 0.3.x would pop up a meaningless account picker
+                # and then force the account_hint user to re-input password. Not what we want.
+                # https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1744492
+                login_hint = None  # Mimicing the AAD behavior
+                logger.warning("Using both select_account and login_hint is ambiguous. Ignoring login_hint.")
+            params.set_select_account_option(
+                pymsalruntime.SelectAccountOption.SHOWLOCALACCOUNTSCONTROL)
```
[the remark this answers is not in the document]
rayluo: Yes, MSAL CPP *should* handle this case. Has this been fixed in the MSAL CPP side? Please let us know (via future pymsalruntime's release notes?) if a specific version of pymsalruntime fix this, then we will revisit this. Until then, the current implementation here is the correct behavior we want.

In other words, when preparing production-ready deliverable, we don't write code against a future expectation, we write code based on our dependency's current behavior.

msal/broker.py
rayluo: Thanks for the "hint"! (Pun not intended 😄)  This workaround was necessary in pymsalruntime 0.2.0. I was not aware of its fix in pymsalruntime 0.3+. Sending in a new commit now!

msal/broker.py
rayluo: If we did it that way, it would be more complicated on how to surface the inner helper's error return value. Again, as explained above, we can revisit this in the future, when/if the need arises.

msal/broker.py
```diff
@@ -0,0 +1,198 @@
...
+        max_age=None,  # TODO
+        correlation_id=None,
+        window=None,
+        **kwargs):
+    raise NotImplementedError("We ended up not currently using this function")
+    account = _read_account_by_id(account_id)
+    error = account.get_error()
+    if error:
+        return _convert_error(error, client_id)
+    params = pymsalruntime.MSALRuntimeAuthParameters(client_id, authority)
```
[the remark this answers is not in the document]
rayluo: Since we are not currently using this helper, it is already marked as NotImplemented. But since you brought this up, I'm sending a new commit to add the redirect_uri treatment (in the same way that we did in `_signin_interactively()`. #Resolve

setup.cfg
rayluo: > Currently we are at 0.5 and we will rev the version once the removing select_account work is done. So does it make sense to set it <= 0.6?

No, that would defeat the purpose of managing dependencies by their versioning. The current approach is correct: MSAL Python is written on top of the `pymsalruntime >=0.3,<=0.5.x`'s behavior and declare that we will NOT pull in pymsalruntime 0.6+. When and only when pymsalruntime 0.6 becomes available, MSAL Python will make relevant changes in the business logic AND bump the version requirement.

> And does 0.6 includes all versions within 0.6.X ?
It is declared as `<0.6`, so it means all versions with 0.5.X.

P.S.: [Semantic Versioning](https://semver.org/) is the versioning scheme all MSALs use. We strongly recommend PyMsalRuntime would also use this scheme.

msal/broker.py
rayluo: Thanks. Good to know. MSAL Python can consider using it when the need arises. So far, there seems to be unnecessary to add a cancel behavior into MSAL (Python)'s API surface: broker's silent operations seem reasonably responsive, and its interactive operations have a Close button in its pop-up window, anyway.

MSamWils: Any operations that will hit the wire has potential latency, even for the silent request. I will let you decide when to add this feature but the functionality of returning a cancel for cancel operation is there.

msal/broker.py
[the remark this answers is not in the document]
MSamWils: Yes, WAM will throw away but it may not be the same as other broker like Linux. I think the API surface with the APP need to be maintain the same such that the parameters could be used in other broker instead of changing the API signature.

msal/broker.py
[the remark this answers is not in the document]
MSamWils: Let's sync about this.

msal/broker.py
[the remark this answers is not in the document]
rayluo: > but it may not be the same as other broker like Linux.

When pymsalruntime starts to bridge the broker on Linux, our customer would NOT be affected by this line of code, because MSAL Python conservatively declares pymsalruntime dependency only when running on Windows.

This line would fail in OUR development device when we are porting this implementation to Linux. And we will refactor this at that time. That is kind of the TDD effect that we want.

> the API surface with the APP need to be maintain the same ... instead of changing the API signature.

None of this implementation detail would affect MSAL Python's API signature.

We can sync up on this tomorrow. :-)

msal/broker.py
rayluo: We clarified it during our offline sync. MSAL Python's `acquire_token_silent()` api maps to MsalRuntime's `acquire_token_silently()` only. #Resolved

setup.py
```diff
@@ -85,6 +85,14 @@
...
 
         "mock;python_version<'3.3'",
-    ]
+        ],
+    extras_require={  # It does not seem to work if being defined inside setup.cfg
+        "broker": [
+            # The broker is defined as optional dependency,
+            # so that downstream apps can opt in. The opt-in is needed, partially because
+            # most existing MSAL Python apps do not have the redirect_uri needed by broker.
+            "pymsalruntime>=0.3,<0.6;python_version>='3' and platform_system=='Windows'",
```
eugeneoh28, line 94: >python_version>='3' [](http://example.com/codeflow?start=37&length=19)

Does this include py 3.5? We dropped support for 3.5 I believe. @MSamWils as FYI

msal/broker.py
eugeneoh28: Is there aplan to expose this in the future?

msal/broker.py
```diff
@@ -0,0 +1,202 @@
...
+
+PyMsalRuntime currently supports these Windows versions, listed in this MSFT internal link
+https://github.com/AzureAD/microsoft-authentication-library-for-cpp/pull/2406/files
+"""
+from threading import Event
+import json
+import logging
+import uuid
+
+import pymsalruntime  # ImportError would be raised on unsupported platforms such as Windows 8
```
eugeneoh28, line 12: ># ImportError would be raised on unsupported platforms such as Windows 8 [](http://example.com/codeflow?start=22&length=72)

Do you need to catch this error and fallback?

msal/broker.py
```diff
@@ -0,0 +1,202 @@
...
+    # Its API description is available in site-packages/pymsalruntime/PyMsalRuntime.pyi
+
+logger = logging.getLogger(__name__)
```
eugeneoh28, line 15: >logger [](http://example.com/codeflow?start=0&length=6)

Any plans to integrate this logger with the msalruntime logging? We have a logging callback

eugeneoh28: Just an FYI the logger is VERY verbose.... something to consider

msal/broker.py
```diff
@@ -0,0 +1,202 @@
...
+            if login_hint:
+                # FWIW, AAD's browser interactive flow would honor select_account
+                # and ignore login_hint in such a case.
+                # But pymsalruntime 0.3.x would pop up a meaningless account picker
+                # and then force the account_hint user to re-input password. Not what we want.
+                # https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1744492
+                login_hint = None  # Mimicing the AAD behavior
+                logger.warning("Using both select_account and login_hint is ambiguous. Ignoring login_hint.")
+            params.set_select_account_option(
+                pymsalruntime.SelectAccountOption.SHOWLOCALACCOUNTSCONTROL)
```
eugeneoh28, line 124: This will no longer be an available API in the latest release. Lets sync about this sometime. I'll set up a meeting with msalcpp to discuss the remaining prompt types

msal/broker.py
rayluo: Indeed. We let it bubble up from here, and then catch it in the MSAL Python layer.  #Resolved

msal/broker.py
rayluo: > > logger
> 
> Any plans to integrate this logger with the msalruntime logging? We have a logging callback

Definitely. The initial "release note" message of pymsalruntime 0.5.0 did not mention logging callback, so I got that impression that logging callback were not yet available. Now I realize that I am unblocked. I'll work on it.



> Just an FYI the logger is VERY verbose.... something to consider

Hopefully you use different log levels wisely. :-)

msal/broker.py
rayluo: By "latest release", did you mean the already-released 0.5.0? I got an impression that it is coming soon, but not yet in 0.5.0. This PR here currently pins the dependency upper bound as 0.5.x.

msal/broker.py
```diff
@@ -0,0 +1,202 @@
...
+    callback_data.signal.wait()
+    return _convert_result(callback_data.auth_result, client_id)
+
+
+def _acquire_token_interactively(
+        authority,
+        client_id,
+        account_id,
+        scopes,
+        prompt=None,  # TODO
```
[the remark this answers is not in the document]
rayluo: This, and a couple more down below, were some features that we thought we would need to implement. Currently we are using the `signin_interactively()`, so this TODO comments are not really applicable.

msal/broker.py
rayluo: > Is there a plan to expose this in the future?

For the scope and purpose of this PR, we are providing the broker behaviors inside the existing MSAL Python API/pattern, for the existing MSAL Python customers. In the future, we may evaluate when/whether to expose a different calling API/pattern.

setup.py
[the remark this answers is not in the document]
rayluo: Today in this PR, `0.3` lower bound is correct. We will bump it to `0.5` soon, when we pick up the logging callback from pymsalruntime 0.5.

rayluo: Good point. Previously I did not notice [pymsalruntime officially declared "requires: Python >= 3.5"](https://pypi.org/project/pymsalruntime/).

Does pymsalruntime use any Python 3.6+ syntax so that it has zero chance to run on Python <=3.5? Or is it more about our only providing build artifacts (Python wheels) for Python 3.6+? In the latter case, customers could still theoretically use msalruntime with lower version of Python 3, by building from source code.

Regardless, my next commit will align with PyMsalRuntime's declaration on PyPI.

rayluo: I misunderstood that you wanted MSAL Python to declare the lower bound as ">=3.5" (because the PyMsalRuntime itself also declares the same lower bound). Today I notice that [PyMsalRuntime is not built at Python 3.5 at all](https://pypi.org/project/pymsalruntime/#files). So, ok our next commit can bump the lower bound to ">=3.6", assuming that we do not plan to support Python<3.6 customers (which is fine, considering that [those Python versions are already end-of-support](https://endoflife.date/python)).

msal/application.py
```diff
@@ -1552,6 +1655,19 @@ def acquire_token_interactive(
 
             New in version 1.15.
 
+        :param int window:
+            OPTIONAL. If your app is a GUI app running on modern Windows system,
+            and your app opts in to use broker,
+            you are recommended to also provide its window handle,
+            so that the sign in UI window will properly pop up on top of your window.
+
+        :param boolean enable_msa_passthrough:
```
xiangyan99, line 1664: Maybe we want to add its default value which is False I gues?

rayluo: Thanks. Added one more line in the latest rebase.

msal/application.py
```diff
@@ -1560,6 +1676,59 @@ def acquire_token_interactive(
...
+                        enable_msa_pt=enable_msa_passthrough,
+                        **kwargs.get("data", {}))
+                    import pymsalruntime
+                    if prompt == "none" or response.get("_broker_status") not in (
+                            pymsalruntime.Response_Status.Status_AccountUnusable,
+                            pymsalruntime.Response_Status.Status_InteractionRequired,
+                            ):
+                        return self._process_broker_response(
+                            response, scopes, kwargs.get("data", {}))
+                response = _signin_interactively(
```
MSamWils, line 1720: >_signin_interactively [](http://example.com/codeflow?start=27&length=21)

Should we call _signin API instead?

rayluo: Indeed, that was how we did it here between [line 1704 to line 1730](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/415/files#diff-4d32d284fa6acf43894719eff11cb366e5969e0b61593bfd00c68a4cf9c441caR1704-R1730).

msal/broker.py
rayluo: This question has been addressed in the internal design doc.

msal/broker.py
```diff
@@ -0,0 +1,235 @@
...
+            "AADSTS50011" in context  # In WAM, this could happen on both interactive and silent flows
+            or "AADSTS7000218" in context  # This "request body must contain ... client_secret" is just a symptom of current app has no WAM redirect_uri
+            ):
+        raise RedirectUriError(  # This would be seen by either the app developer or end user
+            "MsalRuntime won't work unless this one more redirect_uri is registered to current app: "
+            "ms-appx-web://Microsoft.AAD.BrokerPlugin/{}".format(client_id))
+        # OTOH, AAD would emit other errors when other error handling branch was hit first,
+        # so, the AADSTS50011/RedirectUriError is not guaranteed to happen.
+    return {
+        "error": "broker_error",  # Note: Broker implies your device needs to be compliant.
```
jiasli, line 59: If getting a certificate fails in Cloud Shell, MSAL also raises `broker_error`. 

https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/292e28bbb5a9488de1853a985a994882781068ac/msal/cloudshell.py#L63-L68

Certainly this "broker" (WAM) is different from Cloud Shell's broker (pseudo managed identity). Expected?

PR #433, opened 2021-11-02
Define some Cloud Instance constants and the usage pattern of using them

### LINKED ISSUE #221: Add public convenience string constants for endpoints of each cloud (sovereign and public)
**Context**
Customers have had difficulty finding the right documentation pages:
https://docs.microsoft.com/en-us/azure/azure-government/documentation-government-developer-guide#endpoint-mapping
https://docs.microsoft.com/en-us/azure/china/resources-developer-guide#check-endpoints-in-azure

**See also other libraries exposing these:**
- [obj-c](https://azuread.github.io/microsoft-authentication-library-for-objc/Enums/MSALAzureCloudInstance.html) 
- [.net](https://docs.microsoft.com/en-us/dotnet/api/microsoft.identity.client.azurecloudinstance?view=azure-dotnet)

**Reference:**
https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1063014

**Other:**
https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/258

rayluo:
The new usage pattern is documented in [our doc staging site (please search the keyword "authority (str)") in this page](https://msal-python.readthedocs.io/en/docs-staging/).

You can also try the new syntax by pulling its feature branch:

    pip install git+https://github.com/AzureAD/microsoft-authentication-library-for-python.git@cloud-instances

This is not a breaking change. The old school, low-level url string input is still supported.

This PR, once agreed upon and merged in, will resolve #221.

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -231,8 +231,23 @@ class ClientApplication(object):
 
         :param str authority:
             A URL that identifies a token authority. It should be of the format
-            https://login.microsoftonline.com/your_tenant
-            By default, we will use https://login.microsoftonline.com/common
+            ``https://login.microsoftonline.com/your_tenant``
+            By default, we will use ``https://login.microsoftonline.com/common``
+
+            *Changed in version 1.17*: you can also use predefined constant
+            and a builder like this::
+
+                from msal.authority import (
+                    AuthorityBuilder,
+                    AZURE_US_GOVERNMENT, AZURE_CHINA, AZURE_PUBLIC)
+                my_authority = AuthorityBuilder(AZURE_PUBLIC, "contoso.onmicrosoft.com")
+                # Now you get an equivalent of
+                # "https://login.microsoftonline.com/contoso.onmicrosoft.com"
+
+                # You can feed such an authority to msal's ClientApplication
+                from msal import PublicClientApplication
+                app = PublicClientApplication("my_client_id", authority=my_authority, ...)
+
         :param bool validate_authority: (optional) Turns authority validation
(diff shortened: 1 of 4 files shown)


msal/application.py
```diff
@@ -231,8 +231,23 @@ def __init__(
 
         :param str authority:
             A URL that identifies a token authority. It should be of the format
-            https://login.microsoftonline.com/your_tenant
-            By default, we will use https://login.microsoftonline.com/common
+            ``https://login.microsoftonline.com/your_tenant``
+            By default, we will use ``https://login.microsoftonline.com/common``
```
henrik-me, line 235: >By default, we will use ``https://login.microsoftonline.com/common`` [](http://example.com/codeflow?start=12&length=68)

for CCA it should be tenanted and common should not be allowed?

rayluo: It is true that AAD recommend to avoid "common" for confidential client, yet at the same time they can't really block such a usage due to backward compatibility concern.

We are essentially in the same boat here. Historically the input parameter `authority` is optional and defaults to that ".../common" value. We have, however, long been updated our samples to guide app developers to use their tenant. See [here](https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/1.16.0/sample/confidential_client_secret_sample.py#L5) and [there](https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/1.16.0/sample/confidential_client_certificate_sample.py#L5). In practice, few people would read our API Reference doc (sad but true :-) ) but they will typically start from our samples. So, I think we are good.

PR #435, opened 2021-11-05
Emit warning when common or organizations is used in acquire_token_for_client()

rayluo:
This is the equivalent of MSAL .Net's https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2888

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -1675,6 +1675,11 @@ class ConfidentialClientApplication(ClientApplication):  # server-side web app
             - an error response would contain "error" and usually "error_description".
         """
         # TBD: force_refresh behavior
+        if self.authority.tenant.lower() in ["common", "organizations"]:
+            warnings.warn(
+                "Using /common or /organizations authority "
+                "in acquire_token_for_client() is unreliable. "
+                "Please use a specific tenant instead.", DeprecationWarning)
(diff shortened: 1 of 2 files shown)


msal/application.py
```diff
@@ -1642,6 +1642,11 @@ def acquire_token_for_client(self, scopes, claims_challenge=None, **kwargs):
             - an error response would contain "error" and usually "error_description".
         """
         # TBD: force_refresh behavior
+        if self.authority.tenant.lower() in ["common", "organizations"]:
+            warnings.warn(
+                "Using /common or /organizations authority "
+                "in acquire_token_for_client() is unreliable. "
+                "Please use a specific tenant instead.", DeprecationWarning)
```
trwalke, line 1649: Would you happen to have an aka.ms link with more details?

rayluo: No, we do not currently have any wiki page. Not for this Client Credential Grant flow, at least.

Actually, I checked with the [MSAL .Net's wiki page on this flow](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/Client-credential-flows#code-snippet) before I created this PR. May I say that page does not really provide any extra information on this particular topic? It contains one line "do not use common...", which is kind of hidden as a comment in the code snippet. For that reason, I feel that this exception message above is sufficient.

PR #443, opened 2021-11-26
Descriptive error messages for troubleshooting

rayluo:
This is inspired from, and can also resolve https://github.com/Azure/azure-cli/issues/20507

### DIFF
--- a/msal/authority.py
+++ b/msal/authority.py
@@ -109,7 +109,8 @@ class Authority(object):
             raise ValueError(
                 "Unable to get authority configuration for {}. "
                 "Authority would typically be in a format of "
-                "https://login.microsoftonline.com/your_tenant_name".format(
+                "https://login.microsoftonline.com/your_tenant "
+                "Also please double check your tenant name or GUID is correct.".format(
                 authority_url))
         logger.debug("openid_config = %s", openid_config)
         self.authorization_endpoint = openid_config['authorization_endpoint']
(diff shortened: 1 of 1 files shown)


msal/authority.py
```diff
@@ -170,7 +171,10 @@ def tenant_discovery(tenant_discovery_endpoint, http_client, **kwargs):
     if 400 <= resp.status_code < 500:
         # Nonexist tenant would hit this path
         # e.g. https://login.microsoftonline.com/nonexist_tenant/v2.0/.well-known/openid-configuration
-        raise ValueError("OIDC Discovery endpoint rejects our request")
+        raise ValueError(
+            "OIDC Discovery endpoint rejects our request. Error: {}".format(
+                resp.text  # Expose it as-is b/c OIDC defines no error response format
+            ))
```
jiasli, line 177: Actually, I am rethinking about the error type. Since MSAL raises `ValueError`, CLI can't really tell whether the error comes from MSAL or other places if we use a centralized error handling logic.

Now we handle all `PersistenceError` in a centralized place:

https://github.com/Azure/azure-cli/blob/3dbcafbd273a5fefce793ba1e520dc61d8c468a8/src/azure-cli-core/azure/cli/core/util.py#L140-L151

```py
    elif isinstance(ex, PersistenceError):
        # errno is already in strerror. str(ex) gives duplicated errno.
        az_error = azclierror.CLIInternalError(ex.strerror)
        if ex.errno == 0:
            az_error.set_recommendation(
                "Please report to us via Github: https://github.com/Azure/azure-cli/issues/20278")
        elif ex.errno == -2146893813:
            az_error.set_recommendation(
                "Please report to us via Github: https://github.com/Azure/azure-cli/issues/20231")
        elif ex.errno == -2146892987:
            az_error.set_recommendation(
                "Please report to us via Github: https://github.com/Azure/azure-cli/issues/21010")
```

But for this `ValueError`, we will have to capture it everywhere when MSAL interaction is made:

https://github.com/Azure/azure-cli/blob/1d973cceb38980181eeaa45934497d2148a3e7b2/src/azure-cli-core/azure/cli/core/auth/identity.py#L141-L143

```py
        result = self._msal_app.acquire_token_interactive(
            scopes, prompt='select_account', port=8400 if self._is_adfs else None,
            success_template=success_template, error_template=error_template, **kwargs)
```

If we don't handle the error, the raw exception with traceback is shown, like in https://github.com/Azure/azure-cli/issues/20507.

How about MSAL does the save for this `ValueError` and make a new error type which inherents `ValueError`, such as `AuthorityError`, `MsalValueError`?

PR #445, opened 2021-12-22
Use assertRaisesRegex instead of assertRaisesRegexp for Python 3.12 compatibility

tirkarthi:
The deprecated unittest aliases were removed in python/cpython#28268 . The PR is backwards compatible with Python 2


tests/test_client.py
```diff
@@ -190,11 +190,12 @@ def test_auth_code_flow(self):
             self.assertLoosely(result, lambda: self.assertIn('access_token', result))
 
     def test_auth_code_flow_error_response(self):
-        with self.assertRaisesRegexp(ValueError, "state missing"):
+        assertRaisesRegex = getattr(self, "assertRaisesRegex", self.assertRaisesRegexp)
```
rayluo, line 193: Thanks for going the extra mile to remain backwards compatible with Python 2. However, this line would defeat the purpose in Python 3.11 because it would remove `assertRaisesRegexp`.

tirkarthi: Thanks for the catch. I actually copied it from other place in tests which has similar issue. It should be fixed now in both places.

jiasli:
Ha. This is another evidence showing that Python 2 support should be dropped (https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/406). 🤣

rayluo:
Interestingly, we just added Python 3.11alpha into our test matrix in another PR, the test automation passed without the changes in this PR. Manually trying in a Python 3.11alpha environment, and the `assertRaisesRegexp` is still available.

tirkarthi:
Probably the tests are getting skipped in test_client.py

bgavrilMS:
@rayluo - please let's prioritize community contributions when possible and avoid keeping PRs open for longer than a few days / weeks.

PR #449, opened 2021-12-28
Actionable suggestion for ID token validation failures

rayluo:
This PR resolves #446 (Note: It turns out a PR can not automatically close a Discussion topic. I'll manually close it there.)

### DIFF
--- a/msal/oauth2cli/oidc.py
+++ b/msal/oauth2cli/oidc.py
@@ -44,10 +44,11 @@ def decode_id_token(id_token, client_id=None, issuer=None, nonce=None, now=None)
     err = None  # https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation
     _now = int(now or time.time())
     skew = 120  # 2 minutes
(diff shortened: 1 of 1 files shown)


msal/oauth2cli/oidc.py
```diff
@@ -44,10 +44,11 @@ def decode_id_token(id_token, client_id=None, issuer=None, nonce=None, now=None)
     err = None  # https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation
     _now = int(now or time.time())
     skew = 120  # 2 minutes
+    TIME_SUGGESTION = "Make sure your computer's time is correctly synchronized."
```
jiasli, line 47: Perhaps we also need to mention time zone? The time on the right bottom corner may seem correct, but the time zone may be wrong.

image

rayluo: Good point. I'll change that line to "Make sure your computer's time **(including timezone)** is correctly synchronized.".

PR #454, opened 2022-01-19
Lazy load dependencies

### LINKED ISSUE #423: MSAL can consider using lazy import for `request`, `jwt`
**Describe the bug**

Importing MSAL is can cost ~300ms on Windows due to some heavy libraries like `request` and `jwt`: 

```
python -X importtime -c "import msal"  2>perf.log; tuna .\perf.log
```

image

MSAL can consider using lazy import for these libraries so that importing MSAL can be lightweight. (Azure CLI lazily imports MSAL: https://github.com/Azure/azure-cli/pull/19898)

rayluo:
Lazy load MSAL's dependencies. This resolves #423.

Also remove an old workaround. P.S.: @xiangyan99, after this change, the Azure Identity 1.3.* and 1.4 versions are expected to throw this exception:

    AttributeError: module 'msal.authority' has no attribute 'requests'


msal/authority.py
```diff
@@ -58,11 +53,6 @@ class Authority(object):
     """
     _domains_without_user_realm_discovery = set([])
 
-    @property
-    def http_client(self):  # Obsolete. We will remove this in next major release.
```
xiangyan99, line 62: This looks like a breaking change.

rayluo: Technically yes, but there should be no need for an app to make a dependency (i.e. monkeypatching) MSAL's `authority.http_cache`. Let's use this opportunity to make this intention clear. Internally, This `authority` module uses its own `_authority` member.

xiangyan99 (approved):
LGTM

msal/authority.py
jiasli: What is `_authority` memeber?

jiasli:
Actually Azure CLI already lazily imports MSAL. The issue #423 is _only_ a (possibly not appropriate) suggestion.

I am not sure how much this change will benefit other MSAL users, who use the default `requests` HTTP client. But yea, lazy import `requests` is certainly beneficial to users how choose to use their own HTTP client.

jiasli (approved):

rayluo:
> Actually Azure CLI already lazily imports MSAL. The issue #423 is _only_ a (possibly not appropriate) suggestion.
> 
> I am not sure how much this change will benefit other MSAL users, who use the default `requests` HTTP client. But yea, lazy import `requests` is certainly beneficial to users how choose to use their own HTTP client.

It turns out, lazy loading our dependencies makes MSAL easier to be used in some restricted environment. So, we will still ship this.

jiasli:
Could you share more info about "restricted environment"? If `requests` is not required, perhaps we need to tweak 

https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/636a58ccb4e000335ae251cc374bdc17baa68fc2/setup.py#L75

PR #456, opened 2022-02-08
Actionable exception from ADFS username password flow

rayluo:
Inspired by https://github.com/Azure/azure-cli/issues/21068#issuecomment-1018752545 , this PR wraps most "expected" ADFS username password flow into one `RuntimeError` with actionable message.

That being said, this actionable message is meant for MSAL's app developer. If you are the app developer, you probably want to catch this `RuntimeError` and translate it into your own exception or error message, for example "... consider use 'az login' instead".  @jiasli

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -1417,12 +1417,18 @@ class ClientApplication(object):
             user_realm_result = self.authority.user_realm_discovery(
                 username, correlation_id=headers[msal.telemetry.CLIENT_REQUEST_ID])
             if user_realm_result.get("account_type") == "Federated":
-                response = _clean_up(self._acquire_token_by_username_password_federated(
-                    user_realm_result, username, password, scopes=scopes,
-                    data=data,
-                    headers=headers, **kwargs))
-                telemetry_context.update_telemetry(response)
-                return response
+                try:
+                    response = _clean_up(self._acquire_token_by_username_password_federated(
+                        user_realm_result, username, password, scopes=scopes,
+                        data=data,
+                        headers=headers, **kwargs))
+                except (ValueError, RuntimeError):
+                    raise RuntimeError(
+                        "ADFS is not configured properly. "
+                        "Consider use acquire_token_interactive() instead.")
+                else:
+                    telemetry_context.update_telemetry(response)
+                    return response
         response = _clean_up(self.client.obtain_token_by_username_password(
                 username, password, scope=scopes,
                 headers=headers,
--- a/msal/wstrust_request.py
+++ b/msal/wstrust_request.py
@@ -44,8 +44,8 @@ def send_request(
             soap_action = Mex.ACTION_2005
         elif '/trust/13/usernamemixed' in endpoint_address:
             soap_action = Mex.ACTION_13
-    assert soap_action in (Mex.ACTION_13, Mex.ACTION_2005), (  # A loose check here
-        "Unsupported soap action: %s" % soap_action)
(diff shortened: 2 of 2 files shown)


msal/application.py
```diff
@@ -1417,12 +1417,18 @@ def acquire_token_by_username_password(
...
-                return response
+                try:
+                    response = _clean_up(self._acquire_token_by_username_password_federated(
+                        user_realm_result, username, password, scopes=scopes,
+                        data=data,
+                        headers=headers, **kwargs))
+                except (ValueError, RuntimeError):
+                    raise RuntimeError(
+                        "ADFS is not configured properly. "
+                        "Consider use acquire_token_interactive() instead.")
```
jiasli, line 1428: How about we include the original exception as an "inner" exception?

msal/application.py
```diff
@@ -1417,12 +1417,18 @@ def acquire_token_by_username_password(
...
-                    data=data,
-                    headers=headers, **kwargs))
-                telemetry_context.update_telemetry(response)
-                return response
+                try:
+                    response = _clean_up(self._acquire_token_by_username_password_federated(
+                        user_realm_result, username, password, scopes=scopes,
+                        data=data,
+                        headers=headers, **kwargs))
+                except (ValueError, RuntimeError):
```
jiasli, line 1425: https://github.com/Azure/azure-cli/issues/21068#issue-1109567002 is an `AssertionError`. Shouldn't we catch it as well?

rayluo: Nah.

You may know that the `AssertionError` is not reliable in Python. Previously I used it [here](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/456/files#diff-dd5486a6b885e238cc315b1ad8820acb0a7ccbde95fbcaee8a091825e58115c9L47) anyway, and that's why I also made it clear with an explicit inline comment "a loose check".

This time, since we are trying to use that exception to provide some useful info, I already switched it to `ValueError` in this PR, too.

msal/application.py
rayluo: Currently, the "inner" exception would show up in Python3's error trace already, therefore visible by default.

However, I would expect Azure CLI to do this, to have an actionable guidance most suitable in the Azure CLI's context:

```python
try:
    result = app.acquire_token_by_username_password(...)
except RuntimeError:
    raise CLIError('Username password login failed. You may consider using "az login" instead.')
```

msal/application.py
jiasli: Then perhaps explicitly use Exception Chaining (https://www.python.org/dev/peps/pep-3134/)?

```
raise RuntimeError ... from ex
```

rayluo: Let's just settle with the implicit chaining for now. :-)

So, overall I guess I can deduce that you think the exception messages are fine. I'm going to merge this in tomorrow, and then start the release process.

jiasli (approved):

msal/application.py
jiasli: Won't this message be too vague? This exception chaining pattern is difficult for Azure CLI to get the inner exception message, like that of the original `(ValueError, RuntimeError)` and print it in the terminal.

rayluo: You can probably print the exception you caught. It would contain some (perhaps too much) info. Alas, it is hard to strike a balance here. :-(
```python
try:
    result = app.acquire_token_by_username_password(...)
except RuntimeError as e:
    raise CLIError('Username password login failed: {}. You may consider using "az login" instead.'.format(e))
```

PR #458, opened 2022-02-10
Removes the middle-layer exception

rayluo:
@jiasli Azure CLI can then just catch all exceptions happened in `acquire_token_by_username_password()`, and then chain it with your CLIError:

```python
try:
    app.acquire_token_by_username_password(...)
except Exception as ex:
    raise CLIError('"az login -u ... -p ..." was unsuccessful. You may consider use "az login" instead.') from ex
```

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -1417,18 +1417,12 @@ class ClientApplication(object):
             user_realm_result = self.authority.user_realm_discovery(
                 username, correlation_id=headers[msal.telemetry.CLIENT_REQUEST_ID])
             if user_realm_result.get("account_type") == "Federated":
-                try:
-                    response = _clean_up(self._acquire_token_by_username_password_federated(
-                        user_realm_result, username, password, scopes=scopes,
-                        data=data,
-                        headers=headers, **kwargs))
-                except (ValueError, RuntimeError):
-                    raise RuntimeError(
-                        "ADFS is not configured properly. "
-                        "Consider use acquire_token_interactive() instead.")
-                else:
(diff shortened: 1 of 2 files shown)


rayluo:
Summarizing our offline conversation here:

> Actually, the exception chaining mechanism is designed exactly for such purpose. We shouldn't fear to use it.

Exception chaining was indeed designed for that purpose. But I wonder how such a noble intention would pan out, if python developers have loooong chaining design. IMO, there are 3 conceptual layers: app <-> middle-layer library <-> low level api calls. In reality we can easily see 4, 5, 6+ layers, but still, I believe only the inner-most layer can provide the exact "crime scene", and only the outer-most layer can provide the most relevant actionable mitigation. Everything in between is meaningless noise, especially to end users. A long exception stack trace would just scare them away.

That is why, regardless of when MSAL would drop Python 2 support, I am still not convinced MSAL has strong reason to add/chain the middle-layer exception only for providing MSAL's recommendations. If, unlike in PR #456, MSAL in this PR will always bubble up the inner-most exception, then Azure CLI can just catch it (with or without chaining) and provide your actionable message.

The `WstrustError` sounds like a reasonable idea, though. We will give it some thoughts. The thing is not just about this one `WstructError`. It is about if we go with this direction, how many these kinds of errors we would need eventually, (and if we also potentially backport many historical RuntimeError into different AbcError, XyzError, etc.). Meanwhile, AzureCLI can just catch `ValueError` and `RuntimeError` + the current PR, as a solution for now.

Issue #396, opened 2021-08-24
response_mode in authorization code flow


uta-pmizilin:
Is form_post response_mode supported for authorization code flow (ConfidentialClientApplication)?
Looking at application.py it seems like response_mode parameter is not implemented at all.

msal==1.13.0

rayluo:
It is true that MSAL Python does not currently expose the `response_mode` parameter, because [the use of `response_mode` parameter was "not recommended" per the original specs](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest). But we can add it, per [newer guidance](https://openid.net/specs/oauth-v2-form-post-response-mode-1_0.html).

Marking this issue as an enhancement.

uta-pmizilin:
@rayluo Thanks for a quick response; I assume the newer guidance is the reason why https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-protocols-oidc recommends form_post for web applications:

> response_mode - Recommended
Specifies the method that should be used to send the resulting authorization code back to your app. Can be form_post or fragment. For web applications, we recommend using response_mode=form_post, to ensure the most secure transfer of tokens to your application.

ndrsn:
Has any progress been made on this issue? If not, would a pull request implementing `response_mode` be welcome, and would you have any pointers as to where to start?

PR #469, opened 2022-05-06
implement response_mode

eopeter:
this addresses issue #396 and allows specifying response_mode when using authorization code flow. It checks valid values for response_mode


msal/application.py
```diff
@@ -673,6 +674,16 @@ def initiate_auth_code_flow(
             If the elapsed time is greater than this value,
             Microsoft identity platform will actively re-authenticate the End-User.
 
+        :param str response_mode:
+            OPTIONAL. Specifies the method with which response parameters should be returned.
+            For security purposes, the value should be form_post. In this mode, response parameters
```
rayluo, line 679: The current wording "For security purposes, the value should be form_post" could be perceived as the default behavior is insecure, but a web app powered by MSAL Python does not obtain tokens via query parameter thus is insusceptible to the "[query mode concern](https://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#Security)" in the first place.

How about this?
```suggestion
            The default value is equivalent to ``query``, which is still secure enough in MSAL Python
            (because MSAL Python does not transfer tokens via query parameter in the first place).
            For even better security, we recommend using the value ``form_post``.
            In "form_post" mode, response parameters
```

msal/application.py
```diff
@@ -673,6 +674,16 @@ def initiate_auth_code_flow(
...
+            will be encoded as HTML form values that are transmitted via the HTTP POST method and
+            encoded in the body using the application/x-www-form-urlencoded format.
+            Valid values can be either "form_post" for HTTP POST to callback URI or
+            "query" (the default) for HTTP GET with parameters encoded in query string.
+            More information on possible values
+            `here <https://openid.net/specs/oauth-v2-form-post-response-mode-1_0.html>`
```
rayluo, line 685: That link only describes one possible value, not all of them. Better change it to:

```suggestion
            `here <https://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#ResponseModes>`
            and `here <https://openid.net/specs/oauth-v2-form-post-response-mode-1_0.html#FormPostResponseMode>`
```

msal/application.py
```diff
@@ -673,6 +674,16 @@ def initiate_auth_code_flow(
...
+
```
rayluo, line 686: This entire paragraph is currently being inserted into the middle of other parameter's doc, though. Please move it into around line 690.

msal/oauth2cli/oidc.py
```diff
@@ -184,6 +184,8 @@ def initiate_auth_code_flow(
             # Here we just automatically add it. If the caller do not want id_token,
             # they should simply go with oauth2.Client.
             _scope.append("openid")
+        if kwargs.get("response_mode") not in (None, "form_post", "query"):
+            raise ValueError('only response_mode="form_post" or response_mode="query" is allowed. default is query')
```
rayluo, line 188: The `oidc.py` is designed to be a generic and stable OIDC layer which already accepts all (and even future) `response_mode` values. There is no need to add this check here. If you really want to have such check, please move this part to `application.py` line 712.5. Otherwise, we can just remove them here.

```suggestion
```

tests/test_ccs.py
```diff
@@ -16,7 +16,7 @@ def test_acquire_token_by_auth_code_flow(self):
         app = msal.ClientApplication("client_id")
         state = "foo"
         flow = app.initiate_auth_code_flow(
-            ["some", "scope"], login_hint="johndoe@contoso.com", state=state)
+            ["some", "scope"], login_hint="johndoe@contoso.com", state=state, response_mode="form_post")
```
rayluo, line 19: Most of the existing test cases were designed to test one particular behavior, and showed in the test case's file name and/or its class name.

This test case here was not meant for testing the new `response_mode` behavior. I would recommend to remove this part of change here.

You can add your dedicated test case if you prefer, but I'm even comfortable to NOT add any test case at all, given that (1) the change is minimal; (2) you have already manually tested it in your application; (3) we both reviewed this PR.

```suggestion
            ["some", "scope"], login_hint="johndoe@contoso.com", state=state)
```

rayluo (changes requested):
Thank you, @eopeter, for working out this PR, and even proactively [refining it](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/469/commits/cd79594378728b1e1543a89f5ec085d03e2595fc) for editorial changes.

I leave some comments below.

msal/oauth2cli/oidc.py
```diff
@@ -184,6 +184,7 @@ def initiate_auth_code_flow(
             # Here we just automatically add it. If the caller do not want id_token,
             # they should simply go with oauth2.Client.
             _scope.append("openid")
+
```
rayluo, line 187: Let's also remove this unnecessary line, and call it a day. :-)

PR #471, opened 2022-05-19
MSAL Python 1.18.0b1

rayluo:
* New feature: Optional `initiate_auth_code_flow(..., response_mode="form_post")` to allow the auth code being delivered to your app by form post, which is considered even more secure. (#396, #469)
* New feature: `acquire_token_interactive(..., prompt="none")` can obtain some tokens from within Cloud Shell, without any prompt. (#420)

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -25,7 +25,7 @@ from .cloudshell import _is_running_in_cloud_shell
 
 
 # The __init__.py will import this. Not the other way around.
-__version__ = "1.17.0"  # When releasing, also check and bump our dependencies's versions if needed
+__version__ = "1.18.0b1"  # When releasing, also check and bump our dependencies's versions if needed
 
 logger = logging.getLogger(__name__)
 _AUTHORITY_TYPE_CLOUDSHELL = "CLOUDSHELL"
--- a/msal/application.py
+++ b/msal/application.py
@@ -1417,18 +1417,12 @@ class ClientApplication(object):
             user_realm_result = self.authority.user_realm_discovery(
                 username, correlation_id=headers[msal.telemetry.CLIENT_REQUEST_ID])
             if user_realm_result.get("account_type") == "Federated":
-                try:
-                    response = _clean_up(self._acquire_token_by_username_password_federated(
-                        user_realm_result, username, password, scopes=scopes,
-                        data=data,
-                        headers=headers, **kwargs))
-                except (ValueError, RuntimeError):
-                    raise RuntimeError(
-                        "ADFS is not configured properly. "
-                        "Consider use acquire_token_interactive() instead.")
-                else:
-                    telemetry_context.update_telemetry(response)
-                    return response
+                response = _clean_up(self._acquire_token_by_username_password_federated(
+                    user_realm_result, username, password, scopes=scopes,
+                    data=data,
+                    headers=headers, **kwargs))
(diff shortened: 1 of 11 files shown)


msal/application.py
```diff
@@ -21,13 +21,14 @@
 import msal.telemetry
 from .region import _detect_region
 from .throttled_http_client import ThrottledHttpClient
+from .cloudshell import _is_running_in_cloud_shell
 
 
 # The __init__.py will import this. Not the other way around.
-__version__ = "1.17.0"  # When releasing, also check and bump our dependencies's versions if needed
+__version__ = "1.18.0b1"  # When releasing, also check and bump our dependencies's versions if needed
```
rayluo, line 28: @jiasli , usually this is the only thing that I need to change for a new release.

So, this time we will go with "1.180**b1**", and it should just work?

jiasli: This, it will!

jiasli: Oh, maybe changelog/history notes?

jiasli (approved):

msal/application.py
rayluo: No, we don't have that. What we did is to copy the each release PR's description as release notes. And then all release notes become the history. https://github.com/AzureAD/microsoft-authentication-library-for-python/releases

msal/cloudshell.py
```diff
@@ -0,0 +1,122 @@
...
+except:  # Python 3
+    from urllib.parse import urlparse
+from .oauth2cli.oidc import decode_part
+
+
+logger = logging.getLogger(__name__)
+
+
+def _is_running_in_cloud_shell():
+    return os.environ.get("AZUREPS_HOST_ENVIRONMENT", "").startswith("cloud-shell")
```
jiasli, line 23: Just as an FYI, Azure CLI uses another env var `ACC_CLOUD` to detect if it is run in Cloud Shell:

https://github.com/Azure/azure-cli/blob/f0b5572c4ccafb383de08beb509045145fdc871f/src/azure-cli-core/azure/cli/core/util.py#L688

```py
def in_cloud_console():
    return os.environ.get('ACC_CLOUD', None)
```

```
$ env | grep ACC_CLOUD
ACC_CLOUD=PROD
```

rayluo: I dug this out from my chat messages history.

> 10/19/2021 11:54 AM
> Ray: ... Az CLI and MSAL would need to detect whether they are currently running inside Cloud Shell. I confirmed with Robin that the recommended way is to use AZUREPS_HOST_ENVIRONMENT env var. I am just letting you know, to keep this env var in the future. Otherwise, its removal would become a breaking change for us. :-)

> Edwin: I have previously advised folks to look for the ACC_CLOUD variable. Either should be OK

> Ray: The "problem" of ACC_CLOUD is its content seems to vary by design, therefore the consumers would have to detect its presence, without any keyword in its value to "double check". I would prefer that "AZUREPS_HOST_ENVIRONMENT=cloud-shell/1.0" because we can then use some "value.startswith('cloud-shell')" logic just to be sure.
Either way, we just need your blessing to say "yes, that would become a formal contract that would last forever". 

> Edwin: ACC_CLOUD is a different value per-cloud (Public, Fairfax, etc). Other than that it is static. You can use the AZUREPS one if you like

PR #474, opened 2022-05-30
Fix typo in code

OvervCW:
I stumbled upon this typo while investigating a different issue in this file.


msal/oauth2cli/assertion.py
```diff
@@ -115,7 +115,7 @@ def create_normal_assertion(
                 payload, self.key, algorithm=self.algorithm, headers=self.headers)
             return _str2bytes(str_or_bytes)  # We normalize them into bytes
         except:
-            if self.algorithm.startswith("RS") or self.algorithm.starswith("ES"):
+            if self.algorithm.startswith("RS") or self.algorithm.startswith("ES"):
```
rayluo, line 118: Thanks for the catch! I had no idea that line exists for 4 years and nobody every ran into it. :-)

rayluo (approved):

PR #480, opened 2022-06-02
Managed Identity for Azure VM, App Service, Service Fabric, etc.

### LINKED ISSUE #548: Generate access token for a Managed identity
**Bug description**

I am an Azure costumer. I can't use MSAL to generate an access token for the scope of my app reg using my managed identity.

I have this setup:
- I deployed an Azure function app that is validating access token of an Application registration. It checks that the Client has the correct role.
- I have a compute instance that is assigned a Managed identity. I want to send a HTTP request to my azure function with an access token generated for my Managed Identity.

It might be related to #58 

**To Reproduce**
1. Create an app reg with a custom role
2. Create a managed identity
3. Assign the custom role to the managed identity
4. Create a compute instance and assigned it the managed identity
5. In the compute instance run:
```python
from azure.identity import ManagedIdentityCredential
cred = ManagedIdentityCredential(client_id="<managed-identity-client-id>")
token = cred.get_token("api://<app-reg-client-id>/.default")
```
The code will run indefinitely

**Expected behavior**
After running `token = cred.get_token("api://<app-reg-client-id>/.default")` the token should be generated

**What you see instead**
image
Code is still running and no token is generated.

1.21.0

rayluo:
Note: This is a proof-of-concept, which means there is no guarantee that this behavior will be eventually included into MSAL Python.

There are two new APIs added.

* ~The high level API works for your confidential client which federated with a managed identity.~ This will be moved into a separated PR for its own consideration.
  ```python
   import msal
   cca = msal.ConfidentialClientApplication(
       "my_client_id",
       client_credential=msal.SystemAssignedManagedIdentity(),  # Or it can be an msal.UserAssignedManagedIdentity(client_id="guid")
       ...)
   result = cca.acquire_token_for_client(scopes["scope1", "scope2"])  # It uses scopes
  ```

* The low level API acquires token for managed identity

   ```python
   import msal, requests
   mi = msal.ManagedIdentityClient(
       msal.SystemAssignedManagedIdentity(),  # Or it can be an msal.UserAssignedManagedIdentity(client_id="guid")
       http_client=requests.Session(),  # This is a required parameter
       token_cache=msal.TokenCache(),  # Optional. In your production code, you shall persist a SerializableTokenCache https://msal-python.readthedocs.io/en/latest/#msal.SerializableTokenCache
   )
   result = mi.acquire_token_for_client(resource="resource_abc")  # It uses resource
   ```

More details of the new APIs are available [here](https://msal-python.readthedocs.io/en/mi/#managed-identity).

* In order to test this PR on Azure VM, you would need to:

  1. Create and then ssh into your Azure VM
  1. Install this proof-of-concept by `pip install --force-reinstall "git+https://github.com/AzureAD/microsoft-authentication-library-for-python.git@mi"`
  1. Write your script using the calling pattern above.

* To test this on App Service

  1. [Create your App Service with Python runtime](https://learn.microsoft.com/en-us/azure/app-service/quickstart-python?tabs=flask%2Cwindows%2Cazure-cli%2Clocal-git-deploy%2Cdeploy-instructions-azportal%2Cterminal-bash%2Cdeploy-instructions-zip-azcli#2---create-a-web-app-in-azure)
  2. [SSH into your App Service](https://learn.microsoft.com/en-us/azure/app-service/configure-linux-open-ssh-session)
  3. Follow the last two steps of Azure VM test method

* To test this on Azure Functions

  1. We have not yet tested it end-to-end, but the Managed Identity in Azure Function is expected to be the same as App Service.

* To test this on Azure Automation (we have not tested this)

  1. [Create your Automation account](https://learn.microsoft.com/en-us/azure/automation/automation-create-standalone-account?tabs=azureportal)
  3. You will need to install `msal` package. But it seems Azure Automation only supports [installing a package with its dependencies from PyPI](https://learn.microsoft.com/en-us/azure/automation/python-3-packages?tabs=py3#import-a-package-with-dependencies). This PR is not currently available from PyPI, so, we are unable to test this.
  4. After step 2, you can [create a new Python runbook](https://learn.microsoft.com/en-us/azure/automation/learn/automation-tutorial-runbook-textual-python-3) and test the Managed Identity

* To test this on Service Fabric

  1. We have not yet tested this end-to-end, but you can reference to the
     [test steps in Azure SDK](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/identity/azure-identity/tests/managed-identity-live/service-fabric/service_fabric.md)

Note:

* At the end of [this internal document](https://microsoft.sharepoint.com/:w:/t/aad/devex/EXyuh-XFvj5EuoV6w8UuniUBQNg2EzQR5Ey7d_vbKPGayw?e=tTjoNQ), there are brief descriptions for the 6 variations of MIs.
* Cloud Shell's IMDS is NOT part of this PR, because we already provide a higher level API for it in `https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/420`.

Once merged, this PR will resolve #548. Also, it will officially close #487 as the callback is no longer needed.

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -537,7 +537,11 @@ class ClientApplication(object):
             self.http_client.mount("https://", a)
         self.http_client = ThrottledHttpClient(
             self.http_client,
-            {} if http_cache is None else http_cache,  # Default to an in-memory dict
+            http_cache=http_cache,
+            default_throttle_time=60
+                # The default value 60 was recommended mainly for PCA at the end of
+                # https://identitydivision.visualstudio.com/devex/_git/AuthLibrariesApiReview?version=GBdev&path=%2FService%20protection%2FIntial%20set%20of%20protection%20measures.md&_a=preview
+                if isinstance(self, PublicClientApplication) else 5,
             )
 
         self.app_name = app_name
--- a/msal/managed_identity.py
+++ b/msal/managed_identity.py
@@ -6,6 +6,7 @@ import json
 import logging
 import os
 import socket
+import sys
 import time
 from urllib.parse import urlparse  # Python 3+
 from collections import UserDict  # Python 3+
@@ -507,6 +508,14 @@ def _obtain_token_on_service_fabric(
         raise
 
 
+_supported_arc_platforms_and_their_prefixes = {
+    "linux": "/var/opt/azcmagent/tokens",
+    "win32": os.path.expandvars(r"%ProgramData%\AzureConnectedMachineAgent\Tokens"),
+}
+
+class ArcPlatformNotSupportedError(ManagedIdentityError):
+    pass
+
 def _obtain_token_on_arc(http_client, endpoint, resource):
     # https://learn.microsoft.com/en-us/azure/azure-arc/servers/managed-identity-authentication
     logger.debug("Obtaining token via managed identity on Azure Arc")
@@ -525,7 +534,16 @@ def _obtain_token_on_arc(http_client, endpoint, resource):
             len(challenge) == 2 and challenge[0].lower() == "basic realm"):
         raise ManagedIdentityError(
             "Unrecognizable WWW-Authenticate header: {}".format(resp.headers))
-    with open(challenge[1]) as f:
+    if sys.platform not in _supported_arc_platforms_and_their_prefixes:
+        raise ArcPlatformNotSupportedError(
+            f"Platform {sys.platform} was undefined and unsupported")
+    filename = os.path.join(
+        # This algorithm is documented in an internal doc https://msazure.visualstudio.com/One/_wiki/wikis/One.wiki/233012/VM-Extension-Authoring-for-Arc?anchor=2.-obtaining-tokens
+        _supported_arc_platforms_and_their_prefixes[sys.platform],
+        os.path.splitext(os.path.basename(challenge[1]))[0] + ".key")
+    if os.stat(filename).st_size > 4096:  # Check size BEFORE loading its content
+        raise ManagedIdentityError("Local key file shall not be larger than 4KB")
+    with open(filename) as f:
         secret = f.read()
     response = http_client.get(
         endpoint,
--- a/msal/managed_identity.py
+++ b/msal/managed_identity.py
@@ -330,6 +330,15 @@ def _obtain_token(http_client, managed_identity, resource):
             managed_identity,
             resource,
         )
+    if "MSI_ENDPOINT" in os.environ and "MSI_SECRET" in os.environ:
+        # Back ported from https://github.com/Azure/azure-sdk-for-python/blob/azure-identity_1.15.0/sdk/identity/azure-identity/azure/identity/_credentials/azure_ml.py
+        return _obtain_token_on_machine_learning(
+            http_client,
+            os.environ["MSI_ENDPOINT"],
+            os.environ["MSI_SECRET"],
+            managed_identity,
+            resource,
+        )
     if "IDENTITY_ENDPOINT" in os.environ and "IMDS_ENDPOINT" in os.environ:
         if ManagedIdentity.is_user_assigned(managed_identity):
             raise ManagedIdentityError(  # Note: Azure Identity for Python raised exception too
@@ -346,6 +355,7 @@ def _obtain_token(http_client, managed_identity, resource):
 
 
 def _adjust_param(params, managed_identity):
+    # Modify the params dict in place
     id_name = ManagedIdentity._types_mapping.get(
         managed_identity.get(ManagedIdentity.ID_TYPE))
     if id_name:
@@ -417,6 +427,39 @@ def _obtain_token_on_app_service(
             "error": "invalid_scope",  # Empirically, wrong resource ends up with a vague statusCode=500
             "error_description": "{}, {}".format(
                 payload.get("statusCode"), payload.get("message")),
+            }
+    except json.decoder.JSONDecodeError:
+        logger.debug("IMDS emits unexpected payload: %s", resp.text)
+        raise
+
+def _obtain_token_on_machine_learning(
+    http_client, endpoint, secret, managed_identity, resource,
+):
+    # Could not find protocol docs from https://docs.microsoft.com/en-us/azure/machine-learning
+    # The following implementation is back ported from Azure Identity 1.15.0
+    logger.debug("Obtaining token via managed identity on Azure Machine Learning")
+    params = {"api-version": "2017-09-01", "resource": resource}
+    _adjust_param(params, managed_identity)
+    if params["api-version"] == "2017-09-01" and "client_id" in params:
+        # Workaround for a known bug in Azure ML 2017 API
+        params["clientid"] = params.pop("client_id")
+    resp = http_client.get(
+        endpoint,
+        params=params,
+        headers={"secret": secret},
+        )
+    try:
+        payload = json.loads(resp.text)
+        if payload.get("access_token") and payload.get("expires_on"):
+            return {  # Normalizing the payload into OAuth2 format
+                "access_token": payload["access_token"],
+                "expires_in": int(payload["expires_on"]) - int(time.time()),
+                "resource": payload.get("resource"),
+                "token_type": payload.get("token_type", "Bearer"),
+                }
+        return {
+            "error": "invalid_scope",  # TODO: To be tested
+            "error_description": "{}".format(payload),
             }
     except json.decoder.JSONDecodeError:
         logger.debug("IMDS emits unexpected payload: %s", resp.text)
--- a/msal/managed_identity.py
+++ b/msal/managed_identity.py
@@ -11,7 +11,8 @@ from urllib.parse import urlparse  # Python 3+
 from collections import UserDict  # Python 3+
 from typing import Union  # Needed in Python 3.7 & 3.8
 from .token_cache import TokenCache
-from .throttled_http_client import ThrottledHttpClient
+from .individual_cache import _IndividualCache as IndividualCache
+from .throttled_http_client import ThrottledHttpClientBase, _parse_http_429_5xx_retry_after
 
 
 logger = logging.getLogger(__name__)
@@ -107,6 +108,22 @@ class UserAssignedManagedIdentity(ManagedIdentity):
                 "client_id, resource_id, object_id")
 
 
+class _ThrottledHttpClient(ThrottledHttpClientBase):
+    def __init__(self, http_client, http_cache):
+        super(_ThrottledHttpClient, self).__init__(http_client, http_cache)
+        self.get = IndividualCache(  # All MIs (except Cloud Shell) use GETs
+            mapping=self._expiring_mapping,
+            key_maker=lambda func, args, kwargs: "POST {} hash={} 429/5xx/Retry-After".format(
+                args[0],  # It is the endpoint, typically a constant per MI type
+                _hash(
+                    # Managed Identity flavors have inconsistent parameters.
+                    # We simply choose to hash them all.
+                    str(kwargs.get("params")) + str(kwargs.get("data"))),
+                ),
+            expires_in=_parse_http_429_5xx_retry_after,
+            )(http_client.get)
+
+
 class ManagedIdentityClient(object):
     """This API encapsulates multiple managed identity back-ends:
     VM, App Service, Azure Automation (Runbooks), Azure Function, Service Fabric,
@@ -138,6 +155,7 @@ class ManagedIdentityClient(object):
         *,
         http_client,
         token_cache=None,
+        http_cache=None,
     ):
         """Create a managed identity client.
 
@@ -163,6 +181,10 @@ class ManagedIdentityClient(object):
         :param token_cache:
             Optional. It accepts a :class:`msal.TokenCache` instance to store tokens.
             It will use an in-memory token cache by default.
+
+        :param http_cache:
+            Optional. It has the same characteristics as the
+            :paramref:`msal.ClientApplication.http_cache`.
 
         Recipe 1: Hard code a managed identity for your app::
 
@@ -191,12 +213,21 @@ class ManagedIdentityClient(object):
             token = client.acquire_token_for_client("resource")
         """
         self._managed_identity = managed_identity
-        if isinstance(http_client, ThrottledHttpClient):
-            raise ValueError(
-                # It is a precaution to reject application.py's throttled http_client,
-                # whose cache life on HTTP GET 200 is too long for Managed Identity.
-                "This class does not currently accept a ThrottledHttpClient.")
-        self._http_client = http_client
+        self._http_client = _ThrottledHttpClient(
+            # This class only throttles excess token acquisition requests.
+            # It does not provide retry.
+            # Retry is the http_client or caller's responsibility, not MSAL's.
+            #
+            # FWIW, here is the inconsistent retry recommendation.
+            # 1. Only MI on VM defines exotic 404 and 410 retry recommendations
+            #    ( https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/how-to-use-vm-token#error-handling )
+            #    (especially for 410 which was supposed to be a permanent failure).
+            # 2. MI on Service Fabric specifically suggests to not retry on 404.
+            #    ( https://learn.microsoft.com/en-us/azure/service-fabric/how-to-managed-cluster-managed-identity-service-fabric-app-code#error-handling )
+            http_client.http_client  # Patch the raw (unpatched) http client
+                if isinstance(http_client, ThrottledHttpClientBase) else http_client,
+            {} if http_cache is None else http_cache,  # Default to an in-memory dict
+        )
         self._token_cache = token_cache or TokenCache()
 
     def acquire_token_for_client(self, *, resource):  # We may support scope in the future
--- a/msal/managed_identity.py
+++ b/msal/managed_identity.py
@@ -14,6 +14,7 @@ from typing import Union  # Needed in Python 3.7 & 3.8
 from .token_cache import TokenCache
 from .individual_cache import _IndividualCache as IndividualCache
 from .throttled_http_client import ThrottledHttpClientBase, RetryAfterParser
+from .cloudshell import _is_running_in_cloud_shell
 
 
 logger = logging.getLogger(__name__)
@@ -303,6 +304,35 @@ def _scope_to_resource(scope):  # This is an experimental reasonable-effort appr
     if u.scheme:
         return "{}://{}".format(u.scheme, u.netloc)
     return scope  # There is no much else we can do here
+
+
+APP_SERVICE = object()
+AZURE_ARC = object()
+CLOUD_SHELL = object()  # In MSAL Python, token acquisition was done by
+    # PublicClientApplication(...).acquire_token_interactive(..., prompt="none")
+MACHINE_LEARNING = object()
+SERVICE_FABRIC = object()
+DEFAULT_TO_VM = object()  # Unknown environment; default to VM; you may want to probe
+def get_managed_identity_source():
+    """Detect the current environment and return the likely identity source.
+
+    When this function returns ``CLOUD_SHELL``, you should use
+    :func:`msal.PublicClientApplication.acquire_token_interactive` with ``prompt="none"``
+    to obtain a token.
+    """
+    if ("IDENTITY_ENDPOINT" in os.environ and "IDENTITY_HEADER" in os.environ
+            and "IDENTITY_SERVER_THUMBPRINT" in os.environ
+    ):
+        return SERVICE_FABRIC
+    if "IDENTITY_ENDPOINT" in os.environ and "IDENTITY_HEADER" in os.environ:
+        return APP_SERVICE
+    if "MSI_ENDPOINT" in os.environ and "MSI_SECRET" in os.environ:
+        return MACHINE_LEARNING
+    if "IDENTITY_ENDPOINT" in os.environ and "IMDS_ENDPOINT" in os.environ:
+        return AZURE_ARC
+    if _is_running_in_cloud_shell():
+        return CLOUD_SHELL
+    return DEFAULT_TO_VM
 
 
 def _obtain_token(http_client, managed_identity, resource):
--- a/msal/managed_identity.py
+++ b/msal/managed_identity.py
@@ -12,7 +12,7 @@ from collections import UserDict  # Python 3+
 from typing import Union  # Needed in Python 3.7 & 3.8
 from .token_cache import TokenCache
 from .individual_cache import _IndividualCache as IndividualCache
-from .throttled_http_client import ThrottledHttpClientBase, _parse_http_429_5xx_retry_after
+from .throttled_http_client import ThrottledHttpClientBase, RetryAfterParser
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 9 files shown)


msal/application.py
```diff
@@ -1702,6 +1702,21 @@ def acquire_token_for_client(self, scopes, claims_challenge=None, **kwargs):
             - an error response would contain "error" and usually "error_description".
         """
         # TBD: force_refresh behavior
+        if self.client_credential is None:
```
jmprieur, line 1705: I think you should have ManagedIdentity of something like that, because in the future we'll have the certificateless ones (several type of not client credentials). You could also have Pod Identity when deployed in a container in AKS

msal/imds.py
```diff
@@ -0,0 +1,164 @@
...
+except:  # Python 3
+    from urllib.parse import urlparse
+
+logger = logging.getLogger(__name__)
+
+def _scope_to_resource(scope):  # This is an experimental reasonable-effort approach
+    u = urlparse(scope)
+    if u.scheme:
+        return "{}://{}".format(u.scheme, u.netloc)
+    return scope  # There is no much else we can do here
```
jiasli, line 21: Azure CLI's implementation only removes `/.default`, `/user_impersonation` suffix.

https://github.com/Azure/azure-cli/blob/8474a10fc16b5a168f96dbbe0fd94afc8eecdf4e/src/azure-cli-core/azure/cli/core/auth/util.py#L77-L96

msal/imds.py
```diff
@@ -0,0 +1,164 @@
...
+            If it is None, it means to use a system-assigned managed identity.
+
+        :param token_cache:
+            Optional. It accepts a :class:`msal.TokenCache` instance to store tokens.
+        """
+        self._http_client = http_client
+        self._client_id = client_id
+        self._token_cache = token_cache
+
+    def acquire_token(self, resource):
```
jiasli, line 124: Can we let `ManagedIdentity.acquire_token` take `scopes`, similar to `acquire_token_silent_with_error`?

https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/dd75a284c1b74ecb8e9eb4d2d4598bd602b4638f/msal/application.py#L1224-L1226

msal/imds.py
```diff
@@ -0,0 +1,164 @@
...
+
+def _obtain_token(http_client, resource, client_id=None):
+    if "IDENTITY_ENDPOINT" in os.environ and "IDENTITY_HEADER" in os.environ:
+        return _obtain_token_on_app_service(
+            http_client, os.environ["IDENTITY_ENDPOINT"], os.environ["IDENTITY_HEADER"],
+            resource, client_id=client_id)
+    return _obtain_token_on_azure_vm(http_client, resource, client_id=client_id)
+
+
+def _obtain_token_on_azure_vm(http_client, resource, client_id=None):
```
jiasli, line 32: Besides `client_id`, `object_id` and `msi_res_id` should also be supported:

- https://github.com/Azure/msrestazure-for-python/blob/8e3def8c9927a349596e575de93df76de7a3a2b6/msrestazure/azure_active_directory.py#L468
- https://github.com/MicrosoftDocs/azure-docs/issues/15288#issuecomment-619531505
- https://learn.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#get-a-token-using-http

jiasli, line 32: Besides `client_id`, `object_id` and `msi_res_id` should also be supported. Azure CLI heavily relies on all 3 of them: https://github.com/Azure/azure-cli/blob/060b414d8ac5e4d8f4fc17cc2e23c2154c762226/src/azure-cli-core/azure/cli/core/_profile.py#L195-L227

They are documented at - https://learn.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#get-a-token-using-http (Also see https://github.com/MicrosoftDocs/azure-docs/issues/15288#issuecomment-619531505)

`msrestazure`'s implementation: https://github.com/Azure/msrestazure-for-python/blob/8e3def8c9927a349596e575de93df76de7a3a2b6/msrestazure/azure_active_directory.py#L468

msal/imds.py
rayluo: This experimental helper is currently NOT being used and not really exposed. And we may remove it later.

Do we have some documentation that defines a score-to-scope algorithm? CC: @jmprieur , @bgavrilMS

msal/imds.py
rayluo: See the comment above on `_scope_to_resource()`. Partially due to the immaturity of that helper, this API only accepts resource input.

msal/imds.py
rayluo: Different Managed Identity services support different IDs. `client_id` seems to be the only common ID.

@jiasli , do you know some scenarios that the customer would have to use a different-than-`client_id` id?

msal/imds.py
jiasli: This introduces inconsistency and Azure CLI has to make the conversion by itself.

In my opinion, managed identity should support `scope` finally.

msal/imds.py
jiasli: > do you know some scenarios that the customer would have to use a different-than-`client_id` id?

No, but since the managed identity endpoint supports all 3 of them, we need to do the same. Dropping `object_id` and `msi_res_id` will be a breaking change.

jiasli:
Cloud Shell is just one type of managed identity. Perhaps we should follow the same interface for Cloud Shell?

msal/imds.py
jiasli: > Do we have some documentation that defines a canonical score-to-scope algorithm?

Well, I don't think so. `scope` is like the version 2.0 of `resource`. Converting `scope` back to `resource` certainly losses some information.

bgavrilMS: @rayluo - consider supportting both? The public API can mention resource, but if the user passes in the scope (i.e. graph/.default, just strip the suffix)

rayluo: > @jiasli - for Managed Identity / S2S tokens, there is a 1:1 mapping between scope and resource.
> 
> E.g. **resource**: https://graph.microsoft.com **scope**: https://graph.microsoft.com/.default

Can the `https://resource/user_impersonation` mentioned in @jiasli's first message above also be converted to `https://resource`? (You see, I am really open to add this `scope_to_resource()` helper. I just want to have a precise algorithm to follow.)

rayluo: I gave it more thoughts.

In the absence of a written documentation on "what suffixes can be safely trimmed in a scope-to-resource conversion for AAD", let's follow this [conceptual doc "Scopes and permissions in the Microsoft identity platform"](https://learn.microsoft.com/en-us/azure/active-directory/develop/scopes-oidc) and the empirical evidence in [this video (although it was mainly talking about the .default permission)](https://dev.to/425show/just-what-is-the-default-scope-in-the-microsoft-identity-platform-azure-ad-2o4d), that scopes are typically in the form of `https://resource_uri/permission`. In that sense, the algorithm in this PR's helper is functional.

The `ManagedIdentity.acquire_token(resource)` will remain accepting resource only. Together with the `scope_to_resource()` helper at the developer's disposal, we are technically supporting both. But `acquire_token(resource)` does not automatically call `scope_to_resource()` conversion, because the latter would not know whether the developer would want the trailing `/` (currently, the `scope_to_resource()` helper removes it). It is a responsibility of downstream caller - and ultimately the human developer/user - to decide what resource should be used to satisfy the token audience.

CC: @bgavrilMS

msal/imds.py
rayluo: > Besides client_id, object_id and msi_res_id should also be supported. [Azure CLI heavily relies on all 3 of them](https://github.com/Azure/azure-cli/blob/060b414d8ac5e4d8f4fc17cc2e23c2154c762226/src/azure-cli-core/azure/cli/core/_profile.py#L195-L227)

On a close look, Azure CLI does not surface an "identity id type" concept, therefore the usage pattern is based on inefficient trial-and-error. Don't you consider either adding the identity id type, or [tentatively deprecate](https://docs.python.org/3/library/exceptions.html#PendingDeprecationWarning) the object_id and msi_res_id and see how many users are actually using it?

jiasli: Fortunately, the "inefficient trial-and-error" only occurs during `az login --identity`. Later operations will know the type of the ID and directly specify the type.

Unfortunately, we don't have telemetry for monitoring the usage of `object_id` and `msi_res_id`.

msal/imds.py
rayluo: FWIW, in the latest commit, this was changed to `acquire_token(self, resource=None)`, to prepare for a potential future `scope` optional parameter. But, for now, the `resource` is still a required parameter.

msal/imds.py
rayluo: The latest commit added support for `object_id` and `msi_res_id`.

msal/imds.py
```diff
@@ -0,0 +1,243 @@
...
+        :param str client_id:
+            Optional.
+            It accepts the Client ID (NOT the Object ID) of your user-assigned managed identity.
+            If it is None, it means to use a system-assigned managed identity.
+
+        :param token_cache:
+            Optional. It accepts a :class:`msal.TokenCache` instance to store tokens.
+        """
+        if len(list(filter(bool, [client_id, object_id, mi_res_id]))) > 1:
+            raise ValueError("You can use up to one of these: client_id, object_id, mi_res_id")
```
jiasli, line 182:
```suggestion
            raise ValueError("You can use only one of these: client_id, object_id, mi_res_id")
```

rayluo: I am not an English teacher, but the word "only" seems to have an ambiguity of "choose one and only one". In this context, we mean to say "choose zero or one".

If you do not like "up to", how about I change it to `raise ValueError("You cannot use more than one of these: ...")`?

jiasli: You made a very good point. both `up to`, `cannot use more than` feel a little bit cumbersome to me. `zero or one` is more direct. Perhaps `You can omit id, or pass one of client_id, object_id, mi_res_id`? @dbradish-microsoft is our expert. 😊

rayluo: Sure. Any suggestion is welcome.

@jiasli , for the sake of completeness, we could go with a totally different design here. The little disadvantage is that the calling pattern would be a little bit verbose. But perhaps worth it?

```python
class ManagedIdentityId(object):
    def __init__(self, id):
        self._id = id

class ClientId(ManagedIdentityId):
    pass

class ObjectId(ManagedIdentityId):
    pass

class ResourceId(ManagedIdentityId):
    pass

class ManagedIdentity(object):
    def __init__(self, managed_identity_id, ...):
        if not isinstance(managed_identity_id, (ClientId, ObjectId, ResourceId)):
            raise ValueError("You should provide one of the three objects: ...")
        ...

mi = ManagedIdentity(ClientId("foo"), ...)
```

rayluo: With the latest API change, which is now documented in this PR's description, we can close this topic because the new API simply will not accept multiple IDs. Resolving this conversation.

bgavrilMS:
@rayluo - is this the "public preview" branch? Does MSI support include all 5 sources such as CloudShell and Azure ARC?

bgavrilMS:
Do we have any feedback from Azure SDK / other customers? Is it ok to close on this?

rayluo:
> @rayluo - is this the "public preview" branch? Does MSI support include all 5 sources such as CloudShell and Azure ARC?

Yes.



> Do we have any feedback from Azure SDK / other customers? Is it ok to close on this?

No feedback yet. There seems to be no partner in urgent need of this. We will keep an eye on this, while we are actively working on SLC.

vmarcella:
@rayluo Are there any updates on when this will move out of public preview and into the official MSAL package?

rayluo:
> @rayluo Are there any updates on when this will move out of public preview and into the official MSAL package?

We are getting close. Some of our partner teams are testing it, as we speak. At least one improvement on Arc is identified.

You can also test/use it, and report your findings.

AlmogBentz:
Hello @rayluo , it's been a long time
do I understand correctly that this will push the ability to get creds from managed identity with MSAL?
if so is there a timeline?

bgavrilMS:
> Hello @rayluo , it's been a long time do I understand correctly that this will push the ability to get creds from managed identity with MSAL? if so is there a timeline?

We aim to GA this by end of March @AlmogBentz, but it depends on testing etc. etc. 

Note that Azure SDK (including the Python version) fully supports Managed Identity - https://learn.microsoft.com/en-us/python/api/overview/azure/identity-readme?view=azure-python

bgavrilMS:
@rayluo - for the higher level API bear in mind that: 

- the `api://tokenExchage` magic string is different on other clouds :(
- we should get some consistency in all MSALs 

I recommend you split the higher level API from the lower level one.

msal/managed_identity.py
```diff
@@ -0,0 +1,528 @@
...
+        self.get = IndividualCache(  # All MIs (except Cloud Shell) use GETs
+            mapping=self._expiring_mapping,
+            key_maker=lambda func, args, kwargs: "POST {} hash={} 429/5xx/Retry-After".format(
+                args[0],  # It is the endpoint, typically a constant per MI type
+                _hash(
+                    # Managed Identity flavors have inconsistent parameters.
+                    # We simply choose to hash them all.
+                    str(kwargs.get("params")) + str(kwargs.get("data"))),
+                ),
+            expires_in=_parse_http_429_5xx_retry_after,
```
bgavrilMS: I don't remember the 5xx rule, but please ensure the throttling timeout is not bigger than a few seconds.

bgavrilMS (approved):

msal/managed_identity.py
rayluo: FYI: We are reusing the existing throttling logic defined in MSAL. The actual timeout time is largely controlled by the `Retry-After` header which is determined by the MI service. So, we shall be good here.

bgavrilMS: Sure, but in case of 5xx you default to some constant delay. I remember that the constant was set to 1-2 minutes, which might make sense for public client, but it's unacceptable for CCA.

rayluo: You made a good point about throttling timeout for CCA could have been different than PCA.

Would you also suggest us to change the existing throttling behavior inside MSAL's non-MSI code path? It turns out MSAL Python's PCA and CCA share same default throttling timeout (which is 60 seconds) for years now. :-/

bgavrilMS: Yeah, same for MSAL .NET :(. But I think MSAL.NET will be able to interally retry without hitting the throttling.

arsdragonfly:
Is there an ETA for a new release with this feature?

rayluo:
> Is there an ETA for a new release with this feature?

@arsdragonfly , priorities may change, so we don't provide an ETA. You may go ahead to use this PR's feature branch as a preview (`pip install --force-reinstall "git+https://github.com/AzureAD/microsoft-authentication-library-for-python.git@mi"`); and subscribe/watch this repo to receive its release notifications.

xiangyan99 (approved):

sample/.env.sample.managed_identity
```diff
@@ -0,0 +1,13 @@
+# This sample can be configured to work with Microsoft Entra uD's Managed Identity.
```
jiasli:
```suggestion
# This sample can be configured to work with Microsoft Entra ID's Managed Identity.
```

msal/managed_identity.py
```diff
@@ -0,0 +1,592 @@
...
+        token_cache=None,
+        http_cache=None,
+    ):
+        """Create a managed identity client.
+
+        :param dict managed_identity:
+            It accepts an instance of :class:`SystemAssignedManagedIdentity`
+            or :class:`UserAssignedManagedIdentity`.
+            They are equivalent to a dict with a certain shape,
+            which may be loaded from a json configuration file or an env var,
```
jiasli: When an instance of `SystemAssignedManagedIdentity` is passed as `managed_identity`, PyCharm shows warning:

> Expected type 'dict', got 'SystemAssignedManagedIdentity' instead

msal/managed_identity.py
```diff
@@ -0,0 +1,592 @@
...
+        elif not client_id and resource_id and not object_id:
+            super(UserAssignedManagedIdentity, self).__init__(
+                id_type=self.RESOURCE_ID, identifier=resource_id)
+        elif not client_id and not resource_id and object_id:
+            super(UserAssignedManagedIdentity, self).__init__(
+                id_type=self.OBJECT_ID, identifier=object_id)
+        else:
+            raise ManagedIdentityError(
+                "You shall specify one of the three parameters: "
+                "client_id, resource_id, object_id")
```
jiasli, line 109: It is a common misunderstanding that **user-assigned managed identity requires an ID**. However, this is not true.

When there is no system-assigned managed identity configured and there is only one user-assigned managed identity, you may omit ID to acquire an access token for that user-assigned managed identity. For more details, see https://github.com/Azure/azure-cli/issues/13188

jiasli, line 110: It is a common misunderstanding that **user-assigned managed identity requires an ID**. However, this is not true.

When there is no system-assigned managed identity configured and there is only one user-assigned managed identity, you may omit ID to acquire an access token for that user-assigned managed identity. For more details, see https://github.com/Azure/azure-cli/issues/13188

jiasli (changes requested):

msal/managed_identity.py
bgavrilMS: This a behavior that we want to get away from...

msal/managed_identity.py
```diff
@@ -0,0 +1,592 @@
...
+                "not during runtime. See also "
+                "https://learn.microsoft.com/en-us/azure/service-fabric/configure-existing-cluster-enable-managed-identity-token-service")
+        return _obtain_token_on_service_fabric(
+            http_client,
+            os.environ["IDENTITY_ENDPOINT"],
+            os.environ["IDENTITY_HEADER"],
+            os.environ["IDENTITY_SERVER_THUMBPRINT"],
+            resource,
+        )
+    if "IDENTITY_ENDPOINT" in os.environ and "IDENTITY_HEADER" in os.environ:
```
jiasli, line 356: The detection conditions used here and in `get_managed_identity_source` are very fragile. There is no document claiming these environment variables can uniquely identify a managed identity type.

For example, if a managed identity not on App Service also exposes `IDENTITY_ENDPOINT` and `IDENTITY_HEADER`, the name `_obtain_token_on_app_service` is no longer accurate and may even result in wrong logic. If Azure Arc exposes `IDENTITY_HEADER` one day, it'll trick MSAL into the code for App Service.

msal/managed_identity.py
jiasli: Unless managed identity service rejects the "default user-assigned managed identity" usage, this pattern causes confusion.

msal/managed_identity.py
bgavrilMS: This is a managed identity design issue which we will address with MSIv2, which is currently being designed. The logic here is what Azure SDK have everywhere. This is working in production for some time.

In particular to your comment, you cannot install Azure Arc on any Azure host. Azure Arc is about importing your own VM or another cloud host into Azure.

msal/managed_identity.py
bgavrilMS: @neha-bhargava  to also provide a perspective.

bgavrilMS: Afaik, this is handled by Managed Identity API itself, not by the SDK:

- if the app developer configures a SAMI
- no SAMI exists
- but a single UAMI exists
- then that UAMI is used

msal/managed_identity.py
jiasli: > In particular to your comment, you cannot install Azure Arc on any Azure host. Azure Arc is about importing your own VM or another cloud host into Azure.

I understand. This is merely an example showing the fragility of using available environment variables to detect the platform.

msal/managed_identity.py
jiasli: I understand the user can use `SystemAssignedManagedIdentity` to trigger this behavior, but the name `SystemAssignedManagedIdentity` becomes misleading in that case, as a user-assigned managed identity is actually used.

msal/managed_identity.py
rayluo: Made some changes here in the rebased branch. @jiasli you may try it out, as we are preparing its shipping soon.

msal/managed_identity.py
rayluo: Quoting our offline conclusion.

1. That is the underlying behavior baked into Azure VM's endpoint (but not on any other flavors of Managed Identities). I left [a trace here in MSAL code base](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/480/files#diff-24c0727ff4626c6c535d05c13b61fa4b4a47d6fc4496ec0ceadc734191de19cbR235-R242), which points to [Azure VM's official documentation](https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/how-to-use-vm-token#get-a-token-using-http). We do not like it or advertise it, but, that behavior on the endpoint layer will probably stay forever, until/unless we drop that endpoint and switch to Managed Identity v2 in the future.

2. With our programming model of encouraging/enforcing the choice between SAMI and UAMI, we at least conceptually guide app developers to make that decision upfront. So, when app developer wants to use the only UAMI on their VM, they are expected to create a UAMI with its client_id. They are not supposed to use SAMI. There is no DefaultManagedIdentity concept here. There will be no ambiguity in this usage pattern.
Of course, nothing stops app developer from knowing `#1` and then purposely (ab)use our SAMI interface to get a token that is actually for the UAMI. But we (MSAL) did our part.

msal/managed_identity.py
rayluo: This was discussed in an offline conversation. We agreed that it is not a blocking issue to the release.

msal/managed_identity.py
```diff
@@ -0,0 +1,598 @@
...
+        token_cache=None,
+        http_cache=None,
+    ):
+        """Create a managed identity client.
+
+        :param managed_identity:
+            It accepts an instance of :class:`SystemAssignedManagedIdentity`
+            or :class:`UserAssignedManagedIdentity`.
+            They are equivalent to a dict with a certain shape,
+            which may be loaded from a json configuration file or an env var,
```
jiasli: 1. It's better to capitalize `json` just like how it appears elsewhere.
2. The sentence should end with a period, not a comma.

```suggestion
            which may be loaded from a JSON configuration file or an env var.
```

msal/managed_identity.py
```diff
@@ -0,0 +1,598 @@
...
+            "error_description": resp.text,
+            }
+    except json.decoder.JSONDecodeError:
+        logger.debug("IMDS emits unexpected payload: %s", resp.text)
+        raise
+
+
+_supported_arc_platforms_and_their_prefixes = {
+    "linux": "/var/opt/azcmagent/tokens",
+    "win32": os.path.expandvars(r"%ProgramData%\AzureConnectedMachineAgent\Tokens"),
```
jiasli, line 543: Where are these values documented?

msal/managed_identity.py
```diff
@@ -0,0 +1,598 @@
...
+    if not (  # https://datatracker.ietf.org/doc/html/rfc7617#section-2
+            len(challenge) == 2 and challenge[0].lower() == "basic realm"):
+        raise ManagedIdentityError(
+            "Unrecognizable WWW-Authenticate header: {}".format(resp.headers))
+    if sys.platform not in _supported_arc_platforms_and_their_prefixes:
+        raise ArcPlatformNotSupportedError(
+            f"Platform {sys.platform} was undefined and unsupported")
+    filename = os.path.join(
+        _supported_arc_platforms_and_their_prefixes[sys.platform],
+        os.path.splitext(os.path.basename(challenge[1]))[0] + ".key")
```
jiasli: According to the example at https://learn.microsoft.com/en-us/azure/azure-arc/servers/managed-identity-authentication#acquiring-an-access-token-using-rest-api, shouldn't `challenge[1]` already be the secret file path?

```ps
$secretFile = ($wwwAuthHeader -split "Basic realm=")[1]
...
Write-Host "Secret file path: " $secretFile`n
```


Why do we want to extract the filename then adding back the containing folder and file extension `.key`?

rayluo: Added a comment with an internal doc.

msal/managed_identity.py
rayluo: Added a comment inline with an internal doc.

jiasli:
Congratulations on MSAL's support for managed identity. I am sure our customers will benefit from this great new feature! 🎉

yonzhan:
[heart]         Yong Zhang reacted to your message:
________________________________

PR #481, opened 2022-06-08
Add test case to show that OBO supports SP

rayluo:
* There is no code change to MSAL's implementation in this PR. This means the existing MSAL Python needs no change to support the OBO-for-SP scenario, perhaps since the initial OBO implementation.
* This PR contains changes to MSAL's test case, to demonstrate that how OBO for SP could work. It is a matter of how the mid-tier app chooses to manage the potentially-more-than-one SP "accounts", which can be an open topic for discussion.
* Currently, the test certificate is *not* committed, so, the new test case will not be run in CI. But it was manually run by PR author.

### DIFF
--- a/tests/test_e2e.py
+++ b/tests/test_e2e.py
@@ -649,19 +649,27 @@ class LabBasedTestCase(E2eTestCase):
                 # Here we just test regional apps won't adversely break OBO
             http_client=None,
             ):
-        # 1. An app obtains a token representing a user, for our mid-tier service
-        pca = msal.PublicClientApplication(
-            config_pca["client_id"], authority=config_pca["authority"],
-            azure_region=azure_region,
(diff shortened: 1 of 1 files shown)


bgavrilMS:
We have an E2E in MSAL.NET - please ping the .NET team to discuss about the certificate.

micwoj92:
This branch has conflicts that must be resolved.

rayluo:
> This branch has conflicts that must be resolved.

Thanks. We will be informed of this by github when we revisit this PR (possibly while attempting to solve the test cert topic mentioned above).

Meanwhile, assuming your interest on this topic was driven by your trying to find an OBO example for Service Principal, this PR in its current form (i.e., before merging) still worked as-is, and can serve its current purpose of showcasing how to do OBO with SP.

PR #484, opened 2022-06-17
Use provided authority port when building the tenant discovery endpoint

sarathys:
Add support for authority (STS) endpoint hosted on non-443 port.  This allows Azure CLI to use STS and Graph hosted on different endpoints.  e.g.
```
>  az cloud show -n DevEnv
{
  "endpoints": {
    "activeDirectory": "https://localhost:3001",
    "activeDirectoryDataLakeResourceId": "https://datalake.azs/",
    "activeDirectoryGraphResourceId": "https://localhost:3002/v1.0/",
    "activeDirectoryResourceId": "https://login.azs",
...
  },
  "isActive": false,
  "name": "DevEnv",
  "profile": "latest",
  "suffixes": {
   ...
  }
}

```
This change will allow such configuration for Azure CLI and users can use az login to use it.

See https://github.com/Azure/azure-cli/issues/22553 for the error we get without this change.

### DIFF
--- a/msal/authority.py
+++ b/msal/authority.py
@@ -73,10 +73,9 @@ class Authority(object):
         authority, self.instance, tenant = canonicalize(authority_url)
 
         # extract authority port
-        parsedUrl = urlparse(authority_url)
-        authorityPort = parsedUrl.port
+        authorityPort = authority.port
 
-        if not authorityPort and parsedUrl.scheme == 'https':
+        if not authorityPort and authority.scheme == 'https':
             authorityPort = 443
 
         parts = authority.path.split('/')
--- a/msal/authority.py
+++ b/msal/authority.py
@@ -71,13 +71,6 @@ class Authority(object):
         if isinstance(authority_url, AuthorityBuilder):
             authority_url = str(authority_url)
         authority, self.instance, tenant = canonicalize(authority_url)
-
-        # extract authority port
-        authorityPort = authority.port
-
-        if not authorityPort and authority.scheme == 'https':
-            authorityPort = 443
-
         parts = authority.path.split('/')
         is_b2c = any(self.instance.endswith("." + d) for d in WELL_KNOWN_B2C_HOSTS) or (
             len(parts) == 3 and parts[2].lower().startswith("b2c_"))
@@ -100,7 +93,7 @@ class Authority(object):
             tenant_discovery_endpoint = (
                 'https://{}:{}{}{}/.well-known/openid-configuration'.format(
                     self.instance,
-                    authorityPort,
+                    443 if authority.port is None else authority.port,
                     authority.path,  # In B2C scenario, it is "/tenant/policy"
                     "" if tenant == "adfs" else "/v2.0" # the AAD v2 endpoint
                     ))
--- a/msal/authority.py
+++ b/msal/authority.py
@@ -71,6 +71,14 @@ class Authority(object):
         if isinstance(authority_url, AuthorityBuilder):
             authority_url = str(authority_url)
         authority, self.instance, tenant = canonicalize(authority_url)
+
+        # extract authority port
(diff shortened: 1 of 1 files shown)


msal/authority.py
```diff
@@ -71,6 +71,14 @@ def __init__(self, authority_url, http_client, validate_authority=True):
         if isinstance(authority_url, AuthorityBuilder):
             authority_url = str(authority_url)
         authority, self.instance, tenant = canonicalize(authority_url)
+
+        # extract authority port
+        parsedUrl = urlparse(authority_url)
```
keystroke, line 76: I think "authority" is already parsed url: https://github.com/sarathys/microsoft-authentication-library-for-python/blob/f4036a38be7d0136383390422b98e9fe67082b93/msal/authority.py#L143

msal/authority.py
```diff
@@ -91,8 +99,9 @@ def __init__(self, authority_url, http_client, validate_authority=True):
             tenant_discovery_endpoint = payload['tenant_discovery_endpoint']
         else:
             tenant_discovery_endpoint = (
-                'https://{}{}{}/.well-known/openid-configuration'.format(
+                'https://{}:{}{}{}/.well-known/openid-configuration'.format(
```
keystroke, line 102: need / after port?

msal/authority.py
```diff
@@ -91,8 +99,9 @@ def __init__(self, authority_url, http_client, validate_authority=True):
...
                     self.instance,
+                    authorityPort,
                     authority.path,  # In B2C scenario, it is "/tenant/policy"
                     "" if tenant == "adfs" else "/v2.0" # the AAD v2 endpoint
```
keystroke, line 106: ?? [@msal python team] "tenant" having value "adfs" looks like wrong pattern. In ADFS-based azure system there is still a tenantId value that customer would have / know / use. The different is in constructing URLs, the "authority" for adfs endpoints is /adfs whereas for AAD systems it is /tenantId. So customer will instead pass authority value with /adfs segement on it, rather than "hacking" the tenatnId input to provide "adfs" to "cheese" these URI builder methods. And instead, you may ignore tenantId input parameter for these URL building flows, but it may still be relevant in other areas (e.g. caching tokens with tenantId in some cache key perhaps? etc.)

msal/authority.py
rayluo: It did not have `/` in previous implementation. Looks like the `path` already contains a leading `/`.

msal/authority.py
rayluo: The snippet here is not really about "adfs" vs "tenant_id". It is constructing the tenant discovery uri.

Perhaps you can provide a couple examples on an actual scenario (without any chese :-) ) to show where the problem is?

msal/authority.py
```diff
@@ -71,6 +71,14 @@ def __init__(self, authority_url, http_client, validate_authority=True):
         if isinstance(authority_url, AuthorityBuilder):
             authority_url = str(authority_url)
         authority, self.instance, tenant = canonicalize(authority_url)
+
+        # extract authority port
```
rayluo, line 75: Can we have some introduction in this PR's description, to describe the scenario that you are trying to support?

msal/authority.py
sarathys: correct, path already starts with /.

msal/authority.py
keystroke: Yup so on closer look this looks fine, I got ahead of myself worried about the patterns in CLI which use msal. They have separate parameters for “authority” and “tenantId” and have logic about concatenation them together which causes issues in various parametrized scripts and such.

aluty (approved):

aluty (commented):
Always adding port (:{}) will not cause any problems in default port 443 case?

rayluo (approved):

PR #487, opened 2022-06-29
Implement an app_token_provider callback

rayluo:


rayluo:
No longer necessary, as MSAL will soon support Managed Identity natively in #480.

PR #488, opened 2022-07-25
Test latest 3.11 beta

tonybaloney:
This PR will change the test matrix to use the latest beta of Python 3.11 (currently b4) instead of alpha 5.


rayluo (approved):
Works like a charm! Thanks. Merging.

PR #492, opened 2022-08-08
Implementing known_authority_hosts

rayluo:
Implementing the known_authorities behaviors based on the [internal design](https://microsoft.sharepoint.com/:w:/t/ADAL/ESzbQvzlMuBDjAsqfaHdjNUBsTmwIUR5g17IAY4rvOY_Hw?e=7dv5hZ).

The [unit tests of this PR](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/492/files#diff-1982837bc801e709175e25839e1046a445df689d7a2e6309ccb6856e350dee7eR142-R223) is almost as readable as plain English, and are considered as generic acceptance tests for this feature.

This PR also contains the ["api reference documentation" for the newly introduced `known_authority_hosts` parameter](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/492/files#diff-4d32d284fa6acf43894719eff11cb366e5969e0b61593bfd00c68a4cf9c441caR165-R185).

When merged, this PR will close this [internal workitem](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/823221).

---

This entire PR has been shelved, because later we discovered some [new requirements for Azure Stack scenarios (internal link)](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=/AuthorityValidation/authority_validation_concepts_and_api.md&_a=preview&anchor=skip-authority-validation-and-alias-metadata-behaviors-without-a-set-of-predefined-authorities), so we proceeded with #496.

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -2,9 +2,9 @@ import functools
 import json
 import time
 try:  # Python 2
-    from urlparse import urljoin
+    from urlparse import urljoin, urlparse
 except:  # Python 3
-    from urllib.parse import urljoin
+    from urllib.parse import urljoin, urlparse
 import logging
 import sys
 import warnings
@@ -13,7 +13,7 @@ import os
 
 from .oauth2cli import Client, JwtAssertionCreator
 from .oauth2cli.oidc import decode_part
-from .authority import Authority
+from .authority import Authority, WORLD_WIDE
 from .mex import send_request as mex_send_request
 from .wstrust_request import send_request as wst_send_request
 from .wstrust_response import *
@@ -146,7 +146,6 @@ class _ClientWithCcsRoutingInfo(Client):
 
 
 class ClientApplication(object):
-
     ACQUIRE_TOKEN_SILENT_ID = "84"
     ACQUIRE_TOKEN_BY_REFRESH_TOKEN = "85"
     ACQUIRE_TOKEN_BY_USERNAME_PASSWORD_ID = "301"
@@ -159,6 +158,48 @@ class ClientApplication(object):
     REMOVE_ACCOUNT_ID = "903"
 
     ATTEMPT_REGION_DISCOVERY = True  # "TryAutoDetect"
+
+    _known_authority_hosts = None
+
+    @classmethod
+    def set_known_authority_hosts(cls, known_authority_hosts):
+        """Declare a list of hosts which you allow MSAL to operate with.
+
+        If your app operates with some authorities that you know and own,
+        such as some ADFS or B2C or private cloud,
+        it is recommended and sometimes required that you declare them here,
+        so that MSAL will use your authorities without discovery,
+        and reject most of the other undefined authorities.
+
+        ``known_authority_hosts`` is meant to be a static and per-deployment setting.
+        This classmethod shall be called at most once,
+        during your entire app's starting-up,
+        before your initializing any ``PublicClientApplication`` or
+        ``ConfidentialClientApplication`` instance(s).
+
+        :param list[str] known_authority_hosts:
+            Authorities that you known, for example::
+
+                [
+                    "contoso.com",  # Your own domain
+                    "login.azs",  # This can be a private cloud
+                ]
+
+        New in version 1.19
+        """
+        new_input = frozenset(known_authority_hosts)
+        if (cls._known_authority_hosts is not None
+                and cls._known_authority_hosts != new_input):
+            raise ValueError(
+                "The known_authority_hosts are considered static. "
+                "Once configured, they should not be changed.")
+        cls._known_authority_hosts = new_input
+        logger.debug('known_authority_hosts is set to %s', known_authority_hosts)
+
+    def _union_known_authority_hosts(cls, url=None, host=None):
+        host = host if host else urlparse(url).netloc.split(":")[0]
+        return (cls._known_authority_hosts.union([host])
+            if cls._known_authority_hosts else frozenset([host]))
 
     def __init__(
             self, client_id,
@@ -453,9 +494,12 @@ class ClientApplication(object):
 
         # Here the self.authority will not be the same type as authority in input
         try:
+            authority_to_use = authority or "https://{}/common/".format(WORLD_WIDE)
             self.authority = Authority(
-                authority or "https://login.microsoftonline.com/common/",
-                self.http_client, validate_authority=validate_authority)
+                authority_to_use,
+                self.http_client, validate_authority=validate_authority,
+                known_authority_hosts=self.__class__._known_authority_hosts,
+                )
         except ValueError:  # Those are explicit authority validation errors
             raise
         except Exception:  # The rest are typically connection errors
@@ -463,8 +507,11 @@ class ClientApplication(object):
                 # Since caller opts in to use region, here we tolerate connection
                 # errors happened during authority validation at non-region endpoint
                 self.authority = Authority(
-                    authority or "https://login.microsoftonline.com/common/",
-                    self.http_client, validate_authority=False)
+                    authority_to_use,
+                    self.http_client,
+                    known_authority_hosts=self._union_known_authority_hosts(
+                        url=authority_to_use),
+                    )
             else:
                 raise
 
@@ -534,10 +581,12 @@ class ClientApplication(object):
                     "sts.windows.net",
                     )
                 else "{}.{}".format(region_to_use, central_authority.instance))
-            return Authority(
+            return Authority(  # The central_authority has already been validated
                 "https://{}/{}".format(regional_host, central_authority.tenant),
                 self.http_client,
-                validate_authority=False)  # The central_authority has already been validated
+                known_authority_hosts=self._union_known_authority_hosts(
+                    host=regional_host),
+                )
         return None
 
     def _build_client(self, client_credential, authority, skip_regional_client=False):
@@ -789,7 +838,8 @@ class ClientApplication(object):
         # Multi-tenant app can use new authority on demand
         the_authority = Authority(
             authority,
-            self.http_client
+            self.http_client,
+            known_authority_hosts=self.__class__._known_authority_hosts,
             ) if authority else self.authority
 
         client = _ClientWithCcsRoutingInfo(
@@ -1012,14 +1062,21 @@ class ClientApplication(object):
             }
         return list(grouped_accounts.values())
 
+    def _get_instance_metadata(self):  # This exists so it can be mocked in unit test
+        resp = self.http_client.get(
+            "https://login.microsoftonline.com/common/discovery/instance?api-version=1.1&authorization_endpoint=https://login.microsoftonline.com/common/oauth2/authorize",
+            headers={'Accept': 'application/json'})
+        resp.raise_for_status()
+        return json.loads(resp.text)['metadata']
+
     def _get_authority_aliases(self, instance):
+        if self.authority._is_known_to_developer:
+            # Then it is an ADFS/B2C/known_authority_hosts situation
+            # which may not reach the central endpoint, so we skip it.
+            return []
         if not self.authority_groups:
-            resp = self.http_client.get(
-                "https://login.microsoftonline.com/common/discovery/instance?api-version=1.1&authorization_endpoint=https://login.microsoftonline.com/common/oauth2/authorize",
-                headers={'Accept': 'application/json'})
-            resp.raise_for_status()
             self.authority_groups = [
-                set(group['aliases']) for group in json.loads(resp.text)['metadata']]
+                set(group['aliases']) for group in self._get_instance_metadata()]
         for group in self.authority_groups:
             if instance in group:
                 return [alias for alias in group if alias != instance]
@@ -1189,7 +1246,8 @@ class ClientApplication(object):
             the_authority = Authority(
                 "https://" + alias + "/" + self.authority.tenant,
                 self.http_client,
-                validate_authority=False)
+                known_authority_hosts=self._union_known_authority_hosts(host=alias),
+                )
             result = self._acquire_token_silent_from_cache_and_possibly_refresh_it(
                 scopes, account, the_authority, force_refresh=force_refresh,
                 claims_challenge=claims_challenge,
--- a/msal/application.py
+++ b/msal/application.py
@@ -159,47 +159,10 @@ class ClientApplication(object):
 
     ATTEMPT_REGION_DISCOVERY = True  # "TryAutoDetect"
 
-    _known_authority_hosts = None
-
-    @classmethod
-    def set_known_authority_hosts(cls, known_authority_hosts):
-        """Declare a list of hosts which you allow MSAL to operate with.
-
-        If your app operates with some authorities that you know and own,
-        such as some ADFS or B2C or private cloud,
-        it is recommended and sometimes required that you declare them here,
-        so that MSAL will use your authorities without discovery,
-        and reject most of the other undefined authorities.
-
-        ``known_authority_hosts`` is meant to be a static and per-deployment setting.
-        This classmethod shall be called at most once,
-        during your entire app's starting-up,
-        before your initializing any ``PublicClientApplication`` or
-        ``ConfidentialClientApplication`` instance(s).
-
-        :param list[str] known_authority_hosts:
-            Authorities that you known, for example::
-
-                [
-                    "contoso.com",  # Your own domain
-                    "login.azs",  # This can be a private cloud
-                ]
-
-        New in version 1.19
-        """
-        new_input = frozenset(known_authority_hosts)
-        if (cls._known_authority_hosts is not None
-                and cls._known_authority_hosts != new_input):
-            raise ValueError(
-                "The known_authority_hosts are considered static. "
-                "Once configured, they should not be changed.")
-        cls._known_authority_hosts = new_input
-        logger.debug('known_authority_hosts is set to %s', known_authority_hosts)
-
-    def _union_known_authority_hosts(cls, url=None, host=None):
+    def _union_known_authority_hosts(self, url=None, host=None):
         host = host if host else urlparse(url).netloc.split(":")[0]
-        return (cls._known_authority_hosts.union([host])
-            if cls._known_authority_hosts else frozenset([host]))
+        return (self._known_authority_hosts.union([host])
+            if self._known_authority_hosts else frozenset([host]))
 
     def __init__(
             self, client_id,
@@ -215,6 +178,7 @@ class ClientApplication(object):
                 # when we would eventually want to add this feature to PCA in future.
             exclude_scopes=None,
             http_cache=None,
+            known_authority_hosts=None,
             ):
         """Create an instance of application.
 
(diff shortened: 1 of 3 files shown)


msal/application.py
```diff
@@ -160,6 +159,39 @@ class ClientApplication(object):
...
+        so that MSAL will use them without any validation.
+
+        This setting shall be configured at most once, during your app start-up,
+        rather than being changed to different authorities during runtime.
+
+        :param list[str] known_authorities:
+            Authorities that you known, for example::
+
+                [
+                    "contoso.com",  # Your own domain
```
bgavrilMS, line 180: The variable name is "known_authorities" but these are not authorities, these are environments. I think environments mae sense, so is the var name wrong?

tests/test_authority.py
```diff
@@ -123,3 +128,81 @@ class MockResponse(object):
...
+@patch.multiple(
+    "msal.authority",
+    instance_discovery=DEFAULT,
+    tenant_discovery=Mock(return_value=_mocked_tenant_discovery_response),
+    )
+class TestMsalAuthorityBehaviorsWithoutKnownToDeveloperAuthorities(unittest.TestCase):
+
+    def test_a_known_to_microsoft_authority_should_skip_instance_discovery(self, **mocks):
+        msal.PublicClientApplication("id", authority="https://login.microsoftonline.com/common")
+        mocks["instance_discovery"].assert_not_called()
```
bgavrilMS, line 146: Is it "known to microsoft" because you hardcode it somewhere? 

Not sure I agree with this test, because if a new alias is added, MSAL will get into SSO islands problem.

tests/test_authority.py
```diff
@@ -123,3 +128,81 @@ class MockResponse(object):
...
+
+    def test_validate_authority_boolean_should_skip_instance_discovery(self, **mocks):
+        """Pending deprecation, but kept for backward compatibility, for now"""
+        msal.PublicClientApplication(
+            "id", authority="https://contoso.com/common", validate_authority=False)
+        mocks["instance_discovery"].assert_not_called()
```
bgavrilMS, line 152: As above, instance discovery to figure out aliases -  for SSO islands problem.

tests/test_authority.py
```diff
@@ -123,3 +128,81 @@ class MockResponse(object):
...
+    def test_known_to_developer_authority_should_skip_instance_discovery(self, **mocks):
+        msal.PublicClientApplication.set_known_authorities(["private.cloud"])
+        msal.PublicClientApplication("foo", authority="https://private.cloud/foo")
+        mocks["instance_discovery"].assert_not_called()
+
+    def test_a_known_to_microsoft_authority_should_skip_instance_discovery(self, **mocks):
+        msal.PublicClientApplication.set_known_authorities(["private.cloud"])
+        msal.PublicClientApplication(
+            "id", authority="https://login.microsoftonline.com/common")
+        mocks["instance_discovery"].assert_not_called()
```
bgavrilMS, line 193: as above, we need solution for SSO islands.

msal/application.py
rayluo: Excellent point! I was also torn between different names.

The original Known Authorities specs chose the name, and its pseudo code sometimes contains a full url, sometimes a host name only. And that ambiguity was also observed in MSAL Js, where it was [documented as URI](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/configuration.md#auth-config-options), but the [inline comment suggested otherwise](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/v1.3.4/lib/msal-node/src/config/Configuration.ts#L17). CC @sameerag 

Now I inclined to rename this to `known_hosts`. Thoughts?

tests/test_authority.py
rayluo: > Is it "known to microsoft" because you hardcode it somewhere?

Yes. Historically, many ADAL/MSAL hardcodes a small set of hosts that are known to Microsoft. It was also known as "well known hosts" back then.

> Not sure I agree with this test, because if a new alias is added, MSAL will get into SSO islands problem.

Good point. In this test case, only the known-to-Microsoft check would be skipped, but the authority aliases i.e. instance metadata is still performed. Let me figure out how to rewrite this test case to show that. Stay tuned.

tests/test_authority.py
rayluo: This `validate_authority=False` case is different than the case you commented above. I thought we discussed this in the design doc, that for `validate_authority=False` authority, the authority aliases do not apply, so they can be skipped. Please ping me in Teams so that we can discuss.

msal/application.py
bgavrilMS: known_authority_hosts :)

tests/test_authority.py
rayluo: OK, the test cases are updated to test two behaviors:
* "validation", also written as "authority_validation", but they really just mean "known-to-Microsoft validation" i.e. the Instance Discovery endpoint v1.0 behavior.
* The instance metadata acquisition

tests/test_authority.py
rayluo: Test cases are updated to better show the intention. I still think the instance metadata behavior should be SKIPPED in this case. If you still think otherwise, let's discuss offline.

msal/application.py
```diff
@@ -160,6 +159,39 @@ class ClientApplication(object):
...
+    _known_authorities = None
+
+    @classmethod
+    def set_known_authorities(cls, known_authorities):
+        """Declare a list of domain names which you allow MSAL to operate with.
+
+        If your app operates with some authorities that you know and own,
+        such as some ADFS or B2C or private cloud,
+        it is recommended and sometimes required that you declare them here,
+        so that MSAL will use them without any validation.
```
sameerag, line 171: I do not like the framing here. Can we say: "MSAL will use the endpoints without discovery" instead? "Validation" is what we call the "discovery"

rayluo: Good point about "validation" VS "discovery". I would suggest to avoid the term "endpoint", though, because (1) the endpoint concept is not established in this context; (2) actually we still do a "tenant discovery" (i.e. openid config) to discover the authorization endpoint and token endpoint etc..

So, my next commit will change it to "MSAL will use _these authorities_ without _discovery_".

By the way, are you also OK with those acceptance test cases below, as well as the `known_authority_hosts` naming? Please leave your comment or approval in both this PR and the [internal ApiReview PR](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/6030?_a=files&path=%2FAuthorityValidation%2Fauthority_validation_concepts_and_api.md), then.

bgavrilMS (approved):

tests/test_authority.py
```diff
@@ -123,3 +128,101 @@ class MockResponse(object):
...
+    def test_a_known_to_microsoft_authority_should_skip_validation_but_still_use_instance_metadata(
+            self, instance_metadata, known_to_microsoft_validation, _):
+        msal.PublicClientApplication.set_known_authorities(["private.cloud"])
+        app = msal.PublicClientApplication(
+            "id", authority="https://login.microsoftonline.com/common")
+        known_to_microsoft_validation.assert_not_called()
+        app.get_accounts()  # This could make an instance metadata call for authority aliases
+        instance_metadata.assert_called_once_with()
+
+    def test_unknown_adfs_should_trigger_validation_and_instance_metadata(
```
sameerag, line 209: And will fail with a meaningful error??

sameerag (approved):
Added a few comments but looks good.

msal/application.py
```diff
@@ -453,18 +485,23 @@ def __init__(
 
         # Here the self.authority will not be the same type as authority in input
         try:
+            authority_to_use = authority or "https://login.microsoftonline.com/common/"
```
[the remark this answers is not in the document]
rayluo: Yes, MSAL Python already added those instances into public interface (`https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/221`).

Here, this line is an internal string predating that cloud instance implementation. But yeah, the next commit will use a predefined instance name to replace this inline string.

tests/test_authority.py
rayluo: Not necessarily.

The test case here mainly focuses on demonstrating a "known-to-Microsoft validation" should be performed. The actual outcome would depend on the real validation response from the Instance Discovery endpoint:
* If the Instance Discovery endpoint decides that the input authority is valid, MSAL will proceed, and later also perform Instance Metadata acquisition, which is also demonstrated in this test case.
* If the Instance Discovery endpoint deems the input invalid, MSAL would yield a meaningful error, which is a preexisting behavior, so we do not need to test it here.

msal/application.py
rayluo: Renamed to `known_authority_hosts` in the latest commit.

msal/application.py
```diff
@@ -160,6 +159,45 @@ class ClientApplication(object):
...
+        :param list[str] known_authority_hosts:
+            Authorities that you known, for example::
+
+                [
+                    "contoso.com",  # Your own domain
+                    "login.azs",  # This can be a private cloud
+                ]
+
+        New in version 1.19
+        """
```
xiangyan99, line 189: hosts = frozenset(known_authority_hosts)

if cls._known_authority_hosts == hosts:
    pass

msal/application.py
```diff
@@ -160,6 +159,45 @@ class ClientApplication(object):
...
+        if cls._known_authority_hosts is not None:
```
xiangyan99, line 190: Let's not do the extra check.

xiangyan99: Optional: add a log that known hosts are set.

tests/test_authority.py
rayluo: And now I added a comment inside the test case to call this out. Closing this conversation.

jiasli (approved):

jiasli:
Azure CLI never used `validate_authority` and everything works fine. Per our observation, all Azure Stack environments utilizing ADFS have the `/adfs` postfix in their authentication endpoint URL, which makes MSAL bypass authority verification:

https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/c9a36e952e16f42c6d13315dc3226a0a65c1f59f/msal/authority.py#L78

xiangyan99 (approved):

tests/test_authority.py
```diff
@@ -123,3 +128,104 @@ class MockResponse(object):
...
+        known_to_microsoft_validation.assert_not_called()
+        app.get_accounts()  # This could make an instance metadata call for authority aliases
+        instance_metadata.assert_not_called()
+
+@patch("msal.authority.tenant_discovery", return_value=_mocked_tenant_discovery_response)
+@patch("msal.authority.instance_discovery")
+@patch.object(msal.PublicClientApplication, "_get_instance_metadata", return_value=[])
+class TestMsalAuthorityBehaviorsWithKnownToDeveloperAuthorities(unittest.TestCase):
+
+    def test_known_authorities_should_allow_multiple_configuration_but_only_one_modification(self, *mocks):
```
siddhijain, line 181: What does 'multiple configuation' mean here? Also, why are we allowing 1 modification?

tests/test_authority.py
```diff
@@ -123,3 +128,104 @@ class MockResponse(object):
...
+
+    def test_known_to_developer_authority_should_skip_validation_and_instance_metadata(
+            self, instance_metadata, known_to_microsoft_validation, _):
+        msal.PublicClientApplication.set_known_authority_hosts(["private.cloud"])
+        app = msal.PublicClientApplication("foo", authority="https://private.cloud/foo")
+        known_to_microsoft_validation.assert_not_called()
+        app.get_accounts()  # This could make an instance metadata call for authority aliases
+        instance_metadata.assert_not_called()
+
+    def test_a_known_to_microsoft_authority_should_skip_validation_but_still_use_instance_metadata(
```
siddhijain, line 201: nit 2 test cases seem to have the same name. It might be working in Python but does not work in Java. For the sake of consistency, can this be renamed to something else?

rayluo: Oh, they belong to 2 different classes. Each class is conceptually a group of test scenarios. Group 1 starting from line 140 is testing the absence of new `known_authority_hosts`, the group 2 starting from line 179 is for with the new setting. This group separation makes same name method work.

I could have also mention that "with or without known hosts" into the test case name, but as you can see, each test case's name is already around full-line long. :-)

tests/test_authority.py
rayluo: Good catch! Sorry for ambiguity. Let me change it in next commit. I mean to say "multiple idempotent set operation but zero modification".

tests/test_authority.py
```diff
@@ -123,3 +128,102 @@ class MockResponse(object):
...
+        known_to_microsoft_validation.assert_not_called()
+        app.get_accounts()  # This could make an instance metadata call for authority aliases
+        instance_metadata.assert_not_called()
+
+    def test_by_default_adfs_should_skip_validation_and_instance_metadata(
+            self, instance_metadata, known_to_microsoft_validation, _):
+        """Not strictly required, but when/if we already supported it, we better keep it"""
+        app = msal.ClientApplication("id", authority="https://contoso.com/adfs")
+        known_to_microsoft_validation.assert_not_called()
+        app.get_accounts()  # This could make an instance metadata call for authority aliases
```
keystroke, line 162: nit: update comment to say this should NOT make an instance metadata call for authority aliases

keystroke: oh I think I understand point of comment, we are calling this method because it could make such a call under the hood then we verify it did or did not right after.

tests/test_authority.py
```diff
@@ -123,3 +128,102 @@ class MockResponse(object):
...
+        msal.ClientApplication.set_known_authority_hosts(["private.cloud"])
+        with self.assertRaises(ValueError):
+            msal.ClientApplication(
+                "id",
+                authority="https://unknown.com/common", validate_authority=False)
+
+    def test_known_to_developer_authority_should_skip_validation_and_instance_metadata(
+            self, instance_metadata, known_to_microsoft_validation, _):
+        msal.ClientApplication.set_known_authority_hosts(["private.cloud"])
+        app = msal.ClientApplication("foo", authority="https://private.cloud/foo")
```
keystroke, line 194: Is this the expected code pattern for most / all cloud apps and tools? As it would be quite regressive for a "developer" to be building a tool which only works in public clouds / doesn't work in air-gapped or private-cloud environments. Take Azure CLI for example. The devs of that tool would have the user pass-in the authority endpoint for the cloud they are connecting to. And they can't know whether any of these features are supported by the target cloud, so they'd just take the user-provided input and set it as a known authority? Is that right? Or is the idea they'd leave the known authorities collection empty and expose separate flags for enabling authority validation, instance discovery, and alias lookups? But how would the "user" know whether to enable them or disable them? Or, would Azure CLI devs just always disable them entirely?

rayluo: > Is this the expected code pattern for most / all cloud apps and tools?

In short, yes, but that new optional `set_known_authority_hosts()` call is only needed when a downstream app is deployed to work with a small set of known (-to-developer) authorities. To give you a concrete example, the first four test cases in this PR do not need that `set_known_authority_hosts()` call.

> As it would be quite regressive for a "developer" to be building a tool which only works in public clouds / doesn't work in air-gapped or private-cloud environments. Take Azure CLI for example. The devs of that tool would have the user pass-in the authority endpoint for the cloud they are connecting to. And they can't know whether any of these features are supported by the target cloud, so they'd just take the user-provided input and set it as a known authority? Is that right?
> Or is the idea they'd leave the known authorities collection empty and expose separate flags for enabling authority validation, instance discovery, and alias lookups? But how would the "user" know whether to enable them or disable them? Or, would Azure CLI devs just always disable them entirely?

First of all, there is no longer such a thing as "(whether) they can't know whether *any of these features (enabling authority validation, instance discovery, and alias lookups)* are supported by the target cloud". Based on the simplified proposal, here this implementation PR exposes only one conceptual setting "what are the authorities which are unknown to Microsoft, but *you* still allow/want this app to operate with". It is a pure conceptual declaration, "I'm deploying this app to my contoso.com so I tell this app to proceed with authority contoso.com", and then MSAL will figure out the rest (which is basically skip all under-the-hood checks).

And yes, a MSAL-powered tool needs to expose such a new setting, when and only when the tool developer envisions their tool to be potentially deployed into a private cloud. The actual logistics is decided by each tool. Take Azure CLI for example, their ["az cloud" commands](https://docs.microsoft.com/en-us/cli/azure/cloud?view=azure-cli-latest) could be considered as a strong signal that the end user is aware of, and explicitly opts in to use an unusual cloud. So, that cloud could be fed into MSAL's new `set_known_authority_cloud()`. On the other hand, an app somehow encountered a "phishing.fabricam.com" authority during runtime, should not blindly set it as a "known authority" and subsequently bring end user to that compromised website and input their username and password there.

> Or, would any devs just always disable the 3 sub-features entirely? By always setting any "unreliable.fabricam.com" as a known authority?

I rephrased your question to make it more clear. The answer is:

* They don't need to. See the paragraph above.
* But if they insist, we cannot prevent them from doing so. Just keep in mind that, this defeats the purpose of authority validation, which is a security feature.
* But at least our new declarative `set_known_authority_hosts(...)` makes a developer less likely to "always disable authority validation", because they won't be able to enumerate all "phishing.fabricam.com", "hacked.contoso.com" etc. beforehand. (FWIW, in older version of MSALs, there used to be a boolean flag to simply turn off authority validation regardless of the incoming host name. That boolean flag could potentially be misused, unknowingly (pun not intended).)

rayluo: UPDATE: We worked out #496 to give a decisive boolean flag to turn off Instance Discovery endpoint I/O.

tests/test_authority.py
```diff
@@ -123,3 +128,95 @@ class MockResponse(object):
...
+@patch("msal.authority.instance_discovery")
+@patch.object(msal.ClientApplication, "_get_instance_metadata", return_value=[])
+class TestMsalBehaviorsWithoutAndWithKnownAuthorityHosts(unittest.TestCase):
+    """Test cases use ClientApplication, which is a base class of both PCA and CCA"""
+
+    def test_by_default_a_known_to_microsoft_authority_should_skip_validation_but_still_use_instance_metadata(
+            self, instance_metadata, known_to_microsoft_validation, _):
+        app = msal.ClientApplication("id", authority="https://login.microsoftonline.com/common")
+        known_to_microsoft_validation.assert_not_called()
+        app.get_accounts()  # This could make an instance metadata call for authority aliases
```
chlowell, line 145: Is it expected in this case that `get_accounts()` does instance discovery? That is to say, `ClientApplication()` doesn't call `known_to_microsoft_validation` but `get_accounts()` does.

rayluo: The answer is two-fold.

* The "validating whether an input authority looks legit" and "getting some metadata for alias grouping" are two different things. This test case, as described in line 141, made a point that one does not necessarily turn off the other. In this case, `ClientApplication()` skips validation, but `get_accounts()` would like to know alias groups.
* The actual timing of when to send out the metadata call, is an implementation detail. Since MSAL Python delays that metadata call to `get_accounts()`, so this test case is written this way. Your MSAL's implementation does not have to call `get_accounts()`.

Last but not least, this entire PR has been shelved, because later we discovered some [new requirements for Azure Stack scenarios (internal link)](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=/AuthorityValidation/authority_validation_concepts_and_api.md&_a=preview&anchor=skip-authority-validation-and-alias-metadata-behaviors-without-a-set-of-predefined-authorities), so we proceeded with #496.

rayluo:
Closing this without merging, because we ended up going with #496

PR #496, opened 2022-09-03
Implement instance_discovery=False

rayluo:
Driven by [this conversation](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/492#pullrequestreview-1082931221), this PR implements `instance_discovery=False` based on [internal design 6504](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/6504). (Note: This PR does NOT contain `known_authority_hosts`, which was implemented in a different PR #492. The two features would co-exist when both PRs are merged in, but they can also work independently. Here they are implemented independently.)

As usual, the unit tests in this PR also serve as acceptance test for the aforementioned design.

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -13,7 +13,7 @@ import os
 
 from .oauth2cli import Client, JwtAssertionCreator
 from .oauth2cli.oidc import decode_part
-from .authority import Authority
+from .authority import Authority, WORLD_WIDE
 from .mex import send_request as mex_send_request
 from .wstrust_request import send_request as wst_send_request
 from .wstrust_response import *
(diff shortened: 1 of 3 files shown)


msal/application.py
```diff
@@ -174,6 +173,7 @@ def __init__(
                 # when we would eventually want to add this feature to PCA in future.
             exclude_scopes=None,
             http_cache=None,
+            instance_discovery=None,
```
xiangyan99, line 176: Is instance_discovery=None here means default to False?

I guess we want the default value to be true?

rayluo: Basically, I think I am moving away from using `True` or `False` as the default value, because they would force downstream yet mid-tier libraries (such as Azure Identity) to also duplicate the same default value in their corresponding API surface.

I would instead use `None` as default value, and it just means *default*. It does NOT imply a boolean evaluation to make it a False.

And the actual default behavior will and only will be available via documentation. See line 419 in this PR.

xiangyan99: Do you mean the caller (me :)) can specify None as the input value of instance_discovery to say "use default"?

rayluo: Yes :-)

PR #511, opened 2022-11-02
AT POP for Public Client based on broker

rayluo:
This is the implementation of the [internal work item](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1791492), and [this internal design doc](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=/PoPTokensProtocol/PoP_API_In_MSAL.md&_a=preview).
Currently, this PR provides a low level API to obtain a Proof-of-Possesson (PoP) access token.

We may add more end-to-end test case in the future. And we may provide a high-level API.

The beta preview of this feature is available by `pip install --ignore-installed git+https://github.com/AzureAD/microsoft-authentication-library-for-python.git@pop`

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -182,6 +182,10 @@ class ClientApplication(object):
     _TOKEN_SOURCE_BROKER = "broker"
 
     _enable_broker = False
+    _AUTH_SCHEME_UNSUPPORTED = (
+        "auth_scheme is currently only available from broker. "
+        "You can enable broker by following these instructions. "
+        "https://msal-python.readthedocs.io/en/latest/#publicclientapplication")
 
     def __init__(
             self, client_id,
@@ -556,6 +560,10 @@ class ClientApplication(object):
                     "Broker is unavailable on this platform. "
                     "We will fallback to non-broker.")
         logger.debug("Broker enabled? %s", self._enable_broker)
+
+    def is_pop_supported(self):
+        """Returns True if this client supports Proof-of-Possession Access Token."""
+        return self._enable_broker
 
     def _decorate_scope(
             self, scopes,
@@ -1185,6 +1193,7 @@ class ClientApplication(object):
             authority=None,  # See get_authorization_request_url()
             force_refresh=False,  # type: Optional[boolean]
             claims_challenge=None,
+            auth_scheme=None,
             **kwargs):
         """Acquire an access token for given account, without user interaction.
 
@@ -1205,7 +1214,7 @@ class ClientApplication(object):
             return None  # A backward-compatible NO-OP to drop the account=None usage
         result = _clean_up(self._acquire_token_silent_with_error(
             scopes, account, authority=authority, force_refresh=force_refresh,
-            claims_challenge=claims_challenge, **kwargs))
+            claims_challenge=claims_challenge, auth_scheme=auth_scheme, **kwargs))
         return result if result and "error" not in result else None
 
     def acquire_token_silent_with_error(
@@ -1215,6 +1224,7 @@ class ClientApplication(object):
             authority=None,  # See get_authorization_request_url()
             force_refresh=False,  # type: Optional[boolean]
             claims_challenge=None,
+            auth_scheme=None,
             **kwargs):
         """Acquire an access token for given account, without user interaction.
 
@@ -1241,6 +1251,12 @@ class ClientApplication(object):
             in the form of a claims_challenge directive in the www-authenticate header to be
             returned from the UserInfo Endpoint and/or in the ID Token and/or Access Token.
             It is a string of a JSON object which contains lists of claims being requested from these locations.
+        :param object auth_scheme:
+            You can provide an ``msal.auth_scheme.PopAuthScheme`` object
+            so that MSAL will get a Proof-of-Possession (POP) token for you.
+
+            New in version 1.26.0.
+
         :return:
             - A dict containing no "error" key,
               and typically contains an "access_token" key,
@@ -1252,7 +1268,7 @@ class ClientApplication(object):
             return None  # A backward-compatible NO-OP to drop the account=None usage
         return _clean_up(self._acquire_token_silent_with_error(
             scopes, account, authority=authority, force_refresh=force_refresh,
-            claims_challenge=claims_challenge, **kwargs))
+            claims_challenge=claims_challenge, auth_scheme=auth_scheme, **kwargs))
 
     def _acquire_token_silent_with_error(
             self,
@@ -1261,6 +1277,7 @@ class ClientApplication(object):
             authority=None,  # See get_authorization_request_url()
             force_refresh=False,  # type: Optional[boolean]
             claims_challenge=None,
+            auth_scheme=None,
             **kwargs):
         assert isinstance(scopes, list), "Invalid parameter type"
         self._validate_ssh_cert_input_data(kwargs.get("data", {}))
@@ -1276,6 +1293,7 @@ class ClientApplication(object):
             scopes, account, self.authority, force_refresh=force_refresh,
             claims_challenge=claims_challenge,
             correlation_id=correlation_id,
+            auth_scheme=auth_scheme,
             **kwargs)
         if result and "error" not in result:
             return result
@@ -1298,6 +1316,7 @@ class ClientApplication(object):
                 scopes, account, the_authority, force_refresh=force_refresh,
                 claims_challenge=claims_challenge,
                 correlation_id=correlation_id,
+                auth_scheme=auth_scheme,
                 **kwargs)
             if result:
                 if "error" not in result:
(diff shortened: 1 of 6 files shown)


msal/application.py
```diff
@@ -1385,10 +1404,13 @@ def _acquire_token_silent_from_cache_and_possibly_refresh_it(
                     claims=_merge_claims_challenge_and_capabilities(
                         self._client_capabilities, claims_challenge),
                     correlation_id=correlation_id,
+                    auth_scheme=auth_scheme,
                     **data)
                 if response:  # The broker provided a decisive outcome, so we use it
                     return self._process_broker_response(response, scopes, data)
 
+            if auth_scheme:
+                raise ValueError("auth_scheme is currently only available from broker")
```
bgavrilMS, line 1413: Consider adding more details like how to enable the broker redirect uri, or an aka.ms link

msal/application.py
```diff
@@ -1333,9 +1349,10 @@ def _acquire_token_silent_from_cache_and_possibly_refresh_it(
             force_refresh=False,  # type: Optional[boolean]
             claims_challenge=None,
             correlation_id=None,
+            auth_scheme=None,
             **kwargs):
         access_token_from_cache = None
-        if not (force_refresh or claims_challenge):  # Bypass AT when desired or using claims
+        if not (force_refresh or claims_challenge or auth_scheme):  # Then attempt AT cache
```
bgavrilMS, line 1355: I am not sure if this is the case, but if POP auth scheme is specified, we should also not attempt to use RT in the cache. This will break migration scenario: 

1. App uses webview and points to a cache file. App gets bearer tokens for a user: AT, RT, IdT
2. App switches to WAM, while still pointing at the same cache file, so as to not lose state, i.e. GetAccounts() correctly returns 1 account.
3. App calls AcquireTokenSilent(user) + POP scheme

MSAL now: 

- bypasses the AT token cache 
- looks in the cache for an RT
- finds the RT and ?? I'm not sure what happens here, but we should not use the RT or fail. We should just go to WAM.

Maybe this is a test you want to write? 

PS: we have this behavior now in MSAL.NET and it is one of the GA blockers.

tests/test_e2e.py
```diff
@@ -1036,5 +1040,54 @@ def test_acquire_token_silent_with_an_empty_cache_should_return_none(self):
...
+        auth_scheme = PopAuthScheme(
+            http_method="GET",
+            url="https://www.Contoso.com/Path1/Path2?queryParam1=a&queryParam2=b",
+            nonce="placeholder",
+            )
+        config = self.get_lab_user(usertype="cloud")
+        config["password"] = self.get_lab_user_secret(config["lab_name"])
+        result = self._test_username_password(auth_scheme=auth_scheme, **config)
+        self.assertEqual(result["token_type"], "pop")
+        payload = json.loads(decode_part(result["access_token"].split(".")[1]))
```
bgavrilMS, line 1056: Consider asserting that an RT is not in the cache. WAM should not give back an RT.

tests/test_e2e.py
```diff
@@ -1036,5 +1040,54 @@ def test_acquire_token_silent_with_an_empty_cache_should_return_none(self):
...
+class PopTestCase(LabBasedTestCase):
+    def test_at_pop_should_contain_pop_scheme_content(self):
+        auth_scheme = PopAuthScheme(
+            http_method="GET",
+            url="https://www.Contoso.com/Path1/Path2?queryParam1=a&queryParam2=b",
+            nonce="placeholder",
+            )
+        config = self.get_lab_user(usertype="cloud")
+        config["password"] = self.get_lab_user_secret(config["lab_name"])
+        result = self._test_username_password(auth_scheme=auth_scheme, **config)
```
bgavrilMS, line 1054: Consider calling `AcquireTokenSilent` + same `PopAuthScheme` after this and assert that the token does not come from the cache.

bgavrilMS (approved):

msal/application.py
```diff
@@ -1224,6 +1225,12 @@ def acquire_token_silent(
             returned from the UserInfo Endpoint and/or in the ID Token and/or Access Token.
             It is a string of a JSON object which contains lists of claims being requested from these locations.
 
+        :param object auth_scheme:
```
bgavrilMS, line 1228: Is there public API similar to `IsPOPSupportedByClient` as per spec?

bgavrilMS:
@rayluo - when are you closing this?

rayluo:
> @rayluo - when are you closing this?

After we wrap up the July Focus tasks, we can probably squeeze this in before next big task.

Also, I heard that nowadays the MS Graph supports POP token. Is that true? If so, we might as well figure out a way to test it end-to-end before the shipment.

bgavrilMS:
> > @rayluo - when are you closing this?
> 
> After we wrap up the July Focus tasks, we can probably squeeze this in before next big task.
> 
> Also, I heard that nowadays the MS Graph supports POP token. Is that true? If so, we might as well figure out a way to test it end-to-end before the shipment.

I am not sure about Graph, but SPO and EXO do.

bgavrilMS:
Gentle reminder.

msal/application.py
rayluo: 1. Currently, no. And we can easily add a new `is_broker_enabled()` public API, whose name was derived from [the specs](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=/PoPTokensProtocol/PoP_API_In_MSAL.md&_a=preview&anchor=proposal-2---optional-isproofofposessionsupportedbyclient-helper-(accepted)) which mentioned:

   > Note: internally, MSAL java, py, .NET and js would implement this as `public bool IsBrokerAvailable()`.

2. There is a follow-up question. Does such a new `is_broker_enabled()` public API have any implication to the debatable ["naming of `allow_broker` VS `enable_broker`" topic](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/596#discussion_r1357920721)? Now this new `is_broker_enabled()` helper indeed makes it obvious that "even after opting in `enable_broker_on_windows=True`, there is still no guarantee that the broker is actually enabled". Then we shall probably name the opt-in flag as `allow_broker_on_windows=True`.

msal/application.py
rayluo: > if POP auth scheme is specified, we should also not attempt to use RT in the cache. This will break migration scenario:
> ...
> Maybe this is a test you want to write?
>
> PS: we have this behavior now in MSAL.NET and it is one of the GA blockers.

Nice call-out. MSAL Python implementation does not have such an issue, because the subsequent code path will use broker and will not fall back.

I will see how to add a test case here.

P.S.: The other day, our downstream partner asked whether MSAL .Net is already GA now. Is it?

tests/test_e2e.py
rayluo: Hmm, our latest PR of "token source" happens to allow us easily test whether a token came from cache. Would you mind also review and approve that PR? (`https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/610`)

Afterwards, I'll rebase this old POP PR to the latest `dev`, and then proceed to merge and ship it.

msal/application.py
bgavrilMS: `is_pop_supported_by_client` seems like a better feature flag, because it hides the implementation detail of `is_broker_available`.

rayluo: Based on offline discussion, we agreed on a short `PCA.is_pop_supported()`. It accepts no `scope` parameter thus implies that it is about client's POP readiness.

tests/test_e2e.py
rayluo: I guess what you really want to assert is that the `token_source` should be `broker`. I'm adding that assertion in my next commit.
We do not need to assert "WAM should not give back an RT". That is a WAM behavior that MSAL does not control, and shall not make an assertion on.

msal/application.py
rayluo: UPDATE: [Added a new check](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/511/files#diff-ed3b8af48cd21dc1ac735b839d89b172a6f45569c324af70a2bdb4f60e5fbcddR130-R132) in the existing test helpers. Closing this conversation thread.

tests/test_e2e.py
```diff
@@ -1147,5 +1163,117 @@ def test_acquire_token_silent_with_an_empty_cache_should_return_none(self):
...
+        self.assertEqual(result["token_source"], "broker", "POP is only supported by broker")
+        self.assertEqual(result["token_type"], "pop")
+        payload = json.loads(decode_part(result["access_token"].split(".")[1]))
+        logger.debug("AT POP payload = %s", json.dumps(payload, indent=2))
+        self.assertEqual(payload["m"], auth_scheme._http_method)
+        self.assertEqual(payload["u"], auth_scheme._url.netloc)
+        self.assertEqual(payload["p"], auth_scheme._url.path)
+        self.assertEqual(payload["nonce"], auth_scheme._nonce)
+
+    # TODO: Remove this, as ROPC support is removed by Broker-on-Win
```
bgavrilMS, line 1191: Is this still the case?

rayluo: It is still the case, until/unless that decision is overturned. I have pinged the forum again in its Teams channel. You are also CCed.

Currently, we keep this test case here in case that pending conversation will go another direction. But we actually skip this test case (see [line 1194](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/511/files#diff-ed3b8af48cd21dc1ac735b839d89b172a6f45569c324af70a2bdb4f60e5fbcddR1194)).

By default, I'll still proceed to merge this PR in, by EOD today.

msal/auth_scheme.py
```diff
@@ -0,0 +1,34 @@
+try:
+    from urllib.parse import urlparse
```
xiangyan99, line 2: Now we only support py37+, do we still need it?

xiangyan99:
Done with smoke tests and did not find regressions.

PR #539, opened 2023-03-02
Update setup.cfg with documentation URL

pamelafox:
It'd be nice if pypi.org showed a link to the documentation. I think this is how it's specified, but all the docs these days show pyproject.toml, not setup.cfg, so I'm not positive.

See also https://github.com/rayluo/identity/pull/1

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -5,3 +5,5 @@ universal=1
 project_urls =
     Changelog = https://github.com/AzureAD/microsoft-authentication-library-for-python/releases
     Documentation = https://msal-python.readthedocs.io/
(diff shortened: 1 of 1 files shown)


pamelafox:
Sure, we can try extra links, didn't realize you can put anything.

rayluo:
> Sure, we can try extra links, didn't realize you can put anything.

I haven't tried that. But it seems feasible based on [this doc](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls).

I'll merge this in. And we will see the result in our next release.

rayluo (approved):

PR #541, opened 2023-03-03
Pass additional headers to acquire_token_interactive

andresperezl:
Allow passing additional headers to the `acquire_token_interactive` method, so they don't clash with the ones with the generated from `telemetry_context.generate_headers()`

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -1831,7 +1831,7 @@ class PublicClientApplication(ClientApplication):  # browser app or mobile app
                 "domain_hint": domain_hint,
                 },
             data=dict(data, claims=claims),
-            headers=telemetry_context.generate_headers(),
+            headers=telemetry_context.generate_headers() | kwargs.pop('headers', {}),
             browser_name=_preferred_browser(),
             **kwargs))
         telemetry_context.update_telemetry(response)


rayluo:
Would you please also let us know what header are you trying to send, in what scenario is that useful?

andresperezl:
Is basically this same problem https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/2482 where when running the script locally you get the error 

```
AADSTS9002327: Tokens issued for the 'Single-Page Application' client-type may only be redeemed via cross-origin requests.
```

And the solution is to pass the missing header `Origin` , but trying to set `header` in the `kwargs` conflicts with the telemetry headers

andresperezl:
Sometime you don't have access to the app configuration, which can happen even with different teams within the same company, if it can be done with the Javascript implementation of the library, why not with the python one?

bgavrilMS:
Not a PY scenario, can be done by HTTP Client customization probably if really needed.

rayluo:
> Sometimes you don't have access to the app configuration, which can happen even with different teams within the same company, if it can be done with the Javascript implementation of the library, why not with the python one?

Our understanding is that a SPA app running inside browser would have browser automatically set that `Origin` header for you, and Microsoft Entra ID backend apparently verifies that `Origin` header. Now, if you are using Python to impersonate that SPA app and running outside of browser, mimicking an `Origin` header may bypass that Entra backend check, however we do not know its security implication. For that reason, we hesitate to take this PR. Closing it for now; we may revisit it in the future when/if the need arises again.

PR #543, opened 2023-03-07
Supporting https://contoso.ciamlogin.com as authority

rayluo:
The automation tests ~currently fail because~ passes:

* ~the new domain name is unavailable~
* ~the OIDC discovery endpoint is not available. When it is available, it is expected to also be in the ".../v2.0/..." endpoint.~
* ~we will continue the prototyping after the above two become available~
* ~Waiting on new lab api~
* New [lab api](https://msidlab.com/api/app/b8e9d222-c4ee-414c-ac29-b0eff1f32400) has been updated to return a new authority

Testers shall take a look into logs to double check your auth/token requests were sent out to `https://msidlabciam1.ciamlogin.com/d57fb3d4-4b5a-4144-9328-9c1f7d58179d/oauth2/v2.0/token?dc=ESTS-PUB-EUS-AZ1-FD000-TEST1` which was discovered from [OIDC discovery](https://msidlabciam1.ciamlogin.com/msidlabciam1.onmicrosoft.com/v2.0/.well-known/openid-configuration?dc=ESTS-PUB-EUS-AZ1-FD000-TEST1). If an MSAL you are working on does not obtain those endpoints from OIDC discovery, you could probably use `https://msidlabciam1.ciamlogin.com/TENANT.ONMICROSOFT.COM/oauth2/v2.0/token?dc=ESTS-PUB-EUS-AZ1-FD000-TEST1` instead.

### DIFF
--- a/msal/authority.py
+++ b/msal/authority.py
@@ -5,8 +5,6 @@ except ImportError:  # Fall back to Python 2
     from urlparse import urlparse
 import logging
 
-from .exceptions import MsalServiceError
-
 
 logger = logging.getLogger(__name__)
 
@@ -28,7 +26,9 @@ WELL_KNOWN_B2C_HOSTS = [
     "b2clogin.cn",
     "b2clogin.us",
     "b2clogin.de",
+    "ciamlogin.com",
     ]
+_CIAM_DOMAIN_SUFFIX = ".ciamlogin.com"
 
 
 class AuthorityBuilder(object):
@@ -80,7 +80,8 @@ class Authority(object):
         if isinstance(authority_url, AuthorityBuilder):
             authority_url = str(authority_url)
         authority, self.instance, tenant = canonicalize(authority_url)
-        self.is_adfs = tenant.lower() == 'adfs'
+        is_ciam = self.instance.endswith(_CIAM_DOMAIN_SUFFIX)
+        self.is_adfs = tenant.lower() == 'adfs' and not is_ciam
         parts = authority.path.split('/')
         self._is_b2c = any(
             self.instance.endswith("." + d) for d in WELL_KNOWN_B2C_HOSTS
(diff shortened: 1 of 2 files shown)


tests/test_authority.py
```diff
@@ -78,6 +78,13 @@ def test_invalid_host_skipping_validation_can_be_turned_off(self):
         except:  # Could be requests...RequestException, json...JSONDecodeError, etc.
             pass  # Those are expected for this unittest case
 
+    def test_ciam_authority_should_be_path_less(self):
+        http_client = MinimalHttpClient()
+        with self.assertRaises(ValueError):
+            Authority('https://contoso.ciamlogin.com/undesirable_path', http_client)
+        Authority('https://msidlabciam1.ciamlogin.com', http_client)
```
rayluo, line 85: @jmprieur et al, this test case rejects a ciam authority with path.

tests/test_e2e.py
```diff
@@ -897,6 +897,7 @@ def setUpClass(cls):
         # FYI: Only single- or multi-tenant CIAM app can have other-than-OIDC
         # delegated permissions on Microsoft Graph.
         cls.app_config = cls.get_lab_app_object(cls.user["client_id"])
+        cls.app_config["authority"] = "https://msidlabciam1.ciamlogin.com?dc=ESTS-PUB-EUS-AZ1-FD000-TEST1"  # TODO: Remove this line after lab API returns this value
```
rayluo, line 900: Note-to-self: Remove this line after lab API returns this value

jennyf19 (approved):

tests/test_authority.py
jmprieur: It's actually fine to let it go through (custom domains)

msal/authority.py
```diff
@@ -109,13 +110,16 @@ def __init__(
...
-            tenant_discovery_endpoint = (
-                'https://{}:{}{}{}/.well-known/openid-configuration'.format(
-                    self.instance,
-                    443 if authority.port is None else authority.port,
-                    authority.path,  # In B2C scenario, it is "/tenant/policy"
-                    "" if tenant == "adfs" else "/v2.0" # the AAD v2 endpoint
-                    ))
+            if is_ciam and len(authority.path) > 1:
+                raise ValueError(
+                    "CIAM authority should look like https://contoso.ciamlogin.com")
```
jmprieur, line 115: it's also ok to have
- contoso.ciamlogin.com/contoso.onmicrosoft.com
- contoso.ciamlogin.com/GUID of contoso.onmicrosoft.com
- contoso.ciamlogin.com/contoso.com (if contoso.com is a valid domain of the contoso.onmicrosoft.com tenant)

jmprieur (commented):
Added comments

msal/authority.py
rayluo: OK, I have switched the implementation and added [these 2 new test cases to cover path-less and with-path scenarios](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/543/files#diff-1982837bc801e709175e25839e1046a445df689d7a2e6309ccb6856e350dee7eR89-R99). Please review again, @jmprieur

tests/test_authority.py
rayluo: Thanks. I am convinced that MSAL should not accept path-less only. This behavior and test case have been removed in latest push.

jmprieur (approved):
LGTM
Thanks @rayluo

PR #559, opened 2023-05-02
Clarify that allow_broker is not applicable to ConfidentialClientApplication

### LINKED ISSUE #544: allow_broker=True for ConfidentialClientApplication returns error
**Describe the bug**
Trying to pass allow_broker=True for as an argument for ConfidentialClientApplication returns an error that it is not supported. However, the [documentation](https://msal-python.readthedocs.io/en/latest/index.html#msal.ConfidentialClientApplication.params.allow_broker) says that it is supported.

**Expected behavior**
For the package to try to create a broker.

**What you see instead**
  File "c:\Users\matyvic\Documents\CSPC\colectica-rest-api-demo\python\corestica\corestica.py", line 561, in <module>
    staging_test = ColecticaRepository(
  File "c:\Users\matyvic\Documents\CSPC\colectica-rest-api-demo\python\corestica\corestica.py", line 131, in __init__
    self.authenticate_via_sign_in(username, password)
  File "c:\Users\matyvic\Documents\CSPC\colectica-rest-api-demo\python\corestica\corestica.py", line 171, in authenticate_via_sign_in
    app = msal.ConfidentialClientApplication(
  File "C:\Users\matyvic\AppData\Roaming\Python\Python39\site-packages\msal\application.py", line 555, in __init__
    raise ValueError("allow_broker=True is only supported in PublicClientApplication")
ValueError: allow_broker=True is only supported in PublicClientApplication

1.21.0

rayluo:
Although MSAL provides two top-level classes: PublicClientApplication and ConfidentialClientApplication, there is another less-popular but still-doable pattern which is to use the base class ClientApplication directly. So, that `allow_broker` is applicable to both PublicClientApplication and the base class ClientApplication. So, that dynamic check in base class is necessary.

The ConfidentialClientApplication inherits the `__init__()` and its documentation from base class. Now, we choose to add a sentence in that documentation to clarify that the parameter is not applicable to ConfidentialClientApplication.

See it in action in [the staged documentation](https://msal-python.readthedocs.io/en/docs-staging/#msal.ConfidentialClientApplication.params.allow_broker).

This resolves #544.

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -444,6 +444,8 @@ class ClientApplication(object):
             New in version 1.19.0.
 
         :param boolean allow_broker:
+            This parameter is NOT applicable to :class:`ConfidentialClientApplication`.
+
             A broker is a component installed on your device.
             Broker implicitly gives your device an identity. By using a broker,
             your device becomes a factor that can satisfy MFA (Multi-factor authentication).


msal/application.py
```diff
@@ -444,6 +444,8 @@ def __init__(
             New in version 1.19.0.
 
         :param boolean allow_broker:
+            This parameter is NOT applicable to :class:`ConfidentialClientApplication`.
```
jiasli, line 447: Maybe a positive statement?

> This parameter is ONLY applicable to `PublicClientApplication`.

rayluo: I believe the emphasis of this PR and its corresponding issue were to call out the only invalid situation among the three. So, taking your suggestion into account, we can change it to:

> This parameter is applicable to `PublicClientApplication` and its base class `ClientApplication`, but NOT applicable to `ConfidentialClientApplication`.

How about that?

PR #560, opened 2023-05-09
Proof-of-Concept: MI via CCA

rayluo:
If this PR will be merged in, the subsequent "how to use Managed Identity" docs would look like this.

[Managed Identity](https://learn.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview) is a kind of confidential client that you do not need to manage its credential.

* How to use Managed Identity?

1. In MSAL Python, you can declare a system-assigned managed identity by
```python
mi = msal.SystemAssignedManagedIdentity()
```
or
```python
mi = msal.UserAssignedManagedIdentity(client_id="my_id")
```
or
```python
mi = msal.UserAssignedManagedIdentity(resource_id="my_id")
```

Those managed identity objects are all data objects. They do not give you tokens on their own.

2. Now you need to feed this `mi` object into a `ManagedIdentityClient` object and then use it:
```python
app = msal.ManagedIdentityClientApplication(mi, ...)
app.acquire_token_for_client(...)
```

3. (EXPERIMENTAL) When we said earlier that "managed identity is a kind of confidential client that you do not need to manage its credential", we meant it. MSAL Python's good old `ConfidentialClientApplication` object also accepts the managed identity object `mi` as an input. So you can do:
```python
app = msal.ConfidentialClientApplication(mi)
app.acquire_token_for_client(...)
```

You can choose to do either `1+2` or `1+3`.

* Recipe: Can I still use the familiar dev pattern of write code once, test locally, and then deploy the **same** app to remote server?

Yes but with a caveat. Because the managed identity is only available on a certain Azure environments (such as Azure VMs, Azure App Service, etc.), and not on your local dev machine, you would have to use normal confidential client during your local testing, and switch to real managed identity on remote server.

MSAL Python makes the transition easy for you. In MSAL Python, all Managed Identity objects have their equivalent JSON representation, for example, `UserAssignedManagedIdentity(client_id="my_id")` is equivalent to `{"ManagedIdentityIdType": "ClientId", "Id": "foo"}`. Combining this characteristic and the `1+3` above, you could write your app in this universal way.

```python
import os, json, msal
client_id_or_managed_identity = json.loads(os.getenv("MY_APP")
credential = json.loads(os.getenv("MY_CREDENTIAL")
app = msal.ConfidentialClientApplication(client_id_or_managed_identity, client_credential=credential)
app.acquire_token_for_client(...)
```

Now you can set these two environment variables on your local dev machine:
```
MY_APP=my_normal_client_id
MY_CREDENTIAL=my_apps_client_secret
```
and set them differently on your remote server:
```
MY_APP='{"ManagedIdentityIdType": "ClientId", "Id": "foo"}'
MY_CREDENTIAL=null
```

Your same app will consume those two sets of settings and behave accordingly.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -130,3 +130,24 @@ See `SerializableTokenCache` for example.
 
 .. autoclass:: msal.SerializableTokenCache
    :members:
+
+
+Managed Identity
+================
+MSAL supports
+`Managed Identity <https://learn.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview>`_.
+
+You can create one of these two kinds of managed identity configuration objects:
+
+.. autoclass:: msal.SystemAssignedManagedIdentity
+   :members:
+
+.. autoclass:: msal.UserAssignedManagedIdentity
+   :members:
+
+And then feed the configuration object into a :class:`ManagedIdentityClient` object.
+
+.. autoclass:: msal.ManagedIdentityClient
+   :members:
+
+   .. automethod:: __init__
--- a/msal/__init__.py
+++ b/msal/__init__.py
@@ -33,4 +33,8 @@ from .application import (
     )
 from .oauth2cli.oidc import Prompt
 from .token_cache import TokenCache, SerializableTokenCache
+from .managed_identity import (
+    SystemAssignedManagedIdentity, UserAssignedManagedIdentity,
+    ManagedIdentityClient,
+    )
 
--- a/msal/application.py
+++ b/msal/application.py
@@ -22,6 +22,7 @@ import msal.telemetry
 from .region import _detect_region
 from .throttled_http_client import ThrottledHttpClient
 from .cloudshell import _is_running_in_cloud_shell
+from .imds import ManagedIdentityClient, ManagedIdentity, _scope_to_resource
 
 
 # The __init__.py will import this. Not the other way around.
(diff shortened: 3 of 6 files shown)


jiasli:
Moved this comment to https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/687#issuecomment-2969109557

rayluo:
> There is an internal work item of migrating to **managed identity + multitenant application** (MI+CCA). Below is some sample code:
> 
> * Python: https://gist.github.com/amritpal-singh-98/92c5b7627e7ea0dc1491e03b7b834310
> * .NET: https://gist.github.com/AtOMiCNebula/988383241cd1837a6c3faf6a353cecea
> 
> This PR may need to be resurrected.

The PR here is [not the droid you are looking for](https://static.wikia.nocookie.net/star-wars-memes/images/e/e4/NotTheDroids.png/revision/latest?cb=20200707002606). :-)

To clarify, the PR 560 here was for a usage like this:

```python
mi = msal.SystemAssignedManagedIdentity()
app = msal.ConfidentialClientApplication(mi)  # It does NOT contain client id
app.acquire_token_for_client(...)
```

Note that its usage is "using a managed identity instance **as if it is a client id**", therefore it does NOT even allow specifying a normal `client_id` parameter. So, it was NOT about a confidential client (a.k.a. service principle) federated by a managed identity. It was about providing pure Managed Identity functionality via the existing `ConfidentialClientApplication` API. I was experimenting this approach to avoid adding a dedicated `ManagedIdentityClient` class. We did not choose this approach.

But the good news is we have [another PR that is for the federation scenario](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/687). That PR's usage is "using a managed identity instance **as if it is a client credential**", together with a normal client id, therefore it is the federation pattern. Please subscribe that PR and subsequent conversation can happen there.

jiasli:
Ah, I indeed mean https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/687. My memory is "corrupted".

Issue #563, opened 2023-05-18
Using `allow_broker` with device code flow causes token refreshing to fail: `Account with id '(pii)' not found.`


jiasli:
**Describe the bug**
Using `allow_broker` with device code flow causes token refreshing to fail: `Account with id '(pii)' not found.`

**To Reproduce**
```sh
# Opt-in WAM
az config set core.allow_broker=true

# Turn off token encryption so that we can edit it manually later
az config set core.encrypt_token_cache=false

# Log in with device code, but with an account not in WAM, such as domain account testuser@azuresdkteam.onmicrosoft.com
az login --tenant 54826b22-38d6-4fb2-bad9-b7b93a3e9c5a --use-device-code

# Edit ~/.azure/msal_token_cache.json. Change AccessToken.<key>.expires_on to 0 to make the access token expire
# {
#     "AccessToken": {        
#         "...": {
#             ...
#             "expires_on": "0",

# Trigger token refreshing
az group list
```

**Expected behavior**
Token refreshing should succeed.

**What you see instead**
```
Account with id '(pii)' not found. Status: Response_Status.Status_AccountNotFound, Error code: 0, Tag: 525678464
```

Apparently, this is because device code's refresh token is saved to MSAL token cache, not WAM, but during token refreshing, MSAL tries to get access token from WAM.

**The MSAL Python version you are using**
Latest `dev` branch: 56256e4c516e0d8469b554223f4bf3954f77437e

**Additional context**
I understand there is no point to use device code flow when broker is available, but it's better to make the whole workflow more user-friendly.

If MSAL exposes `enable_broker` rather than `allow_broker` (like what it is doing internally):

https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/2168027c4694b243ddeff08592396557d9848de0/msal/application.py#L564

life will be much easier - simply raise an error if `acquire_token_by_device_flow` is called on a `PublicClientApplication` with `enable_broker=True`.

Such `allow_...` design makes the logic far more complex - both `acquire_token_by_device_flow` and `acquire_token_silent_with_error` need to implement such fallback logic and be consistent with each other. In this issue, `acquire_token_by_device_flow` falls back to MSAL token cache, but `acquire_token_silent_with_error` doesn't and still reads from WAM.

rayluo:
[...]

PR #569, opened 2023-06-02
acquire_token_silent() shall not invoke broker if the account was not established by broker

rayluo:
This is useful when the user somehow chose to use device code flow successfully and now want to do an acquire_token_silent(). The event sequence in this scenario looks like this:

1. End user enabled broker.
2. For whatever reason, end user chooses to sign in using Device Code Flow (DCF). Broker (WAM) does not support DCF but, per our existing design, we do not want to break this scenario, so MSAL automatically falls back to use native DCF implementation. This is already happening today, and the sign-in would be successful, and access token (AT) and refresh token (RT) will be stored inside MSAL's native token cache.
3. Subsequent AcquireTokenSilent() call (especially those after the initial access token expires) would attempt broker code path, and end up with an "account not found" error from MsalRuntime's `read_account_by_id()` call, because the account was not established via MsalRuntime code path in step 2.
4. Currently, MSAL's AcquireTokenSilent() surfaces that MsalRuntime error to calling app, and that results in end user's token refresh attempt failure.
5. ~Per our recent discussion (in May 2023) with Azure CLI team and MsalRuntime team, this PR changes to catch error in step 3, and falls back to native token cache, whose RT will work.~
6. Per our recent discussion (in Nov 1st, 2023) with MSAL C++ team, this PR changes to record which flow established an account, and then bypass broker in its subsequent `acquire_token_silent()` calls. Also applies the same treatment to Username Password flow (a.k.a. ROPC). You may review this [internal design doc here](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/10474).

Note: Most of the conversation below between June to October were based on the outdated bullet point No.5, so, you do not need to read them.

@msamwils, you do not have to review the implementation details in this PR (which contains some other logistic changes), but please review the summary above.

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -17,7 +17,7 @@ from .authority import Authority, WORLD_WIDE
 from .mex import send_request as mex_send_request
 from .wstrust_request import send_request as wst_send_request
 from .wstrust_response import *
-from .token_cache import TokenCache, _get_username
+from .token_cache import TokenCache, _get_username, _GRANT_TYPE_BROKER
 import msal.telemetry
 from .region import _detect_region
 from .throttled_http_client import ThrottledHttpClient
@@ -1104,6 +1104,7 @@ class ClientApplication(object):
                     "home_account_id": a.get("home_account_id"),
                     "environment": a.get("environment"),
                     "username": a.get("username"),
+                    "account_source": a.get("account_source"),
 
                     # The following fields for backward compatibility, for now
                     "authority_type": a.get("authority_type"),
@@ -1398,7 +1399,10 @@ class ClientApplication(object):
             if account and account.get("authority_type") == _AUTHORITY_TYPE_CLOUDSHELL:
                 return self._acquire_token_by_cloud_shell(scopes, data=data)
 
-            if self._enable_broker and account is not None:
+            if self._enable_broker and account and account.get("account_source") in (
+                _GRANT_TYPE_BROKER,  # Broker successfully established this account previously.
+                None,  # Unknown data from older MSAL. Broker might still work.
+            ):
                 from .broker import _acquire_token_silently
                 response = _acquire_token_silently(
                     "https://{}/{}".format(self.authority.instance, self.authority.tenant),
@@ -1409,8 +1413,12 @@ class ClientApplication(object):
                         self._client_capabilities, claims_challenge),
                     correlation_id=correlation_id,
                     **data)
-                if response:  # The broker provided a decisive outcome, so we use it
-                    return self._process_broker_response(response, scopes, data)
+                if response:  # Broker provides a decisive outcome
+                    account_was_established_by_broker = account.get(
+                        "account_source") == _GRANT_TYPE_BROKER
+                    broker_attempt_succeeded_just_now = "error" not in response
+                    if account_was_established_by_broker or broker_attempt_succeeded_just_now:
+                        return self._process_broker_response(response, scopes, data)
 
             if account:
                 result = self._acquire_token_silent_by_finding_rt_belongs_to_me_or_my_family(
@@ -1441,6 +1449,8 @@ class ClientApplication(object):
                 response=response,
                 data=data,
                 _account_id=response["_account_id"],
+                environment=self.authority.instance,  # Be consistent with non-broker flows
+                grant_type=_GRANT_TYPE_BROKER,  # A pseudo grant type for TokenCache to mark account_source as broker
                 ))
             response[self._TOKEN_SOURCE] = self._TOKEN_SOURCE_BROKER
         return _clean_up(response)
@@ -1628,7 +1638,7 @@ class ClientApplication(object):
         """
         claims = _merge_claims_challenge_and_capabilities(
                 self._client_capabilities, claims_challenge)
-        if self._enable_broker:
+        if False:  # Disabled, for now. It was if self._enable_broker:
             from .broker import _signin_silently
             response = _signin_silently(
                 "https://{}/{}".format(self.authority.instance, self.authority.tenant),
--- a/msal/__main__.py
+++ b/msal/__main__.py
@@ -5,12 +5,20 @@
 
 Usage 1: Run it on the fly.
     python -m msal
+    Note: We choose to not define a console script to avoid name conflict.
 
 Usage 2: Build an all-in-one executable file for bug bash.
     shiv -e msal.__main__._main -o msaltest-on-os-name.pyz .
-    Note: We choose to not define a console script to avoid name conflict.
 """
-import base64, getpass, json, logging, sys, msal
+import base64, getpass, json, logging, sys, os, atexit, msal
+
+_token_cache_filename = "msal_cache.bin"
+global_cache = msal.SerializableTokenCache()
+atexit.register(lambda:
+    open(_token_cache_filename, "w").write(global_cache.serialize())
+    # Hint: The following optional line persists only when state changed
+    if global_cache.has_state_changed else None
+    )
 
 _AZURE_CLI = "04b07795-8ddb-461a-bbee-02f9e1bf7b46"
 _VISUAL_STUDIO = "04f0c124-f2bc-4f59-8241-bf6df9866bbd"
@@ -66,7 +74,7 @@ def _select_account(app):
     if accounts:
         return _select_options(
             accounts,
-            option_renderer=lambda a: a["username"],
+            option_renderer=lambda a: "{}, came from {}".format(a["username"], a["account_source"]),
             header="Account(s) already signed in inside MSAL Python:",
             )
     else:
@@ -76,7 +84,7 @@ def _acquire_token_silent(app):
     """acquire_token_silent() - with an account already signed into MSAL Python."""
     account = _select_account(app)
     if account:
-        print_json(app.acquire_token_silent(
+        print_json(app.acquire_token_silent_with_error(
             _input_scopes(),
             account=account,
             force_refresh=_input_boolean("Bypass MSAL Python's token cache?"),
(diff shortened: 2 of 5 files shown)


bgavrilMS:
I'm not sure that I agree with step 5, because it will violate the token protection promise. The scenario that can break is: 

- app uses interactive auth with browser and puts bearer RT in cache
- app upgrades to use broker
- app calls AcquireTokenSIlent and runtime's `read_account_by_id` fails because existing cached account
- app now uses it's cached bearer RT and breaks token protection CA

I would let the app handle this. If the app uses device code flow, they should not use broker when calling AcquireTokenSilent.

rayluo:
> ... it will violate the token protection promise. The scenario that can break is:

Not necessarily.

> 
> * app uses interactive auth with browser and puts bearer RT in cache
> * app upgrades to use broker
> * app calls AcquireTokenSIlent and runtime's `read_account_by_id` fails because (non-)existing cached account (in WAM?)

App may already handles this - Azure CLI did - by [suggesting end user to do an `az account clear` while switching to use broker](https://azureaggregator.wordpress.com/2023/06/01/azure-command-line-tools-build-2023-announcements-6). Therefore, the AcquireTokenSilent() simply will not be called. The end user will do a fresh "az login" which will go with broker code path and then subsequent AcquireTokenSilent() will work smoothly thereafter.

> * app now uses it's cached bearer RT and breaks token protection CA, even though it tries to use WAM

Even if the app or the end user did not clear the native MSAL's token cache, the bearer AT would not meet CA requirements indeed, at that point the resource is expected to reject such an AT with a meaningful error. This shall be a defined and expected error-handling situation of CA. The end user shall be prompted (by the app such as Azure CLI) to do "az login" again which in turn will run the broker code path.

> 
> I would let the app handle this. If the app uses device code flow, they should not use broker when calling AcquireTokenSilent.

Not quite easy. In native MSAL's calling pattern, when AcquireTokenSilent is called, there is no intrinsic state to indicate whether the existing signed-in session was established by device code flow or other flow. App could handle this, but that is an extra burden that shall be handled by MSALs, somehow (we do have other options here, such as expanding token cache schema to indicate what flow an account was established by; I thought that would be more heavyweight fix).

> headless Windows isn't a common scenario, I am not sure why we'd want to support this

You brought this up in an offline chat.

Indeed, this use case is rare, and that is probably why this is not (yet?) reported by end users, by only found by Azure CLI team's internal coverage tests.

On the other hand, besides of the "do not violate the token protection promise" design tenet, another implicit tenet is "do not break existing users and scenarios". The latter is the reason why native MSALs already have many fallback logic for scenarios that we know broker/WAM does not support: ADFS, B2C, etc., and device code flow is just one of them. Currently, when broker is enabled, device code flow becomes broken, which is not our intention.

bgavrilMS:
> > ... it will violate the token protection promise. The scenario that can break is:
> 
> Not necessarily.
> 
> > * app uses interactive auth with browser and puts bearer RT in cache
> > * app upgrades to use broker
> > * app calls AcquireTokenSIlent and runtime's `read_account_by_id` fails because (non-)existing cached account (in WAM?)
> 
> App may already handles this - Azure CLI did - by [suggesting end user to do an `az account clear` while switching to use broker](https://azureaggregator.wordpress.com/2023/06/01/azure-command-line-tools-build-2023-announcements-6). Therefore, the AcquireTokenSilent() simply will not be called. The end user will do a fresh "az login" which will go with broker code path and then subsequent AcquireTokenSilent() will work smoothly thereafter.
> 
> > * app now uses it's cached bearer RT and breaks token protection CA, even though it tries to use WAM
> 
> Even if the app or the end user did not clear the native MSAL's token cache, the bearer AT would not meet CA requirements indeed, at that point the resource is expected to reject such an AT with a meaningful error. This shall be a defined and expected error-handling situation of CA. The end user shall be prompted (by the app such as Azure CLI) to do "az login" again which in turn will run the broker code path.
> 
> > I would let the app handle this. If the app uses device code flow, they should not use broker when calling AcquireTokenSilent.
> 
> Not quite easy. In native MSAL's calling pattern, when AcquireTokenSilent is called, there is no intrinsic state to indicate whether the existing signed-in session was established by device code flow or other flow. App could handle this, but that is an extra burden that shall be handled by MSALs, somehow (we do have other options here, such as expanding token cache schema to indicate what flow an account was established by; I thought that would be more heavyweight fix).
> 
> > headless Windows isn't a common scenario, I am not sure why we'd want to support this
> 
> You brought this up in an offline chat.
> 
> Indeed, this use case is rare, and that is probably why this is not (yet?) reported by end users, by only found by Azure CLI team's internal coverage tests.
> 
> On the other hand, besides of the "do not violate the token protection promise" design tenet, another implicit tenet is "do not break existing users and scenarios". The latter is the reason why native MSALs already have many fallback logic for scenarios that we know broker/WAM does not support: ADFS, B2C, etc., and device code flow is just one of them. Currently, when broker is enabled, device code flow becomes broken, which is not our intention.

In any case, there is a burden on the app - either they have to remember that DCF is used and follow up with ATS + no-broker or they have to provide guidance to the user (!) to clear the cache. 

To me, DCF is never a good solution because it doesn't satisfy device CA. So I wouldn't optimize for DCF.

rayluo:
> In any case, there is a burden on the app - either they have to remember that DCF is used and follow up with ATS + no-broker or they have to provide guidance to the user (!) to clear the cache.

There shouldn't be a burden on the app. With this PR (or other different approaches with similar results), the app does not have to remember and differentiate between DCF and other flows. In particular, Azure CLI's guidance to the user (!) to clear the cache is debatable in itself, is not strictly necessary, and was NOT proposed for this DCF purpose (it was merely proposed as a hygiene measure: to not leave abandoned token cache files behind).

> 
> To me, DCF is never a good solution because it doesn't satisfy device CA. So I wouldn't optimize for DCF.

In a sense, the purpose of this PR is not to optimize for DCF. It is to fix a logical loophole. When MSAL calls MsalRuntime's ReadAccountById() and the outcome is semantically an AccountNotFound, would it be a meaningful indicator to say that "this account was not handled by me (broker), not my business, you (msal) shall deal with it yourself"?

MSamWils:
> This is useful when the user somehow chose to use device code flow successfully and now want to do an acquire_token_silent(). The event sequence in this scenario looks like this:
> 
> 1. End user enabled broker.
> 2. For whatever reason, end user chooses to sign in using Device Code Flow (DCF). Broker (WAM) does not support DCF but, per our existing design, we do not want to break this scenario, so MSAL automatically falls back to use native DCF implementation. This is already happening today, and the sign-in would be successful, and access token (AT) and refresh token (RT) will be stored inside MSAL's native token cache.
> 3. Subsequent AcquireTokenSilent() call (especially those after the initial access token expires) would attempt broker code path, and end up with an "account not found" error from MsalRuntime's `read_account_by_id()` call, because the account was not established via MsalRuntime code path in step 2.
> 4. Currently, MSAL's AcquireTokenSilent() surfaces that MsalRuntime error to calling app, and that results in end user's token refresh attempt failure.
> 5. Per our recent discussion with Azure CLI team and MsalRuntime team, this PR changes to catch error in step 3, and falls back to native token cache, whose RT will work.
> 
> @MSamWils, you do not have to review the implementation details in this PR (which contains some other logistic changes), but please review the summary above.

ReadAccountById returned account was not found means that there is no account corresponding to that ID in MSAL C++ cache. Whether or not how native MSAL interpret and behave is purely based on the scenario. For example:
- In the Azure CLI calling pattern, the assumption is that SignInInteractively has been called before where ReadAccountById should return the account that was just signed in previously. So if ReadAccountById returned account was not found, it could be a bug or the previously interactive call didn't go through the MSALRuntime.
- In other scenario when ReadAccountById return account was not found, it means that there is an account cache in native MSAL where it isn't in MSAL C++. One of the approaches could be to call SignInInteractively with the login_hint to propagate the account in MSAL C++ cache.
- There are also other cases where ReadAccountById did return the account. It will be another layer of complexity when debugging or investigating issues.

I agree with @bgavrilMS that if the MSALRuntime do not support the scenario, all the subsequent acquireToken calls should not go to the MSALRuntime. The application could give us the hint or we (in the native MSAL layer) to maintain the hint in the account cache to "remember" we are in this state.

rayluo:
FYI, @bgavrilMS , et al, you can ignore the lengthy and outdated conversation above, because the controversial bullet point No. 5 in the PR description has been replaced by a new bullet point No. 6.

bgavrilMS (approved):

msal/application.py
```diff
@@ -1628,7 +1638,7 @@ def acquire_token_by_username_password(
         """
         claims = _merge_claims_challenge_and_capabilities(
                 self._client_capabilities, claims_challenge)
-        if self._enable_broker:
+        if False:  # Disabled, for now. It was if self._enable_broker:
```
xiangyan99, line 1641: FYI. In v1, we also expose broker for InteractiveBrowserCred, there is no support for username password for broker in Python identity too.

msal/application.py
```diff
@@ -1398,7 +1399,10 @@ def _acquire_token_silent_from_cache_and_possibly_refresh_it(
             if account and account.get("authority_type") == _AUTHORITY_TYPE_CLOUDSHELL:
                 return self._acquire_token_by_cloud_shell(scopes, data=data)
 
-            if self._enable_broker and account is not None:
+            if self._enable_broker and account and account.get("account_source") in (
+                _GRANT_TYPE_BROKER,  # Broker successfully established this account previously.
+                None,  # Unknown data from older MSAL. Broker might still work.
+            ):
```
jiasli, line 1405: Besides fixing device code flow (https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/563), another great benefit of this PR is that enabling WAM will not break the current login context established by auth code flow. I provided more details in https://github.com/Azure/azure-cli/pull/28085#issuecomment-1870093562.

rayluo: Yes, that was an intentional design and precise implementation. We did not "advertise" it, though, because such an effect is controversial. Some might perceive it as lack of enforcement to utilize broker.

Luckily, that debate is not really necessary in Azure CLI's context, because [Azure CLI's broker instruction suggests end user to clear local cache before enabling broker](https://learn.microsoft.com/en-us/cli/azure/authenticate-azure-cli-web-account-manager).

Let's end this topic here. Further question can be carried out offline.

msal/application.py
rayluo: FYI v2: We [reverted course again](https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/702), back to the original.

CC: @jiasli 

Note: New comments, if any, shall be posted into that new feature request 702.

jiasli:
ROPC via WAM is enabled in  https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/712.

PR #579, opened 2023-07-18
app: Include trust_env as optional argument

WesleyFransin:
trust_env parameter was added to ClientApplication and wired up to requests.Session, allowing users to choose the behavior when dealing with proxies (use environment variables or manually set up). This can be useful when working behind company set proxies.

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -177,7 +177,7 @@ class ClientApplication(object):
             client_credential=None, authority=None, validate_authority=True,
             token_cache=None,
             http_client=None,
-            verify=True, proxies=None, timeout=None,
+            verify=True, proxies=None, trust_env=False, timeout=None,
             client_claims=None, app_name=None, app_version=None,
             client_capabilities=None,
             azure_region=None,  # Note: We choose to add this param in this base class,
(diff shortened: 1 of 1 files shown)


rayluo:
Thanks for your interest in MSAL, @WesleyFransin. However, we hesitate to take this change. The reason being, MSAL is an authentication library, it does not aim to duplicate all the fancy optional parameters that happen to be supported by an underlying http library. So, a generic - albeit less convenient - approach to address these kind of requests is to have the caller provide their own `http_client` with the desirable behavior.

msal/application.py
```diff
@@ -177,7 +177,7 @@ def __init__(
             client_credential=None, authority=None, validate_authority=True,
             token_cache=None,
             http_client=None,
-            verify=True, proxies=None, timeout=None,
+            verify=True, proxies=None, trust_env=False, timeout=None,
```
rayluo, line 180: Thanks for your interest in MSAL, @WesleyFransin. However, we hesitate to take this change. The reason being, MSAL is an authentication library, it does not aim to duplicate all the fancy optional parameters that happen to be supported by an underlying http library. So, a generic - albeit less convenient - approach to address these kind of requests is to have the caller provide their own `http_client` with the desirable behavior.

PR #580, opened 2023-07-19
Guarding against perf regression for acquire_token_for_client()

rayluo:
## Goals defined in early July 2023

This PR may eventually accomplish the following goals.

1. Scenarios: A multi-tenant S2S with memory token cache which is (1) initially empty, and all incoming requests will generate cache miss; or (2) prepopulated with 1000 tenants and 10 tokens per tenant, and all incoming requests will generate cache hit.
2. Metrics: latency in ms.
3. How data is collected and stored (Aim for 1 DB schema)? TBD
4. How do we read the historical data into the build (from that central DB described in Q3)? (The intention is that we will have a baseline to compare to, and break the build on regression.)

## UPDATE at late July 2023

We followed the spirit of the goals above and pivoted along the way.

1. This PR populates the token cache for test cases in the following matrix, and during the benchmark, token cache remains unchanged, so that its size would not grow.

|                                             | New token request result in cache hits | New token requests result in cache miss |
|---------------------------------------------|----------------------------------------|-----------------------------------------|
| ~1 tenant, 1 token in cache~                |                                        |                                         |
| 1 tenant, 10 tokens in cache                |                                        |                                         |
| ~1 tenant, 1000 tokens in cache~            |                                        |                                         |
| 1000 tenants, 10 tokens per tenant in cache |                                        |                                         |

2. Metrics. Our chosen benchmark tool happens to use "operations per second". Internally we use "time elapsed during a test case". They are mathematically equivalent.

3. Collecting data into a central database? The purpose was to render historical data into a nice chart. We found a [lightweight Github Benchmark Action](https://github.com/marketplace/actions/continuous-benchmark) that can render charts without dependency of a central database. See the [diagram generated by this PR](https://azuread.github.io/microsoft-authentication-library-for-python/dev/bench/).

4. How to get baseline data to be used as a baseline to detect regression?

   There are two mechanisms available. Neither of them requires a central DB. We may choose either or both of them.

   * The Github Benchmark Action we use will store historical benchmark data into a special git branch.

   * ~We can also dynamically benchmark a stable computation load as a reference, and then comparing our MSAL performance against that known load. For example, the load can be "the time needed to calculate fibonacci sequence for 20", or just actually benchmark the built-in dictionary implementation of the language we are using. And then we can benchmark MSAL's scenarios (see our goal 1 above), which will be - let's say - 1.234 times slower than the reference. Now, our threshold can be defines as "the performance shall not be longer than 1.234 * (1 + threshold)".~ This experiment was abandoned, because its reference workload has different characteristic than those real workloads, and the ratio between real workload and reference seems different on different test machines.

   In either approach, the variation is in the range of double-digit percentage. The Github Benchmark Action uses a 200% threshold as its default value.

## Implementation Details (useful when you want to review this PR's implementation)

A Perf Test solution contains 3 high level components:

(a) a tailored *simulator* that can generate load for our test subject;
(b) a *perf tester* that can run those *simulators" many times, in deterministic amount of time, producing some statistics, preferably with a small standard deviation;
(c) a mechanism to store historical results of *perf tester*, in order to detect regression and render wow diagrams.

We have to do "a. simulator" ourselves, which we did, in the file `tests/simulator.py`.

We chose the "Continuous Benchmark" github action which is mainly a tool for "c. detect and render", but it cleverly consumes the output of many other tools that operate in "b. perf tester" space, one per language. In this PR, we use `pytest-benchmark` as our "perf tester", with test cases defined in `test_benchmark.py`.

The configuration for Continuous Benchmark is in `python-package.yml`.


.github/workflows/python-package.yml
```diff
@@ -53,15 +53,40 @@ jobs:
...
+    - name: Run benchmark
+      run: pytest --benchmark-only --benchmark-json benchmark.json --log-cli-level INFO tests/test_benchmark.py
+    - name: Render benchmark result
+      uses: benchmark-action/github-action-benchmark@v1
+      with:
+        tool: 'pytest'
+        output-file-path: benchmark.json
+        fail-on-alert: true
+    - name: Publish Gibhub Pages
+      run: git push origin gh-pages
```
pmaytak, line 89: @rayluo Does this work? I was trying to enable the action to run on PRs but not push to charts like [here](https://github.com/benchmark-action/github-action-benchmark#run-only-on-your-branches) but I get [this error](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/actions/runs/5769563984/job/15641719173).

> Run git push origin gh-pages:gh-pages
>   git push origin gh-pages:gh-pages
>   shell: /usr/bin/bash -e {0}
>   env:
>     DOTNET_ROOT: /usr/share/dotnet
**error: src refspec gh-pages does not match any**
**error: failed to push some refs to 'https://github.com/AzureAD/microsoft-authentication-library-for-dotnet'**
> Error: Process completed with exit code 1.

pmaytak: I guess you can disregard this. There seem to be some weird issues internally with the benchmark action.

rayluo: Sorry, I meant to answer this but forgot.

The line here works, as you can see the diagram has been successfully pushed to the Github Page.

I guess the error you ran into was before you created the gh-pages branch. It is fine now. @pmaytak

.github/workflows/python-package.yml
```diff
@@ -42,15 +42,40 @@ jobs:
         python -m pip install --upgrade pip
         python -m pip install flake8 pytest
         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+    - name: Test with pytest
+      run: pytest --benchmark-skip
```
pmaytak, line 46: What does this end up doing if it skips benchmarks? I guess it runs other tests unit//integration?

.github/workflows/python-package.yml
```diff
@@ -42,15 +42,40 @@ jobs:
...
       run: |
-        pytest
+        python -m pip install --upgrade pip
+        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+    - name: Run benchmark
+      run: pytest --benchmark-only --benchmark-json benchmark.json --log-cli-level INFO tests/test_benchmark.py
+    - name: Render benchmark result
+      uses: benchmark-action/github-action-benchmark@v1
+      with:
+        tool: 'pytest'
```
pmaytak, line 74: Should we set alert-threshold? By default it's 200% (so twice as slow perf regression). In .NET we set it to 130%

tests/simulator.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def __init__(self, number_of_tenants=1, tokens_per_tenant=1, cache_hit=False):
+        logger.info(
+            "number_of_tenants=%d, tokens_per_tenant=%d, cache_hit=%s",
+            number_of_tenants, tokens_per_tenant, cache_hit)
+        with patch.object(msal.authority, "tenant_discovery", return_value={
+            "authorization_endpoint": "https://contoso.com/placeholder",
+            "token_endpoint": "https://contoso.com/placeholder",
+        }) as _:  # Otherwise it would fail on OIDC discovery
+            self.apps = [  # In MSAL Python, each CCA binds to one tenant only
```
pmaytak, line 35: > In MSAL Python, each CCA binds to one tenant only

Interesting. Do you mean 1 CCA per tenant or per authority? What about multi-tenant apps or org / consumer / common authorities?

In .NET tests we just use 1 CCA instance with default authority (common) since it doesn't matter for tests..

.github/workflows/python-package.yml
rayluo: Indeed. Basically, we want to run all the other non-perf test first, and then only run perf test when the MSAL is functionally correct. Otherwise there will be no point to benchmark an incorrect commit/PR.

.github/workflows/python-package.yml
rayluo: Based on the [diagram so far](https://azuread.github.io/microsoft-authentication-library-for-python/dev/bench/), the peak and valley can easily have a swing of 160% while the underlying test subject did not change. Since each sample is individually reliable (they have small stdev), it seems like the fluctuation was caused by the Github Action agent?

Anyway, 130% would already result in false positive for MSAL Python, we will stay with the default 200% for now. (At one point, we even discussed using a much higher alert-threshold, didn't we?)

pmaytak (approved):

PR #588, opened 2023-08-09
Fix typo in test names (warnning → warning)

musicinmybrain:
This simply renames a few tests.


rayluo (approved):
Thanks for your catch!

PR #590, opened 2023-08-18
Add enable_pii and wire it up with MsalRuntime

### LINKED ISSUE #568: MSAL logging discrepancy
For posterity, I wanted to open this issue since there has been discussion within our team regarding a recent feature request from the MSAL team where Azure Identity should allow extraction of all MSAL logs by enabling some PII boolean flag or increasing log verbosity: https://github.com/Azure/azure-sdk-for-python/issues/29406. 

As far as I know, other MSAL language SDKs support toggling PII logging through some flag, but this SDK does not. From previous discussion with @rayluo, it was chosen to not enable PII logging here since it was felt that the info wasn't useful for troubleshooting.

Would appreciate some clarification on the following:

- Is this discrepancy is intentional/acceptable? 
- If PII was found to be useful from a customer support standpoint for other languages, could we enable similar functionality in MSAL for Python?
- Currently, is all logging control within MSAL based on setting the log level (`logging.basicConfig(level=logging.DEBUG)`)?

rayluo:
A new `enable_pii` parameter will be added into API surface. To use this new behavior, an app would need to declare dependency of `msal>=1.24,<2`.

This will resolve #568


jiasli:
Can we make sure `msal_telemetry` contains no PII when `enable_pii_log` is set to `True`? In my opinion, `msal_telemetry` should not contain PII under any circumstances.

rayluo:
> Can we make sure `msal_telemetry` contains no PII when `enable_pii_log` is set to `True`? In my opinion, `msal_telemetry` should not contain PII under any circumstances.

If I remember correctly, that was indeed the design. `enable_pii_log` only affects logs. I'm also CCing @msamwils , @ashok672 , @fengga to keep me honest.

PR #593, opened 2023-09-16
Placeholder

rayluo:


bgavrilMS (changes requested):
1. I don't think `http://localhost:port/path` is a valid redirect uri for a desktop app using system browser. The HTTP listener built in MSAL can't tell between `http://localhost:port/path1` and `http://localhost:port/path2`. I am concerned about potential vulnerability this can introduce.

2. Please do not change the public API of MSAL with an OK from @localden and a discussion with other MSAL owners

3. For Public Client, please loop in @ashok672 and @iulico-1  as they need to maintain this.

rayluo:
Agreed with your suggestion 2 & 3. Will wait for the input from those pinged folks.

With regard to suggestion 1, the `http://localhost:port/path` is a valid redirect uri for a desktop app using system browser, per [the specs](https://learn.microsoft.com/en-us/azure/active-directory/develop/reply-url#localhost-exceptions). So, it is a matter of whether/how it will be implemented.

> The HTTP listener built in MSAL can't tell between `http://localhost:port/path1` and `http://localhost:port/path2`. I am concerned about potential vulnerability this can introduce.

That is an implementation detail that can be discussed offline. Traditionally, MSAL uses ephemeral port, state validation and PKCE for security.

bgavrilMS:
> Agreed with your suggestion 2 & 3. Will wait for the input from those pinged folks.
> 
> With regard to suggestion 1, the `http://localhost:port/path` is a valid redirect uri for a desktop app using system browser, per [the specs](https://learn.microsoft.com/en-us/azure/active-directory/develop/reply-url#localhost-exceptions). So, it is a matter of whether/how it will be implemented.
> 
> > The HTTP listener built in MSAL can't tell between `http://localhost:port/path1` and `http://localhost:port/path2`. I am concerned about potential vulnerability this can introduce.
> 
> That is an implementation detail that can be discussed offline. Traditionally, MSAL uses ephemeral port, state validation and PKCE for security.

That's an excellent document you found. I am still not understanding the scenario though:

`To register multiple redirect URIs on localhost to test different flows during development, differentiate them using the path component of the URI. For example, http://localhost/MyWebApp doesn't match http://localhost/MyNativeApp`

Why would this ^^ be useful?

rayluo:
> That's an excellent document you found. So I guess the user scenario is to use 1 app registration for both web and desktop? Smth like: ...

That is correct. That is the scenario that app developers run into, every time they need to start their app locally for development or debug purpose.

msal/application.py
```diff
@@ -1788,7 +1789,16 @@ def acquire_token_interactive(
         :param int port:
             The port to be used to listen to an incoming auth response.
             By default we will use a system-allocated port.
-            (The rest of the redirect_uri is hard coded as ``http://localhost``.)
+            The unspecified parts of the redirect_uri is hard coded as ``http://localhost``.
+            Only useful when using browser (i.e. not broker).
+
+        :param str path:
+            The path to be used in the redirect uri.
```
localden, line 1796:
```suggestion
            The path to be used in the redirect URI.
```

msal/application.py
```diff
@@ -1788,7 +1789,16 @@ def acquire_token_interactive(
         :param int port:
             The port to be used to listen to an incoming auth response.
             By default we will use a system-allocated port.
-            (The rest of the redirect_uri is hard coded as ``http://localhost``.)
+            The unspecified parts of the redirect_uri is hard coded as ``http://localhost``.
```
localden, line 1792:
```suggestion
            The unspecified parts of the ``redirect_uri`` are hard coded as ``http://localhost``.
```

msal/application.py
```diff
@@ -1788,7 +1789,16 @@ def acquire_token_interactive(
...
+            Only useful when using browser (i.e. not broker).
```
localden, line 1793:
```suggestion
            Only useful when using system browser authentication (i.e., not an authentication broker like WAM).
```

msal/application.py
```diff
@@ -1788,7 +1789,16 @@ def acquire_token_interactive(
...
+
+        :param str path:
+            The path to be used in the redirect uri.
+            By default we will use no path.
```
localden, line 1797:
```suggestion
            By default, no path is used.
```

msal/application.py
```diff
@@ -1788,7 +1789,16 @@ def acquire_token_interactive(
...
+            The unspecified parts of the redirect_uri is hard coded as ``http://localhost``.
```
localden, line 1798:
```suggestion
            The unspecified parts of the ``redirect_uri`` are hard coded as ``http://localhost``.
```

msal/application.py
```diff
@@ -1788,7 +1789,16 @@ def acquire_token_interactive(
...
+            Only useful when using browser (i.e. not broker).
```
localden, line 1799:
```suggestion
            Only useful when using system browser authentication (i.e., not an authentication broker like WAM).
```

localden (changes requested):

localden:
Minor feedback on documentation, @rayluo - other than that, looking good.

localden (approved):

PR #596, opened 2023-09-19
Integration with Broker-on-Mac

rayluo:
This PR has been tested by an internal bug bash.

~It will remain as a draft until the dependency `PyMsalRuntime 0.14` is released on PyPI.~
This PR depends on [`PyMsalRuntime 0.17`](https://pypi.org/project/pymsalruntime/0.17.0/) for its full support for SSH Cert and ROPC.


msal/__main__.py
```diff
@@ -204,7 +206,7 @@ def _main():
             header="Input authority (Note that MSA-PT apps would NOT use the /common authority)",
             accept_nonempty_string=True,
             ),
-        allow_broker=allow_broker,
+        enable_broker=enable_broker,
```
bgavrilMS, line 209: I thought we decided on an OS explicit flag, i.e. enable on Windows, enable on Mac etc. ?

bgavrilMS: We discussed in person: 

- MSAL should fallback to browser if msal_runtime cannot be initialized (i.e. broker not available)
- an opt-in flag for each operating system

rayluo: > We discussed in person:

Yes. My message above was sent in the last minute before our discussion yesterday.

> * MSAL should fallback to browser if msal_runtime cannot be initialized (i.e. broker not available)

ACK. I double checked the current implementation. It [differentiates two different kinds of errors](https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/1.24.0/msal/broker.py#L23-L27) during the construction. 

1. If app developer opted in but the package PyMsalRuntime was not installed, MSAL Python emits a helpful error message to guide them install the dependency. Error-out is desirable here because it is actionable to the app developers.
2. If app developer opted in and PyMsalRuntime was also installed but failed in its initialization, MSAL Python [emits logs and falls back](https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/1.24.0/msal/application.py#L596-L600).

> * an opt-in flag for each operating system

I believe we meant "proposal 3" above (now I **highlighted** the differences between proposal 2 & 3 in the [original message](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/596#discussion_r1331939378)). Specifically, we are going with "3 flags, one per OS", and MSAL will automatically ignore "enable_broker_on_X" when current platform is Y.

Let me work out a new commit for that.

bgavrilMS: Thank you @rayluo , that makes sense.

msal/application.py
```diff
@@ -580,34 +541,47 @@ def __init__(
...
+        is_confidential_app = self.client_credential or isinstance(
+            self, ConfidentialClientApplication)
         if is_confidential_app and allow_broker:
             raise ValueError("allow_broker=True is only supported in PublicClientApplication")
-        self._enable_broker = False
-        if (allow_broker and not is_confidential_app
-                and sys.platform == "win32"
+            # Historically, we chose to support ClientApplication("client_id", allow_broker=True)
+        if allow_broker:
+            warnings.warn(
```
bgavrilMS, line 562: To maintain backwards compat, would it be better to enable the windows broker if customer uses old flag?

bgavrilMS (approved):

msal/__main__.py
```diff
@@ -204,7 +204,8 @@ def _main():
             header="Input authority (Note that MSA-PT apps would NOT use the /common authority)",
             accept_nonempty_string=True,
             ),
-        allow_broker=allow_broker,
+        enable_broker_on_windows=enable_broker,
```
xiangyan99, line 207: We used to use `allow_broker` with the meaning try enable it but ignore if it is not available.

Do we want to change the behavior to enable it or raise error if we fail to enable it?

rayluo: @bgavrilMS , fyi, @xiangyan99 also asked a follow-up question offline. "Now that we have two `enable_broker_on_windows` and `enable_broker_on_mac`, aren't they mutually exclusive? An app cannot be running on Windows and Mac *at the same time*".

I think @xiangyan99 made a good point here.

The wording "enable something" was only appropriate when the first commit of this PR used *only one* `enable_broker` parameter and was expecting app developers to set it conditionally per platform. In that case, "enable_broker=True" at runtime would indeed mean "enable it right now, or raise error if we fail to (acquire tokens via broker)".

Now we switched to the *two parameters* approach, then the meaning should really be `allow_broker_on_windows` and `allow_broker_on_mac`.

msal/application.py
```diff
@@ -1807,10 +1834,23 @@ def acquire_token_interactive(
...
+
+              - If your app is a GUI app running on modern Windows system,
+                you are required to also provide its window handle,
+                so that the sign-in window will pop up on top of your window.
+              - If your app is a console app runnong on Windows system,
+                you can use a placeholder
+                ``PublicClientApplication.CONSOLE_WINDOW_HANDLE``.
+              - If your app is running on Mac,
+                you can use a placeholder
+                ``PublicClientApplication.CONSOLE_WINDOW_HANDLE``.
```
rayluo, line 1853: FYI, those lines are the changes (or the lack thereof) of the `parent_window_handle` requirement between Windows and Mac.

The Windows behavior was influenced by the MsalRuntime's desire to make window handle explicit (CC @msamwils), and now on Mac this parameter remains required. (CC @kaisong1990 )

Downstream library may choose to make this parameter optional, by using MSAL's predefined CONSOLE_WINDOW_HANDLE placeholder as a default value. (CC @xiangyan99 )

xiangyan99: Sorry, you mean it is required on Mac because it is required on Windows?

I am a little confused.

rayluo: > Sorry, you mean it is required on Mac because it is required on Windows?
> 
> I am a little confused.

Put it this way. Most Python scripts are console apps and they shall ideally be cross-platform. How do we want those app developers to write their source code?

1. 
   ```python
   app.acquire_token_interactive(
       ...,
       parent_window_handle=app.CONSOLE_WINDOW_HANDLE,  # Same placeholder across all platforms, even though it can be a no-op on some platforms
       ...)
   ```
2. 
   ```python
   if sys.platform == "win32":
       app.acquire_token_interactive(
           ...,
           parent_window_handle=app.CONSOLE_WINDOW_HANDLE,  # Required on Windows
       ...)
   else:
       app.acquire_token_interactive(
           ...,
           # No need to specify parent_window_handle at all
       ...)
   ```

msal/application.py
```diff
@@ -1728,9 +1702,62 @@ class PublicClientApplication(ClientApplication):  # browser app or mobile app
...
+            This parameter defaults to None, which means MSAL will not utilize a broker.
         """
         if client_credential is not None:
             raise ValueError("Public Client should not possess credentials")
+        # Using kwargs notation for now. We will switch to keyword-only arguments.
+        enable_broker_on_windows = kwargs.pop("enable_broker_on_windows", False)
+        enable_broker_on_mac = kwargs.pop("enable_broker_on_mac", False)
+        self._enable_broker = bool(
+            enable_broker_on_windows and sys.platform == "win32"
+            or enable_broker_on_mac and sys.platform == "darwin")
```
jiasli, line 1760: Won't `enable_broker_on_windows and sys.platform == "win32" or enable_broker_on_mac and sys.platform == "darwin"` already return a `bool`? The casting seems to be unnecessary.

rayluo: I was mainly using the `(` and `)` for Python's implicit line joining, and then added a `bool` as well.

It is also technically desirable, because the input `enable_broker_on_windows` might not be a bool.

Issue #602, opened 2023-10-02
[Feature Request] Improve default token caching


bgavrilMS:
As a repair item from incidents where client_credentils apps were not using cached tokens: 

- [ ] Produce an indication that the token comes from the cache (e.g. boolean AuthenticationResult.IsFromCache or AuthenticationResult.TokenSource - IdP, TokenCache, Broker)
- [ ] Allow apps to do static caching (i.e. shared per process), e.g. `cca.WithCacheOptions(SharedCache)`
- [ ] Update client_creds samples to showcase the 2 new APIs.

rayluo:
> Produce an indication that the token comes from the cache (e.g. boolean AuthenticationResult.IsFromCache or AuthenticationResult.TokenSource - IdP, TokenCache, Broker)

While it is easy to add this indication into our public API, how do we envision app developers to utilize it programmatically? It is not actionable to them to do "if TokenSource != TokenCache then DoSomething()", because cache miss is a fairly common event, especially in CAE.

Perhaps what we really want is to guide app developer to persist and reuse token cache. A better indicator can therefore be "is the current token cache completely empty (note that expired tokens still count as non-empty cache)". If the cache is empty, then it is likely the calling app did not persist the cache correctly, which warrants MSAL to emit a warning log and/or send telemetry. No public API change is required for this proposal.

> Allow apps to do static caching (i.e. shared per process), e.g. cca.WithCacheOptions(SharedCache)

[MSAL Python's `cache` parameter](https://msal-python.readthedocs.io/en/latest/#msal.ClientApplication.params.cache) always allow flexible caching. For example:

```python
import msal

global_cache = msal.TokenCache()  # Now this becomes a static cache

# For whatever reason, multiple CCA will be created, but they can all share the same token cache
for i in range(10):
    cca = msal.ConfidentialClientApplication(
        "my_client_id",
        client_credential=...,
        ...,
        cache=global_cache)
    token = cca.acquire_token_for_client(["scope"])
```

> Update client_creds samples to showcase the 2 new APIs.

Looks like no new API is necessary. But we can and shall add a new sample to demonstrate the shared token cache pattern, which can be useful in the "pre-calculated client assertion" scenario.

We can discuss (online or offline) on those proposals above, or I can proceed to draft some PRs to further demonstrate those proposals. Let me know.

bgavrilMS:
### On the topic of TokenSource

[...]

PR #603, opened 2023-10-03
Demonstrate how to use global token cache

rayluo:
This PR addresses the part of #602 .

This is currently a draft but still ready for review. I'll update this PR to modify more samples, probably including PCA samples.

### DIFF
--- a/sample/confidential_client_certificate_sample.py
+++ b/sample/confidential_client_certificate_sample.py
@@ -31,6 +31,7 @@ You can then run this sample with a JSON configuration file:
 import sys  # For simplicity, we'll read config file from 1st CLI param sys.argv[1]
 import json
 import logging
+import time
 
 import requests
 import msal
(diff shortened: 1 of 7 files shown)


sample/confidential_client_secret_sample.py
```diff
@@ -41,26 +42,36 @@
 
 config = json.load(open(sys.argv[1]))
 
-# Create a preferably long-lived app instance which maintains a token cache.
+persisted_cache = msal.SerializableTokenCache()  # See also https://msal-python.readthedocs.io/en/latest/#msal.SerializableTokenCache
```
rayluo, line 45: Moving the [conversation in #602](https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/602#issuecomment-1745425229) to here.

> Agree on persisting the cache to disk for public client samples.
>
> But confidential clients don't really make that much use of I/O. Besides, in some Azure environments you don't even have access to the disk.

Do you mean a CCA sample shall NOT bother demonstrating the persist-to-disk pattern? I understand that it is not one-size-fits-all (which is also why MSAL Python's `token_cache` parameter allows caller to use different flavors of TokenCache), but I also think it is harmless to demonstrate a persist-to-disk pattern here. Besides, the current persist-to-disk sample does not actually incur I/O until the process exits and restarts.

The main reason to implement persist-to-disk pattern, however, is that [the dummy code snippet mentioned in #602](https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/602#issuecomment-1744309904) (quoted below) feels silly:

```python
global_cache = msal.TokenCache()  # Now this becomes a static cache

# For whatever reason, multiple CCA will be created, but they can all share the same token cache
for i in range(10):
    cca = msal.ConfidentialClientApplication(..., cache=global_cache)
    token = cca.acquire_token_for_client(["scope"])
```

In most cases, app developer would better use this simpler approach:

```python
global_cca = msal.ConfidentialClientApplication(...)
# Now, global CCA's default in-memory token cache becomes a static cache,
# because the CCA is global/static, which is what we recommend.

for i in range(10):
    token = cca.acquire_token_for_client(["scope"])
```

So, perhaps these two action items?

* For most CCA samples, we shall simply change all `app = msal.ConfidentialClientApplication(...)` lines into `global_app = ...`, without using persisted cache.
* I will revisit the CCA sample using pre-computed client assertion, and see what changes are needed there. That is the (only?) scenario that it is not obvious how to persist CCA or its token cache.

rayluo:
Hi @bgavrilMS , this PR draft shows two different approaches. Let me know which one you prefer, and then I'll apply that chosen style to all the samples and close this topic.

sample/confidential_client_certificate_sample.py
```diff
@@ -42,27 +43,35 @@
 
 config = json.load(open(sys.argv[1]))
 
-# Create a preferably long-lived app instance which maintains a token cache.
-app = msal.ConfidentialClientApplication(
+# If for whatever reason you plan to recreate same ClientApplication periodically,
+# you shall create one global token cache and reuse it by each ClientApplication
+global_token_cache = msal.TokenCache()  # More options in https://msal-python.readthedocs.io/en/latest/#tokencache
```
bgavrilMS, line 48: Maybe a better name is `in_memory_token_cache`

rayluo: But, my understanding is that this round of effort focuses on "do static caching (i.e. shared per process)" (which was quoted from "[Feature Request] Improve default token caching #602").

The term "global" is closer to "static" in terms of multiple CCA in the same process will share such a cache. The term "in memory" is not. There can be many CCA, each has their own default in-memory cache, but not shared with each other.

Besides, when/if app developer would choose to use a persisted token cache, the line can still be `global_token_cache = msal.SerializableTokenCache()` (I plan to also change the other sample here to use `global_token_cache`).

PR #604, opened 2023-10-06
Mark package as supporting Python 3.12

tonybaloney:
It was already being tested, but 3.12 is GA now.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -26,7 +26,7 @@ jobs:
     runs-on: ubuntu-latest  # It switched to 22.04 shortly after 2022-Nov-8
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9, "3.10", "3.11", "3.12-dev"]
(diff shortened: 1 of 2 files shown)


.github/workflows/python-package.yml
```diff
@@ -26,7 +26,7 @@ jobs:
     runs-on: ubuntu-latest  # It switched to 22.04 shortly after 2022-Nov-8
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9, "3.10", "3.11", "3.12-dev"]
+        python-version: [3.7, 3.8, 3.9, "3.10", "3.11", "3.12"]
```
rayluo, line 29: Do you know whether we can also add "3.13-dev" here?

tonybaloney: No, you can't yet.

rayluo: OK then.

rayluo (approved):

PR #610, opened 2023-10-25
Expose token_source for observability

rayluo:
This will address #602 's first requirement. The successful token acquisition result will have a new key `token_source` whose value will be `idp`, `cache` or `broker`.

Existing unit test cases have been updated to test this new behavior.

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -176,6 +176,10 @@ class ClientApplication(object):
     REMOVE_ACCOUNT_ID = "903"
 
     ATTEMPT_REGION_DISCOVERY = True  # "TryAutoDetect"
+    _TOKEN_SOURCE = "token_source"
+    _TOKEN_SOURCE_IDP = "identity_provider"
+    _TOKEN_SOURCE_CACHE = "cache"
+    _TOKEN_SOURCE_BROKER = "broker"
 
     def __init__(
             self, client_id,
@@ -998,6 +1002,8 @@ class ClientApplication(object):
                     self._client_capabilities,
                     auth_code_flow.pop("claims_challenge", None))),
             **kwargs))
+        if "access_token" in response:
+            response[self._TOKEN_SOURCE] = self._TOKEN_SOURCE_IDP
         telemetry_context.update_telemetry(response)
         return response
 
@@ -1070,6 +1076,8 @@ class ClientApplication(object):
                         self._client_capabilities, claims_challenge)),
                 nonce=nonce,
                 **kwargs))
+            if "access_token" in response:
+                response[self._TOKEN_SOURCE] = self._TOKEN_SOURCE_IDP
             telemetry_context.update_telemetry(response)
             return response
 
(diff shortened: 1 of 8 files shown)


msal/application.py
```diff
@@ -176,6 +176,10 @@ class ClientApplication(object):
     REMOVE_ACCOUNT_ID = "903"
 
     ATTEMPT_REGION_DISCOVERY = True  # "TryAutoDetect"
+    _TOKEN_SOURCE = "token_source"
+    _TOKEN_SOURCE_IDP = "idp"
```
bgavrilMS, line 180: nit: IDP might be an acronym that is too specific to Identity geeks. Consider using "IdentityProvider" string in constants that are part of the public api.

tests/test_application.py
```diff
@@ -581,6 +597,7 @@ def mock_post(url, headers=None, *args, **kwargs):
             self.assertEqual("4|523,0|", (headers or {}).get(CLIENT_CURRENT_TELEMETRY))
             return MinimalResponse(status_code=200, text=json.dumps({"access_token": at}))
         result = self.app.acquire_token_on_behalf_of("assertion", ["s"], post=mock_post)
+        self.assertEqual(result[self.app._TOKEN_SOURCE], self.app._TOKEN_SOURCE_IDP)
```
bgavrilMS, line 600: Is the _CONST notation denoting a private constant? Should it be more "public" ?

bgavrilMS (approved):

msal/application.py
rayluo: I thought you mentioned the term `IdP` in the feature request #602. But yeah, how about this?

```suggestion
    _TOKEN_SOURCE_IDP = "identity_provider"
```

W.R.T. the leading underscore naming, indeed it means private in Python convention. My thought here is that since MSAL Python returns a plain-old dictionary `{"access_token": "...", ..., "token_source": "identity_provider"}` (rather than returning an AuthenticationResult class), those visible string literals themselves (such as `token_source` and `identity_provider`) become the *real* public API. Therefore, the `_TOKEN_SOURCE` and `_TOKEN_SOURCE_IDP` names are mainly just MSAL Python internal constants which we use to avoid duplicating magic strings everywhere.

Bottom line is, I think we can go with the _PRIVATE notation for now. If a need arises in the future that we would want to promote them to become public, we can always do that later.

sample/confidential_client_certificate_sample.py
```diff
@@ -63,6 +63,7 @@ def acquire_and_use_token():
     result = global_app.acquire_token_for_client(scopes=config["scope"])
 
     if "access_token" in result:
+        print("Token was obtain from:", result["token_source"])  # Since MSAL 1.25
```
rayluo, line 66: FYI, @bgavrilMS , this is what I meant by ["the string literals become the real API"](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/610#discussion_r1373509560).

BTW, all the built-in samples in this repo have been updated with this line. So, this will also satisfy the last requirement of #602 .

bgavrilMS: Small typo "obtained"

bgavrilMS (approved):

PR #613, opened 2023-10-27
Deprecate allow_broker, use enable_broker_on_windows

rayluo:
This PR deprecates the old `ClientApplication.__init__(..., allow_broker=True)`. Please use `PublicClientApplication.__init__(..., enable_broker_on_windows=True)` instead.

This change is the first step for us to moving to per-platform opt-in flag.

This PR will also close #510 , because we no longer go with that direction.

@jiasli , @xiangyan99,
* You do not have to immediately make a dependency on this change, but you are still recommended to expose the per-platform opt-in flag in your code base accordingly
* Please run your smoke test with this PR, as it will likely become the last change in an upcoming MSAL Python 1.25

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -181,6 +181,8 @@ class ClientApplication(object):
     _TOKEN_SOURCE_CACHE = "cache"
     _TOKEN_SOURCE_BROKER = "broker"
 
+    _enable_broker = False
+
     def __init__(
             self, client_id,
             client_credential=None, authority=None, validate_authority=True,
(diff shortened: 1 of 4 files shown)


msal/application.py
```diff
@@ -1746,9 +1720,53 @@ class PublicClientApplication(ClientApplication):  # browser app or mobile app
...
+            3. tested with ``acquire_token_interactive()`` and ``acquire_token_silent()``.
+
+        :param boolean enable_broker_on_windows:
+            This setting is only effective if your app is running on Windows 10+.
+            This parameter defaults to None, which means MSAL will not utilize a broker.
         """
         if client_credential is not None:
             raise ValueError("Public Client should not possess credentials")
+        # Using kwargs notation for now. We will switch to keyword-only arguments.
+        enable_broker_on_windows = kwargs.pop("enable_broker_on_windows", False)
```
jiasli, line 1768: Unlike `allow_broker` which is supported by both `PublicClientApplication` and `ClientApplication`, `enable_broker_on_windows` is now only supported by `PublicClientApplication`. Doesn't https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/559 stand anymore?

jiasli: Just because `allow_broker` is deprecated, #559 should no longer be considered a reasonable design.

Using the old `allow_broker` on a `ConfidentialClientApplication` fails on the application level:

https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/9ad3fd7565c91667e111063ee560e1cbdbe53b41/msal/application.py#L570-L571

But now using the new `enable_broker_on_windows` on a `ConfidentialClientApplication` fails on Python level:

```py
  File "d:\cli\azure-cli\src\azure-cli-core\azure\cli\core\auth\msal_authentication.py", line 132, in __init__
    super().__init__(service_principal_auth.client_id, client_credential=client_credential, **kwargs)
TypeError: ClientApplication.__init__() got an unexpected keyword argument 'enable_broker_on_windows'
```

The reason I would like to confirm the new design is because Azure CLI has an open PR https://github.com/Azure/azure-cli/pull/26482 to address the using `allow_broker` on `ConfidentialClientApplication` issue. If using `enable_broker_on_windows` is limited to `PublicClientApplication`, I will make the corresponding change in Azure CLI.

rayluo: > Just because `allow_broker` is deprecated, #559 should no longer be considered a reasonable design.

Why is #559 even relevant here? It was just a one-line document change which does not change any behavior.

Let's get back to the topic of this PR. This PR does not really change the behavior, either, because the intention of rejecting `ConfidentialClientAppication(..., broker=True)` remains the same. Your Azure CLI PR of adopting the new parameter also looks good to me.

Are we on the same page that this PR itself can be merged in, while we will continue working on the username password flow and device code flow issues separately?

bgavrilMS: note: broker is a public client application concept only

jiasli (approved):

PR #626, opened 2023-11-18
Remove newlines from description / Fix summary.

micwoj92:
When trying to build wheel it gives warning
```
writing msal.egg-info/PKG-INFO
/usr/lib/python3.11/site-packages/setuptools/dist.py:173: SetuptoolsDeprecationWarning: Invalid config.
!!

        ********************************************************************************
        newlines are not allowed in `summary` and will break in the future
        ********************************************************************************

!!
  write_field('Summary', single_line(summary))
```

This can be seen in latest wheel on pypi `msal-1.25.0-py2.py3-none-any.whl`
summary field in METADATA file is incomplete, it only states:
`Summary: The Microsoft Authentication Library (MSAL) for Python library`

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -6,12 +6,7 @@ universal=1
 [metadata]
 name = msal
 version = attr: msal.__version__
-description =
-    The Microsoft Authentication Library (MSAL) for Python library
-    enables your app to access the Microsoft Cloud
-    by supporting authentication of users with
-    Microsoft Azure Active Directory accounts (AAD) and Microsoft Accounts (MSA)
-    using industry standard OAuth2 and OpenID Connect.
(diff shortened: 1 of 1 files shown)


rayluo:
Thanks for bringing this to our attention.

Out of curiosity, why would you need to build a wheel (thus run into this issue) in the first place? I believe a standard `pip install msal` will already install [a wheel that we already build for you](https://pypi.org/project/msal/#files).

micwoj92:
The issue (description being cut after first newline) is also present in the wheel.

rayluo:
> The issue (description being cut after first newline) is also present in the wheel.

True, and your finding was legit.

I was just curios to know that in what scenario you would need to build a wheel from source. Most people just pip install an already-built wheel, and would not even notice that issue.

Regardless, this PR will be kept open for one more day, in case you would like to reply back. We will merge this PR afterwards.

rayluo (approved):

micwoj92:
> I was just curios to know that in what scenario you would need to build a wheel from source. Most people just pip install an already-built wheel, and would not even notice that issue.

This module is available in AUR. The package builds python-msal from source.

PR #644, opened 2024-01-03
Order scopes on save, and optimize the happy path for access token read

rayluo:
This PR currently contains 2 commits.

1. The first commit orders scopes on save, which is similar to https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4479. Existing unit tests need adjustment to match the changed scope order.
2. The second commit optimizes the happy path of access token read, which was the idea mentioned in [this conversation](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4479#discussion_r1437466906). Basically, when searching for an access token, this PR uses the token key to attempt an O(1) search, and falls back to O(n) search when necessary.
   * No other unit test adjustment is needed, meaning this change is backward compatible.
   * Existing benchmark test shows ~40% speed improvement on "token cache hit" code path, which is also visible in diagram No. 2 and No. 4 in [this benchmark page](https://azuread.github.io/microsoft-authentication-library-for-python/dev/bench/). (Note that the performance gain can be much higher than ~40% if each tenant has more tokens; it is actually an `O(1)` vs `O(n)` improvement.)

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -1357,13 +1357,14 @@ class ClientApplication(object):
             key_id = kwargs.get("data", {}).get("key_id")
             if key_id:  # Some token types (SSH-certs, POP) are bound to a key
                 query["key_id"] = key_id
-            matches = self.token_cache.find(
-                self.token_cache.CredentialType.ACCESS_TOKEN,
-                target=scopes,
-                query=query)
             now = time.time()
             refresh_reason = msal.telemetry.AT_ABSENT
-            for entry in matches:
+            for entry in self.token_cache._find(  # It returns a generator
+                self.token_cache.CredentialType.ACCESS_TOKEN,
+                target=scopes,
+                query=query,
+            ):  # Note that _find() holds a lock during this for loop;
+                # that is fine because this loop is fast
                 expires_in = int(entry["expires_on"]) - now
                 if expires_in < 5*60:  # Then consider it expired
                     refresh_reason = msal.telemetry.AT_EXPIRED
--- a/msal/application.py
+++ b/msal/application.py
@@ -1493,10 +1493,8 @@ class ClientApplication(object):
             **kwargs) or last_resp
 
     def _get_app_metadata(self, environment):
-        apps = self.token_cache.find(  # Use find(), rather than token_cache.get(...)
-            TokenCache.CredentialType.APP_METADATA, query={
-                "environment": environment, "client_id": self.client_id})
-        return apps[0] if apps else {}
+        return self.token_cache._get_app_metadata(
+            environment=environment, client_id=self.client_id, default={})
 
     def _acquire_token_silent_by_finding_specific_refresh_token(
             self, authority, scopes, query,
--- a/msal/token_cache.py
+++ b/msal/token_cache.py
@@ -104,6 +104,15 @@ class TokenCache(object):
                 ),
             default=default)
 
+    def _get_app_metadata(self, environment, client_id, default=None):  # O(1)
+        return self._get(
+            self.CredentialType.APP_METADATA,
+            self.key_makers[TokenCache.CredentialType.APP_METADATA](
+                environment=environment,
+                client_id=client_id,
+                ),
+            default=default)
+
     def _get(self, credential_type, key, default=None):  # O(1)
         with self._lock:
             return self._cache.get(credential_type, {}).get(key, default)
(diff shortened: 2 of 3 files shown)


msal/token_cache.py
```diff
@@ -88,20 +88,55 @@ def __init__(self):
...
+            return self._cache.get(credential_type, {}).get(key, default)
+
+    def _find(self, credential_type, target=None, query=None):  # O(n) generator
+        target = sorted(target or [])  # Match the order sorted by add()
         assert isinstance(target, list), "Invalid parameter type"
+
+        preferred_result = None
+        if (credential_type == self.CredentialType.ACCESS_TOKEN
+            and "home_account_id" in query and "environment" in query
+            and "client_id" in query and "realm" in query and target
```
bgavrilMS, line 118: realm could be null / empty for ADFS and other OIDC authorities. Not sure if you just add `query=""` in the dictionary or not, presumably you do and this comment is no-op.

bgavrilMS (approved):

msal/token_cache.py
```diff
@@ -88,20 +88,55 @@ def __init__(self):
...
+        with self._lock:
+            return self._cache.get(credential_type, {}).get(key, default)
+
+    def _find(self, credential_type, target=None, query=None):  # O(n) generator
+        target = sorted(target or [])  # Match the order sorted by add()
         assert isinstance(target, list), "Invalid parameter type"
+
+        preferred_result = None
+        if (credential_type == self.CredentialType.ACCESS_TOKEN
+            and "home_account_id" in query and "environment" in query
```
bgavrilMS, line 117: I'm guessing that you only try to use the "preferred_cache" env here?

msal/application.py
```diff
@@ -1357,13 +1357,14 @@ def _acquire_token_silent_from_cache_and_possibly_refresh_it(
...
             if key_id:  # Some token types (SSH-certs, POP) are bound to a key
                 query["key_id"] = key_id
-            matches = self.token_cache.find(
-                self.token_cache.CredentialType.ACCESS_TOKEN,
-                target=scopes,
-                query=query)
             now = time.time()
             refresh_reason = msal.telemetry.AT_ABSENT
-            for entry in matches:
+            for entry in self.token_cache._find(  # It returns a generator
```
jiasli, line 1362: Changing `token_cache.find` to `token_cache._find` makes this line no longer call the `find` method on `msal.token_cache.TokenCache`'s subclasses, e.g. `msal_extensions.token_cache.PersistedTokenCache.find`, but directly `msal.token_cache.TokenCache._find`.

This causes all logic in `msal_extensions.token_cache.PersistedTokenCache.find` being skipped:

https://github.com/AzureAD/microsoft-authentication-extensions-for-python/blob/a10d0929b53d5397a3d5883e97b6925de0252e2b/msal_extensions/token_cache.py#L72-L87

This method contains 2 important functionalities:

1. Load token cache from a file
2. Handle concurrent situation

When a `PersistedTokenCache` is provided to `ConfidentialClientApplication`, as token cache is no longer loaded from a file, `self._cache` becomes an empty dict when this line is called:

https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/84bdfabfc3cbd73bca485b2420fcb7ccc01191d0/msal/token_cache.py#L118

resulting in the token cache being bypassed.

`PublicClientApplication` is not affected, as `application.ClientApplication._find_msal_accounts` still uses the old `find`.

https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/52b1fc5a442ff5dd33f48ce717f1032c8002ea9e/msal/application.py#L1111

rayluo: This will be mitigated by MSAL EX https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/128

Moazzem-Hossain-pixel (commented):
If you can take initiative to resolve that will be highly appreciated.

rayluo:
> If you can take initiative to resolve that will be highly appreciated.

@Moazzem-Hossain-pixel , you are commenting on a topic that has already been implemented. There are some follow-ups in their own github issues linked above, and we are addressing them respectively. But the main change in the current topic is already stable. And our downstream partner - Azure CLI - has been unblocked by https://github.com/AzureAD/microsoft-authentication-extensions-for-python/issues/127

If you are blocked by any specific issue, please add your comment into those linked issues, or even create a new issue to elaborate. Meanwhile, I'm locking this issue here.

PR #651, opened 2024-01-18
Attempts account removal from broker first

rayluo:
This change is inspired by [this conversation](https://github.com/Azure/azure-cli/issues/20231#issuecomment-1887956233).

CC: @jiasli

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -1123,12 +1123,13 @@ class ClientApplication(object):
 
     def remove_account(self, account):
         """Sign me out and forget me from token cache"""
-        self._forget_me(account)
         if self._enable_broker:
             from .broker import _signout_silently
             error = _signout_silently(self.client_id, account["local_account_id"])
             if error:
                 logger.debug("_signout_silently() returns error: %s", error)
+        # Broker sign-out has been attempted, even if the _forget_me() below throws.
+        self._forget_me(account)
 
     def _sign_out(self, home_account):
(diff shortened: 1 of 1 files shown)


ashok672 (commented):
what if broker sign out also throws an unexpected exception? Is this possible. I don't understand why self.forget_me would throw an error skipping execution below.

rayluo:
> what if broker sign out also throws an unexpected exception? Is this possible.

It is always possible for any code path to potentially throw an exception. And we can let those really unexpected exceptions bubble up.

> I don't understand why self.forget_me would throw an error skipping execution below.

In that referenced Azure CLI's case, they persist the MSAL Python's token cache on disk, and the `forget_me()` operation would attempt removing some tokens thus modifying that token cache file. Again, any code path could throw exception, in this case, it can be disk I/O error.

ashok672 (approved):

jiasli:
Thanks for the attempt to solve https://github.com/Azure/azure-cli/issues/20231#issuecomment-1887171392.

However, I don't think this change makes any difference as `OSError: [WinError -2146893813]` is thrown when `get_accounts()` is called. We won't even be able to retrieve the account that can be passed to `remove_account()`, because the account information is also in `msal_token_cache.bin`.

In other word, if `msal_token_cache.bin` can't be accessed, we totally lose track of what accounts are in WAM.

jiasli (changes requested):
Let's rethink about it.

Issue #31, opened 2019-04-07
Feature Request - IWA pass through capabilities - align with .net


danield137:
I am one of the devs working on https://github.com/Azure/azure-kusto-python.
We are currently using *adal* for azure authentication, and we get a lot of requests for silent authentication flow (as we provide in our .net SDK).

Currently, adal python does not expose such a capability, all though it seems possible to achieve using similar low level, OS dependent code.

We would love the ability to provide silent auth for our windows users, and, to take a part of the dev effort, if needed.

Pratik-Gautam:
@rayluo  & @danield137 , as I understand, silent authentication is not available in ADAL for Python yet . So is there a way through which we can still achieve SSO using ADAL for Python?

rayluo:
To be precise, it is true that the Integrated Windows Auth (IWA) - i.e. current Windows login user can sign in without typing password - is not currently available in either ADAL Python and MSAL Python. You can upvote this feature request. As an open source library, we are open to PR contributions. CC: @danield137 ;-)

On a side note, the lastest version of MSAL .Net and MSAL Python are already using same format of token cache. So if you happen to have an app powered by MSAL .Net and its another version powered by MSAL Python, and they share access to same physical token cache, one app can pick up the other app's token, therefore sign in silently.

In a more generic case, though, you can consider an alternative, the [Device Flow](https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/0.3.1/sample/device_flow_sample.py). In such flow, the actual sign-in happens on end user's another device, typically their desktop or mobile, which they already signed in. It is not necessarily more convenient because the end user would still need to input a short device code in such flow, but it is probably more secure than username-password-flow, because the end user does not need to type their password in your app (and worrying whether your app would persist their password - and you should not do that!).

If you have follow-up questions on how to use Device Flow, please create another issue for that topic. Thanks!

jorgst:
This feature would be awesome, how many upvotes does it need before you consider implementing it?

rayluo:
[...]

PR #652, opened 2024-01-22
feat: add integrated windows authentication support under public clients

shajia-deshaw:


rayluo:
Thanks, @shajia-deshaw . Is there any way to test this? Can they be written into test cases?
Anyway, I'll leave this to @ashok672 for reviewing.

shajia-deshaw:
Sure @rayluo  - I can add the test cases shortly (I had this tested with a driver script internally)
```bash
app = PublicClientApplication(
    "<cleint_id>",
    authority="https://login.microsoftonline.com/<tenant_id>")
token = app.acquire_token_integrated_windows_auth("<upn>")
print(token)
```

msal/application.py
```diff
@@ -172,6 +172,7 @@ class ClientApplication(object):
     ACQUIRE_TOKEN_FOR_CLIENT_ID = "730"
     ACQUIRE_TOKEN_BY_AUTHORIZATION_CODE_ID = "832"
     ACQUIRE_TOKEN_INTERACTIVE = "169"
+    ACQUIRE_TOKEN_INTEGRATED_WINDOWS_AUTH_ID = "870"
```
shajia-deshaw, line 175: I'm not sure how this number is mapped. I just added a random number for IWA btw.

msal/token_cache.py
```diff
@@ -243,7 +243,7 @@ def __add(self, event, now=None):
                     at["refresh_on"] = str(now + refresh_in)  # Schema wants a string
                 self.modify(self.CredentialType.ACCESS_TOKEN, at, at)
 
-            if client_info and not event.get("skip_account_creation"):
+            if (client_info or iwa) and not event.get("skip_account_creation"):
```
shajia-deshaw, line 246: The access token returned by IWA did not meet the criteria to be added as an `ACCOUNT` credential type in the `TokenCache` i.e. it didn't have `client_info` in the response. It was being added as a `ACCESS_TOKEN` credential type. The cache test in the e2e tests were failing because of it since the `get_accounts()` call didn't return anything.  Hence, I added a special case of `iwa` to be added to the `ACCOUNT` credential type.

shajia-deshaw:
Hello @ashok672 - Checking if you had a chance to review the PR? Thanks.

aschafs:
I just wanted to check if this PR was going to be reviewed and merged. We're very interested in getting IWA support in the msal python library. Thanks!

bgavrilMS:
I don't think so, we are discussing to deprecate Integrated Windows Auth from all MSALs in favor of WAM (broker use). Integrated Windows Auth relies on SAML and HTTP level contracts which are easily broken by network changes, and this causes a lot of churn and support issues. WAM uses a different protocol (PRT) which is not affected. 

@iulico-1 @localden - can you provide a perspective as scenario owners?

rayluo:
> I don't think so, we are discussing to deprecate Integrated Windows Auth from all MSALs in favor of WAM (broker use). Integrated Windows Auth relies on SAML and HTTP level contracts which are easily broken by network changes, and this causes a lot of churn and support issues. WAM uses a different protocol (PRT) which is not affected.
> 
> @iulico-1 @localden - can you provide a perspective as scenario owners?

For the sake of completeness, we had a similar conversation in the feature request issue with the contributor, who mentioned that [IWA also works on Linux](https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/31#issuecomment-1754454080) and that was part of the reason that they worked out this PR. That is a scenario that not currently satisfied by broker.

shajia-deshaw:
> > I don't think so, we are discussing to deprecate Integrated Windows Auth from all MSALs in favor of WAM (broker use). Integrated Windows Auth relies on SAML and HTTP level contracts which are easily broken by network changes, and this causes a lot of churn and support issues. WAM uses a different protocol (PRT) which is not affected.
> > @iulico-1 @localden - can you provide a perspective as scenario owners?
> 
> For the sake of completeness, we had a similar conversation in the feature request issue with the contributor, who mentioned that [IWA also works on Linux](https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/31#issuecomment-1754454080) and that was part of the reason that they worked out this PR. That is a scenario that not currently satisfied by broker.

Thanks @rayluo for adding context! 

FTR, we've been using the patched version of MSAL with Kerberos / IWA flow for months in our internal linux services and applications -- Things have been working just fine. 

However, we believe this could be a general enhancement in the upstream MSAL library, benefiting similar users, as evident from the comments on the issue/PR.

localden:
Adding my perspective here - on Windows and macOS, we are indeed shifting to authentication brokers; however, that leaves Linux scenarios in the open and we will need to continue falling back on the vanilla API. I am not inherently opposed to having a native IWA implementation in MSAL Python, but we'd need to coordinate how it falls back to broker and when.

@ashok672 @iulico-1 - something for us to discuss.

tests/test_e2e.py
```diff
@@ -223,6 +223,29 @@ def _test_username_password(self,
...
+                  http_client=None,
+                  **ignored):
+        assert authority and client_id and username and scope
+        self.app = self._build_app(
+            client_id, authority=authority,
+            http_client=requests,
+            azure_region=azure_region, client_credential=client_secret)
+        self.assertEqual(
+            self.app.get_accounts(username=username), [], "Cache starts empty")
+        result = self.app.acquire_token_integrated_windows_auth(
```
iulico-1, line 246: >acquire_token_integrated_windows_auth [](http://example.com/codeflow?start=26&length=37)

I am not convinced we need to introduce Windows integrated Auth flow as a separate API.

Windows Integrated Auth is conceptually the same as functionality as default account that brokers (WAM) expose.

Can we follow the same pattern we did with OS account on .net ?

localden: @iulico-1 - how is this going to work on Linux, where there is no WAM? We can't expect every customer to enroll their device either.

alextok: On linux there is no WIA also :) this api will just fail with UIRequired on linux without broker integration.

rayluo: > On linux there is no WIA also :) this api will just fail with UIRequired on linux without broker integration.

But the customer [claimed otherwise](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/652#issuecomment-2427190084).

iulico-1: A form of WIA can be setup on Linux.

rayluo: So, am I understanding correctly that we now agree on there is a need for Integrated Windows Auth (IWA, a.k.a. WIA) for Linux? And then this conversation is back to the original question, i.e. how the IWA shall be exposed in an API? Brainstorming time. :-)

1. This PR's current implementation `acquire_token_integrated_windows_auth(username, scopes)` was probably inspired by [MSAL .Net's dedicated IWA api](https://learn.microsoft.com/en-us/entra/msal/dotnet/acquiring-tokens/desktop-mobile/integrated-windows-authentication).
  If we do not want IWA to be a dedicated API, we will need to fit it into other existing APIs. Such as:
2. Modeling it as a ROPC but without a password `acquire_token_by_username_password(username, password=None, scopes=[...])`. This makes sense from the implementation standpoint, because the actual implementation already shares many building blocks (mex.py, wstrust_request.py) which are only used in ROPC code path.
3. Modeling it as "the same pattern we did with OS account on ~.net~ Python", and that is `acquire_token_interactive(scopes, login_hint="johndoe@example.com", prompt="none")`

Which one do we prefer?

localden: @rayluo @iulico-1 I think this fits best in the `acquire_token_interactive` flow, and the library can then determine that it's IWA based on parameter heuristics. This will also help reduce the API surfaces that customers have to navigate to use the capability, _and_ simplify broker intergration.

shajia-deshaw: Hello, is there anything that I can help with here? (Please feel free to leave any reviews)

jbw-vtl:
Hi All,

We are also very interested in these integrations.
Are there any updates on getting this merged eventually?

shajia-deshaw:
Hello @localden - any chance of reviving this PR?

PR #655, opened 2024-01-24
Rebrand from AAD to Microsoft Entra

SHERMANOUKO:


rayluo:
So far, so good. Are these all the changes needed, or are you still working on more? In the latter case, you can mark this PR as a draft until it is ready for review.

What is the exact term, anyway? Is it ["AAD" = "Microsoft Entra **ID**"](https://learn.microsoft.com/en-us/entra/identity/) (i.e., is the **ID** keyword important)? BTW, do "B2C" and "CIAM" get new names, too?

SHERMANOUKO:
These are the only changes for this PR

Azure AD B2C retained its name
CIAM + B2B = Microsoft Entra External ID as at now though this may change.

When it comes to the branding, it depends. In some cases Azure AD == Microsoft Entra ID. In some cases it's Microsoft Entra. Microsoft Entra is the name for the product family of identity and network access solutions. Microsoft Entra ID is one of the products within that family. For example, roles, users, tokens, groups etc belong to the product Microsoft Entra so

Azure AD users == Microsoft Entra users and not Microsoft Entra ID users.

SHERMANOUKO:
Good catch.

I need to update that too. I'll convert to draft and update that first.

SHERMANOUKO:
@rayluo this PR is now ready for review

rayluo (approved):

PR #667, opened 2024-02-20
Don't use bare except when importing

Singletoned:
Using a bare except statement when importing hides other errors, which then get lost when the next import fails.

### DIFF
--- a/msal/mex.py
+++ b/msal/mex.py
@@ -27,7 +27,7 @@
 
 try:
     from urllib.parse import urlparse
-except:
+except ImportError:
     from urlparse import urlparse
 try:
     from xml.etree import cElementTree as ET


Singletoned:
I don't actually know what caused the non-`ImportError`.  It was a one off occurrence in our logs, and the true error was disguised by the bare except, hence the PR.  I'd be grateful if this could be fixed, so that if it happens again, we'll have the correct traceback in our logs.

I strongly recommend never using a bare `try...except...` at all, and in fact most linters will automatically pick this up and mark it as an error.

Note that you are using a better pattern just a few lines below the change in this PR.

A better pattern overall is to not use `try` at all, and to explicitly test for a version of Python and modify the behaviour according to that (which is what you are doing in other places in the code).

rayluo (approved):
Thanks for the report.

If you would also like to add more commits into this PR to change other bare except, we will happily take it. Or we can also accept this PR as-is.

PR #673, opened 2024-03-01
Update the default broker redirect URI to be a valid URI

iulico-1:
MSAL.cpp fixed the URI parsing logic to not allow invalid URIs which causes the default redirect URI to not be properly set when calling MSAL.runtime. This caused the sev2 ICM with azcli.

Chaning the default to be the default RU for native clients: https://login.microsoftonline.com/common/oauth2/nativeclient

### DIFF
--- a/msal/broker.py
+++ b/msal/broker.py
@@ -169,8 +169,9 @@ def _signin_interactively(
         **kwargs):
     params = pymsalruntime.MSALRuntimeAuthParameters(client_id, authority)
     params.set_requested_scopes(scopes)
-    params.set_redirect_uri("placeholder")  # pymsalruntime 0.1 requires non-empty str,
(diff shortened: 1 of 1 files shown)


bgavrilMS (approved):

rayluo (approved):
Note: Later when we need to merge in the `mac` branch, we will do it this way.

```python
if running_on_mac:
    uri = "msauth.com.msauth.unsignedapp://auth"
else:
    uri = "https://login.microsoftonline.com/common/oauth2/nativeclient"  # Came from https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-auth-code-flow#request-an-authorization-code
```

And then MSALRuntime will also decide whether to honor the MSAL.py request based on the current platform.

PR #678, opened 2024-03-07
Implements a new optional oidc_authority parameter

rayluo:


msal/application.py
```diff
@@ -506,11 +516,12 @@ def __init__(
                 self.http_client,
                 validate_authority=validate_authority,
                 instance_discovery=self._instance_discovery,
+                oidc_authority_url=oidc_authority,
                 )
         except ValueError:  # Those are explicit authority validation errors
             raise
         except Exception:  # The rest are typically connection errors
-            if validate_authority and azure_region:
+            if validate_authority and azure_region and not oidc_authority:
```
bgavrilMS, line 524: I am not sure I fully understand this exception block, but am I right to understand that instance discovery will not be performed for oidc_authority?

msal/authority.py
```diff
@@ -65,12 +66,46 @@ def __init__(
...
+        logger.debug("openid_config = %s", openid_config)
+        self.authorization_endpoint = openid_config['authorization_endpoint']
+        self.token_endpoint = openid_config['token_endpoint']
+        self.device_authorization_endpoint = openid_config.get('device_authorization_endpoint')
+        _, _, self.tenant = canonicalize(self.token_endpoint)  # Usually a GUID
+
+    def _initialize_oidc_authority(self, oidc_authority_url):
+        authority, self.instance, tenant = canonicalize(oidc_authority_url)
+        self.is_adfs = tenant.lower() == 'adfs'  # As a convention
+        self._is_b2c = True  # Not exactly true, but
```
bgavrilMS, line 96: seems iffy :). Maybe rename the flag ?

msal/application.py
bgavrilMS: Never mind, I saw the tests.

bgavrilMS (approved):

msal/authority.py
rayluo: Yeah, I thought about it. There are two approaches here.

1. Implement a cleanest `class OidcAuthority` which contains no `is_adfs`, `_is_b2c`, `_is_known_to_developer` at all. But then it would not be a drop-in replacement for the old `class Authority` (which becomes actually an EntraAuthority then). For a dozen places in `application.py` that currently reference the authority object, they would all need one extra condition in their already complicated `if` clause, such as:

```python
if isinstance(authority, OidcAuthority):
    do_oidc_thing()
else:
    assert isinstance(authority, Authority)  # It is the Entra authority
    if authority.is_adfs or authority._is_b2c:
        do_adfs_or_b2c_thing()  # Basically bypass broker code path
    elif authority._is_known_to_developer:
        skip_instance_discovery()
```

2. Bite the bullet, set `_is_b2c` and `_is_known_to_developer` as True, so that most of the existing logic in `application.py` needs no change.

After some brief attempt on `#1`, this PR is now implemented with `#2` approach. Last but not least, this is an implementation detail, so, it is a 2-way decision which we can always revisit in the future.

bgavrilMS:
@rayluo - can you pls merge this and include it in a release?

rayluo:
> @rayluo - can you pls merge this and include it in a release?

Sure, @bgavrilMS . Actually, since Wednesday, we already informed Azure SDK team (CC @xiangyan99) to start their acceptance test, as part of our release routine. The result came back just now, it is clean.

We will proceed with MSAL 1.28 release now.

PR #680, opened 2024-03-21
Friendly hint for a typical oidc_authority error

rayluo:
In our own web app testing, we sometimes forgot to append the `/v2.0` suffix to a CIAM CUD `oidc_authority`, and ended up with a cryptic error message, "AADSTS500207: The account type can't be used for the resource you're trying to access". This may become an FAQ and a frequent source of customer support requests.

In this PR, we tentatively add a hint into the error message.

> 'Did you forget to append "/v2.0" to your oidc_authority? '

so that a full error page in a web app may look like this:

> **Login Failure**
>
>invalid_request
>    Did you forget to append "/v2.0" to your oidc_authority? AADSTS500207: The account type can't be used for the resource you're trying to access. Trace ID: e4568f2b-f5b3-4e5e-b766-e7689b180000 Correlation ID: 765569d0-7583-45ec-93f1-69d6095164a4 Timestamp: 2024-03-21 03:17:17Z

Note:

* Currently, this treatment is added into the most flows, including CCA's auth code flow, Client Credential flow, Username Password flow (ROPC), and PCA's Device Code flow.
  - Note that the ROPC flow yields a different error number.

* PCA's interactive flow has two places to return the error: the API response, and the error rendered in browser. In this PR, we added the treatment for the former because it was easy (just a one-liner). The latter would also be a good candidate to receive this treatment, however it is not implemented in this PR due to its complexity.

* The so-called "dev samples" inside the MSAL repo are also updated this time. They can become the blueprint when we update the AzureSamples later.

* This PR also contains a change to skip user realm discovery for oidc authority.


msal/application.py
```diff
@@ -973,6 +974,11 @@ def authorize():  # A controller in a web app
             **kwargs))
         if "access_token" in response:
             response[self._TOKEN_SOURCE] = self._TOKEN_SOURCE_IDP
+        if ("AADSTS500207" in response.get("error_description", "") and
+                self._oidc_authority and not self._oidc_authority.endswith("/v2.0")):
+            response["error_description"] = (
+                'Did you forget to append "/v2.0" to your oidc_authority? '
```
SammyO: I feel like the best way that this error should be handled/prevented is for the v2.0 endpoint not even existing for CIAM; which is being worked on by the CIAM team. You could choose to keep this handling here, but just know that in the near future developers should never be getting this error anymore as being caused by forgetting to add v2.0 to the URL path.

bgavrilMS: @SammyO - does the CIAM team consider making this change for CIAM CUD GA? Because once we go GA, it's all over - you'll have to support "v2.0" suffix (maybe through DNS hackery), otherwise you'll break production apps.

SammyO: @bgavrilMS yes, exactly because of the reason you've outlined there.

bgavrilMS (dismissed):

msal/application.py
bgavrilMS: @rayluo - pls open a similar issue to MSAL.NET

rayluo: > the v2.0 endpoint not even existing for CIAM; which is being worked on by the CIAM team

Does that mean the OIDC authority for CUD after its GA will be simply `https://contoso.com/tenant` i.e. without the `/v2.0` suffix? That would essentially mean the V2 URL converges into the V1 URL. We had this proposal before, but it was rejected because it is believed that V1 and V2 URLs need to coexist. Is that no longer the case? @SammyO

SammyO: @rayluo what I think will be happening is a v1.0 version of an authority endpoint returning something like a 404. I don't think there will be a converging (as this would hide logic from developers and make things less explicit, which IMO is not desirable). But let me confirm.

PR #687, opened 2024-04-10
ConfidentialClientApplication federated by a Managed Identity

rayluo:
This is a proof-of-concept. It provides a high-level API which allows your confidential client to federate with a managed identity.

```python
import msal
cca = msal.ConfidentialClientApplication(
    "my_client_id",
    client_credential=msal.SystemAssignedManagedIdentity(),  # Or it can be an msal.UserAssignedManagedIdentity(client_id="guid")
    ...)
result = cca.acquire_token_for_client(scopes["scope1", "scope2"])  # It uses scopes
```

The high-level API also supports some variations. See the last purple box of [the `client_credential` parameter's document](https://msal-python.readthedocs.io/en/fic-by-mi/#msal.ClientApplication.params.client_credential) for more details.

See how it is simpler than the low-level API in the [coming-soon Managed Identity implementation](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/480).

Installation: `pip install --force-reinstall "git+https://github.com/AzureAD/microsoft-authentication-library-for-python.git@fic-by-mi"`

[Internal work item is tracked here](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3289718).

**TODO**: After merging this PR, we will need to somehow update [this internal doc](https://eng.ms/docs/microsoft-security/identity/app-plat-and-graph/app-vertical/aad-first-party-apps/identity-platform-and-access-management/microsoft-identity-platform/federated-identity-credentials?tabs=python).


msal/application.py
```diff
@@ -230,32 +232,80 @@ def __init__(
...
+                Once you setup the federation, the following declarative API
+                takes care of the managed identity token acquisition for you.
+                You just need to assign ``client_credential``
+                with an instance of :py:class:`msal.SystemAssignedManagedIdentity`
+                or :py:class:`msal.UserAssignedManagedIdentity`, for example::
+
+                    app = msal.ConfidentialClientApplication(
+                        "my_client_id",
+                        client_credential=msal.SystemAssignedManagedIdentity(),
+                        ...)
```
jiasli, line 298: My understanding is that **federated identity credential** can only be configured on **user-assigned managed identity**, not **system-assigned managed identity**, according to

https://learn.microsoft.com/en-us/entra/workload-id/workload-identity-federation

> You use workload identity federation to configure a [user-assigned managed identity](https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/how-manage-user-assigned-managed-identities) or [app registration](https://learn.microsoft.com/en-us/entra/identity-platform/app-objects-and-service-principals) in Microsoft Entra ID to trust tokens from an external identity provider (IdP), such as GitHub or Google.

https://learn.microsoft.com/en-us/entra/workload-id/workload-identity-federation-create-trust-user-assigned-managed-identity

> Configuring a federated identity credential on a system-assigned managed identity isn't supported.

msal/application.py
```diff
@@ -650,13 +700,27 @@ def _build_client(self, client_credential, authority, skip_regional_client=False
...
-        if isinstance(client_credential, dict):
+        if ManagedIdentity.is_managed_identity(client_credential):
+            # Federated Identity Credential (FIC), a.k.a. workload identity
+            # https://learn.microsoft.com/entra/workload-id/workload-identity-federation
+            client_assertion_type = Client.CLIENT_ASSERTION_TYPE_JWT
+            client_assertion = AutoRefresher(
+                lambda: ManagedIdentityClient(
+                    client_credential, self.http_client,
+                    ).acquire_token_for_client(
+                        resource="api://AzureADTokenExchange",
```
jiasli: Does this `acquire_token_for_client` call generate an OIDC token?

What is the point of making an app trust managed identity when the managed identity can be used as an identity/service principal by itself?

msal/application.py
jiasli: I see. It looks like you are creating federation between an app and a user-assigned managed identity.

rayluo: > I see. It looks like you are creating federation between an app and a user-assigned managed identity.

Yes. As described in the title of this PR, it is about a Confidential Client Application (CCA) federated by a Managed Identity (MI). It is NOT about letting a (User-assigned) MI federated by other IdP.

In particular, I tried a CCA federated by a system-assigned MI, and it worked.

msal/application.py
rayluo: > Does this `acquire_token_for_client` call generate an OIDC token?

Azure CLI's "OIDC token" is a vague term that I have a hard time remembering what it really is. :-)

In the current CCA-federated-by-MI case, the underlying federation token is conceptually an access token for a specific audience.

> What is the point of making an app trust managed identity when the managed identity can be used as an identity/service principal by itself?

My understanding is that there are several advantages for sticking with CCA as your app's identity.

1. MI has operational limitations. If you deploy your popular app into a fleet of VMs, each of those VMs has a different system-assigned MI. Not sure whether it is possible to customize those MI's branding. Of course, you could configure those VMs to share one user-assigned MI, but then it hits the next limitation.
2. MI has functional limitations. It supports only service-to-service call. There is no plan to add Auth Code flow (i.e. web app sign-in) or OBO to it.
3. A CCA federated by a system-assigned or user-assigned MI gives you the best of both worlds. You have full capabilities of a CCA, yet you do not need to manage a client secret or a cert.

jiasli:
There is an internal work item of migrating to **managed identity + multitenant application** (MI+CCA). Below is some sample code:

- Python: https://gist.github.com/amritpal-singh-98/92c5b7627e7ea0dc1491e03b7b834310
- .NET: https://gist.github.com/AtOMiCNebula/988383241cd1837a6c3faf6a353cecea

We may want to continue working on this PR.

mrpowerus:
Please do continue working on this PR. This functionality would be great.

PR #693, opened 2024-04-30
Promote TokenCache._find() to TokenCache.search()

rayluo:
Hi @jiasli , have you started testing this feature branch yet? Do you need it to be shipped by end of month?

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -1142,7 +1142,7 @@ class ClientApplication(object):
                     "local_account_id": a.get("local_account_id"),  # Tenant-specific
                     "realm": a.get("realm"),  # Tenant-specific
                 }
-            for a in self.token_cache.find(
+            for a in self.token_cache.search(
                 TokenCache.CredentialType.ACCOUNT,
                 query={"environment": environment})
             if a["authority_type"] in interested_authority_types
@@ -1188,18 +1188,22 @@ class ClientApplication(object):
             "home_account_id": home_account["home_account_id"],}  # realm-independent
         app_metadata = self._get_app_metadata(home_account["environment"])
         # Remove RTs/FRTs, and they are realm-independent
-        for rt in [rt for rt in self.token_cache.find(
+        for rt in [  # Remove RTs from a static list (rather than from a dynamic generator),
+                    # to avoid changing self.token_cache while it is being iterated
+            rt for rt in self.token_cache.search(
                 TokenCache.CredentialType.REFRESH_TOKEN, query=owned_by_home_account)
                 # Do RT's app ownership check as a precaution, in case family apps
                 # and 3rd-party apps share same token cache, although they should not.
                 if rt["client_id"] == self.client_id or (
                     app_metadata.get("family_id")  # Now let's settle family business
                     and rt.get("family_id") == app_metadata["family_id"])
-                ]:
+        ]:
             self.token_cache.remove_rt(rt)
-        for at in self.token_cache.find(  # Remove ATs
-                # Regardless of realm, b/c we've removed realm-independent RTs anyway
-                TokenCache.CredentialType.ACCESS_TOKEN, query=owned_by_home_account):
+        for at in list(self.token_cache.search(  # Remove ATs from a static list,
+                # to avoid changing self.token_cache while it is being iterated
+            TokenCache.CredentialType.ACCESS_TOKEN, query=owned_by_home_account,
+            # Regardless of realm, b/c we've removed realm-independent RTs anyway
+        )):
             # To avoid the complexity of locating sibling family app's AT,
             # we skip AT's app ownership check.
             # It means ATs for other apps will also be removed, it is OK because:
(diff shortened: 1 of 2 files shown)


msal/token_cache.py
```diff
@@ -150,8 +151,9 @@ def _find(self, credential_type, target=None, query=None):  # O(n) generator
                     if entry != preferred_result:  # Avoid yielding the same entry twice
                         yield entry
 
-    def find(self, credential_type, target=None, query=None):  # Obsolete. Use _find() instead.
-        return list(self._find(credential_type, target=target, query=query))
+    def find(self, credential_type, target=None, query=None):
+        warnings.warn("Use list(search()) instead of find().", DeprecationWarning)
```
jiasli: If `msal.token_cache.TokenCache.find` is deprecated, MSAL itself should not use it anymore:

https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/48a7bd3c8e7040c02165e4e12f3a8004f18391a3/msal/application.py#L1111-L1113

jiasli:
I don't see a PR at https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pulls yet. I think we need both PRs so that we can test them together.

FYI: `TokenCache._find()` is introduced by #644

msal/token_cache.py
rayluo: Good catch! Github's reference feature somehow did not show me those old `find()` references.

In the new commit, ~most~ ALL of the `find()` have been replaced with `search()`. Note that in some circumstances, the caller would actually want a list, not a generator. But we still deprecate `find()` anyway, in favor of the explicit `list(search(...))` pattern.

Looks all good now, @jiasli ?

jiasli: Good!

msal/token_cache.py
```diff
@@ -150,8 +151,12 @@ def _find(self, credential_type, target=None, query=None):  # O(n) generator
                     if entry != preferred_result:  # Avoid yielding the same entry twice
                         yield entry
 
-    def find(self, credential_type, target=None, query=None):  # Obsolete. Use _find() instead.
-        return list(self._find(credential_type, target=target, query=query))
+    def find(self, credential_type, target=None, query=None):
+        """Equivalent to list(search(...))."""
+        warnings.warn(
+            "Use list(search(...) instead to explicitly get a list.",
```
jiasli:
```suggestion
            "Use list(search(...)) instead to explicitly get a list.",
```

msal/application.py
```diff
@@ -1163,9 +1163,10 @@ def _sign_out(self, home_account):
                     and rt.get("family_id") == app_metadata["family_id"])
                 ]:
             self.token_cache.remove_rt(rt)
-        for at in self.token_cache.find(  # Remove ATs
+        for at in list(self.token_cache.search(  # Remove ATs from a snapshot
```
jiasli: I understand the reason of using `list()` is to avoid changing `self.token_cache` while it is being iterated, but it will be good to add a more detailed comment, otherwise the reader may not know why `list()` is used.

msal/application.py
```diff
@@ -1376,11 +1379,11 @@ def _acquire_token_silent_from_cache_and_possibly_refresh_it(
                 query["key_id"] = key_id
             now = time.time()
             refresh_reason = msal.telemetry.AT_ABSENT
-            for entry in self.token_cache._find(  # It returns a generator
+            for entry in self.token_cache.search(  # We want a generator here for performance, so that we can break early in cache-hit
                 self.token_cache.CredentialType.ACCESS_TOKEN,
                 target=scopes,
                 query=query,
-            ):  # Note that _find() holds a lock during this for loop;
+            ):  # Note that search() holds a lock during this for loop;
```
jiasli, line 1386: `search()` uses a `RLock` which only prevents the search logic from being executed by multiple threads:

https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/c4152d2920f16c0416ebcfa1b88c34d862db56e3/msal/token_cache.py#L142-L151

However, it doesn't prevent deleting an element from the token cache:

```py
import threading

mylist = [1, 2, 3]
lock = threading.RLock()


def get_next():
    with lock:
        for item in mylist:
            yield item


for i in get_next():
    print(i)
    print(mylist)
    mylist.remove(i)
    print(mylist)
```

Output:

```
1
[1, 2, 3]
[2, 3]
3
[2, 3]
[2]
```

jiasli: `TokenCache` uses a `RLock`:

https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/c4152d2920f16c0416ebcfa1b88c34d862db56e3/msal/token_cache.py#L42

This only prevents the enclosed code from being executed by multiple threads:

https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/c4152d2920f16c0416ebcfa1b88c34d862db56e3/msal/token_cache.py#L142-L151

However, it doesn't prevent deleting an element from the token cache while `search`ing as `RLock` allows multiple `acquire` calls by the same thread.

msal/application.py
jiasli: Have we considered using a [`Lock`](https://docs.python.org/3/library/threading.html#lock-objects) to avoid such scenario?

jiasli (approved):

msal/application.py
rayluo: We discussed this offline. `Lock` is not a good choice here because:
1. Internally, TokenCache's methods may call each other and they need a `RLock` to coordinate with each other.
2. In our usage here, the default TokenCache is an in-memory `dict` which would automatically throw an error if we forgot to apply the `list(...)`.

```
  File "/home/demo/Live-usb-storage/repo/microsoft-authentication-library-for-python/msal/application.py", line 1209, in _sign_out
    for at in self.token_cache.search(  # Remove ATs from a static list,
  File "/home/demo/Live-usb-storage/repo/microsoft-authentication-library-for-python/msal/token_cache.py", line 147, in search
    for entry in self._cache.get(credential_type, {}).values():
RuntimeError: dictionary changed size during iteration
```

msal/application.py
rayluo: [Discussed here](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/693#discussion_r1621227729)

PR #712, opened 2024-06-17
Enable public client ROPC via broker

fengga:
If use_broker enabled, ROPC flows on windows will call into WAM, on macOS will call MSAL C++ logic.

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -1710,7 +1710,7 @@ class ClientApplication(object):
         """
         claims = _merge_claims_challenge_and_capabilities(
                 self._client_capabilities, claims_challenge)
-        if False:  # Disabled, for now. It was if self._enable_broker:
+        if self._enable_broker:
             from .broker import _signin_silently
             response = _signin_silently(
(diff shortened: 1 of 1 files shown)


bgavrilMS (approved):

rayluo:
Sorry, @fengga , my bad, I forgot to mention that, in general, all broker-relevant changes must be tested locally, rather than solely rely on the test pipeline, because we did not figure out a way to install broker into MSAL Python's current pipeline. (You may also tackle how to setup broker in test pipeline, but no need to do it in this PR.)

In this particular case, since your one-liner change was reverting a previous change, you could also have use [Github's blame feature](https://github.com/AzureAD/microsoft-authentication-library-for-python/blame/68a7cb342bef68f498dac3d97618f5d817a42986/msal/application.py#L1713) to find [the previous change's PR](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/569/files) and realize that there was a test case there. That test case is now failing because we changed the implementation now. Please come up with a new PR (can be in this same branch) to adjust that test case, too.

PR #714, opened 2024-06-24
Update ROPC broker related tests

fengga:
Add a new broker ROPC test, and fix a broken test involved by [#712](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/712)

All tests passed.
image

### DIFF
--- a/tests/broker-test.py
+++ b/tests/broker-test.py
@@ -46,6 +46,15 @@ def interactive_and_silent(scopes, auth_scheme, data, expected_token_type):
         )
     _assert(result, expected_token_type)
 
+def test_broker_username_password(scopes, expected_token_type):
+    print("Testing broker username password flows by using labs account: 'fidlab@msidlab4.com', please provide the password.")
+    username = "fidlab@msidlab4.com"
+    password = input("Password:")
+    result = pca.acquire_token_by_username_password(username, password, scopes)
+    _assert(result, expected_token_type)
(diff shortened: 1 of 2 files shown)


tests/broker-test.py
```diff
@@ -46,6 +52,17 @@ def interactive_and_silent(scopes, auth_scheme, data, expected_token_type):
         )
     _assert(result, expected_token_type)
 
+def test_broker_username_password(scopes, expected_token_type):
+    print("Testing broker username password flows by using accounts in local .env")
+    username = os.getenv("ENV_LAB4_ACCOUNT")
+    password = os.getenv("ENV_LAB4_ACCOUNT_PASSWORD")
+    assert(username!=None and len(username)>0)
+    assert(password!=None and len(password)>0)
```
rayluo: All content read by `os.getenv()` are already environment variables. They typically do not need to have a redundant "ENV_" prefix. :-)  Also, let's use generic names.

Consider adopting the following implementation. It would need you to also add an `import getpass` at the top of this file.

```suggestion
    username = os.getenv("BROKER_TEST_ACCOUNT") or input("Input test account for broker test: ")
    password = os.getenv("BROKER_TEST_ACCOUNT_PASSWORD") or getpass.getpass("Input test account's password: ")
    assert(username and password, "You need to provide a test account and its password")
```

rayluo (approved):

PR #718, opened 2024-07-08
Support SNI via PFX

rayluo:
1. MSAL Python has long been supporting "Subject Name / Issuer authentication" (defined here `https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/60`) with this API:
   ```python
   ConfidentialClientApplication("client_id", client_credential={
       "public_certificate": "--- BEGIN CERTIFICATE --- ... ",
       ...
   }, ...)
   ```
   but MSAL Python did not use it in MSAL's own test environment.

2. [MSAL Python supports using client certificate in .pfx format since version 1.29](https://github.com/AzureAD/microsoft-authentication-library-for-python/releases/tag/1.29.0), and started using it in the test automation.

3. Recently, MSAL team's lab infrastructure requires "Subject Name / Issuer authentication" (defined here `https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/60`), otherwise the test automation will fail.

So, this PR adds SNI to the pfx code path, using this API.

```python
ConfidentialClientApplication("client_id", client_credential={
    "public_certificate": True,  # The cert will be read from .pfx
    "private_key_pfx_path": "/path/to/cert.pfx",
}, ...)
```

This new API will be released in MSAL Python 1.30.0. [Docs are staged here](https://msal-python.readthedocs.io/en/docs-staging/#msal.ClientApplication.params.client_credential).

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -61,17 +61,23 @@ def _str2bytes(raw):
         return raw
 
 
-def _load_private_key_from_pfx_path(pfx_path, passphrase_bytes):
+def _parse_pfx(pfx_path, passphrase_bytes):
     # Cert concepts https://security.stackexchange.com/a/226758/125264
-    from cryptography.hazmat.primitives import hashes
+    from cryptography.hazmat.primitives import hashes, serialization
     from cryptography.hazmat.primitives.serialization import pkcs12
     with open(pfx_path, 'rb') as f:
         private_key, cert, _ = pkcs12.load_key_and_certificates(  # cryptography 2.5+
             # https://cryptography.io/en/latest/hazmat/primitives/asymmetric/serialization/#cryptography.hazmat.primitives.serialization.pkcs12.load_key_and_certificates
             f.read(), passphrase_bytes)
+    if not (private_key and cert):
+        raise ValueError("Your PFX file shall contain both private key and cert")
+    cert_pem = cert.public_bytes(encoding=serialization.Encoding.PEM).decode()  # cryptography 1.0+
+    x5c = [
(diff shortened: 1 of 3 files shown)


bgavrilMS:
All the other MSALs use the naming `x5c` or  `sendX5C`. The name "public certificate" is difficult to understand, especially as it is a "bool".

 I get that it's for consistency with the other API.

rayluo:
> All the other MSALs use the naming `x5c` or `sendX5C`. The name "public certificate" is difficult to understand, especially as it is a "bool".
> 
> I get that it's for consistency with the other API.

The `x5c` is an overly shortened naming style only observed in JWT, which is exotic to people who are unfamiliar with JWT or OAuth2; MSAL shall ideally abstract those details away from app developer.

[By definition](https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.6), the `x5c` is a container that "... contains the X.509 **public key certificate** ...", so, `public_certificate` sounds about right. Conceptually, it also feels logical that "if you want eSTS to authenticate your app based on your public cert's subject name & issuer, then please provide that public cert"; that is better than "if you want ..., use param `x5c` even though you don't know whether [x5c means a drone](https://www.symatoys.com/goodshow/x5c-syma-x5c-explorers.html)".

That being said, given that `x5c` is used on multiple MSALs, it probably won't harm if MSAL Python also introduces a boolean `x5c=True` as an alias for `public_certificate=True`. The latter is going to stay because we have to continue support the `public_certificate="--- BEGIN CERTIFICATE --- ... PEM format"` usage, so the MSAL Python community already knows and expects `public_certificate`.

bgavrilMS (approved):

PR #722, opened 2024-07-16
Using SHA256 and PSS padding

### LINKED ISSUE #639: [Feature Request] AAD client assertions should be computed using SHA 256 and an approved padding scheme
### MSAL client type

Confidential

### Problem Statement

When MSAL creates the client assertion, it uses PKCS1 padding for digital signature and SHA1 as x5t claim. These are old crypto algorithms and we need to move to newer versions. The STS is building support.

See ESTS work items :

https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2655345
https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2704466

### Proposed solution

Use x5t#s256 and PSS padding when talking to ESTS, CIAM, B2C(?) but not with ADFS.

 ### Original issue

 https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/4428

rayluo:
No new API is added in this PR. The SHA256 behavior will automatically kick in when using the recent .pfx support AND when the authority is not ADFS.

Manually tested and saw the `x5t#S256` and the `"alg": "PS256"` visible in the header. All the existing E2E test cases are now using this new SHA256 behavior, and they still pass.

Also manually tested the SHA1 code path and it still works. (Did not add it as automated test case because currently there is no public API to control the opt-in and opt-out.)

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -75,9 +75,10 @@ def _parse_pfx(pfx_path, passphrase_bytes):
     x5c = [
         '\n'.join(cert_pem.splitlines()[1:-1])  # Strip the "--- header ---" and "--- footer ---"
     ]
+    sha256_thumbprint = cert.fingerprint(hashes.SHA256()).hex()  # cryptography 0.7+
     sha1_thumbprint = cert.fingerprint(hashes.SHA1()).hex()  # cryptography 0.7+
         # https://cryptography.io/en/latest/x509/reference/#x-509-certificate-object
-    return private_key, sha1_thumbprint, x5c
+    return private_key, sha256_thumbprint, sha1_thumbprint, x5c
 
 
 def _load_private_key_from_pem_str(private_key_pem_str, passphrase_bytes):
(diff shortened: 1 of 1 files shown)


localden (approved):

msal/application.py
```diff
@@ -75,9 +75,10 @@ def _parse_pfx(pfx_path, passphrase_bytes):
     x5c = [
         '\n'.join(cert_pem.splitlines()[1:-1])  # Strip the "--- header ---" and "--- footer ---"
     ]
+    sha256_thumbprint = cert.fingerprint(hashes.SHA256()).hex()  # cryptography 0.7+
```
Robbie-Microsoft, line 78: Can you add the below checks to this function so that you're only returning one thumbprint?

rayluo: You mean a test case? We do not currently have such a test case, but its utilization is in [this part of the PR](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/722/files#diff-4d32d284fa6acf43894719eff11cb366e5969e0b61593bfd00c68a4cf9c441caR772-R781). You can see that we pass in only one of the thumbprints, not both.

msal/application.py
```diff
@@ -763,13 +765,22 @@ def _build_client(self, client_credential, authority, skip_regional_client=False
...
                     and isinstance(client_credential.get('public_certificate'), str)
                 ):  # Then we treat the public_certificate value as PEM content
                     headers["x5c"] = extract_certs(client_credential['public_certificate'])
+                if sha256_thumbprint and not authority.is_adfs:
+                    assertion_params = {
+                        "algorithm": "PS256", "sha256_thumbprint": sha256_thumbprint,
+                    }
+                else:  # Fall back
+                    if not sha1_thumbprint:
+                        raise ValueError("You shall provide a thumbprint in SHA1.")
```
Avery-Dunn, line 778: Are the SHA thumbprints something that the developer is meant to provide, and would they see this error message if something went wrong? If so, you may want something more informative like "A SHA1 thumbprint is needed for ADFS" or similar to let them know which SHA version is needed for each flow.

Avery-Dunn (approved):

msal/application.py
rayluo: When an app developer encounters this error message, he or she is already in the ADFS code path, so the "you shall provide a thumbprint in SHA1" is good enough as an actionable item.

I incline to avoid mentioning too-specific info such as "ADFS" in this case, because that kind of wording tends to become outdated when/if we will add or remove some scenarios into this SHA1 group.

PR #723, opened 2024-07-16
Expose refresh_on (if any) to fresh or cached response

### LINKED ISSUE #720: [Feature Request] Expose `refresh_on` when retrieving token from cache
### MSAL client type

Public, Confidential

### Problem Statement

Similar to https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/822, in our Azure SDKs for Python, one of the primary pipeline policies our SDK clients use is the [`BearerTokenCredentialPolicy`](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py#L80). This policy will save a token it retrieves using the passed-in credential and will keep  reusing that token until it's close to expiration. We recently added the ability for the policy to request a token sooner if a `refresh_on` value is set and has elapsed. We'd like to be able to set it if this value is available.

**Request**: Expose `refresh_on` where possible. 

****

### Proposed solution

_No response_

rayluo:
This PR will close #720 

CC @pvaneck, @xiangyan99

### DIFF
--- a/msal/managed_identity.py
+++ b/msal/managed_identity.py
@@ -273,8 +273,10 @@ class ManagedIdentityClient(object):
                     "token_type": entry.get("token_type", "Bearer"),
                     "expires_in": int(expires_in),  # OAuth2 specs defines it as int
                 }
-                if "refresh_on" in entry and int(entry["refresh_on"]) < now:  # aging
-                    break  # With a fallback in hand, we break here to go refresh
+                if "refresh_on" in entry:
+                    access_token_from_cache["refresh_on"] = int(entry["refresh_on"])
+                    if int(entry["refresh_on"]) < now:  # aging
+                        break  # With a fallback in hand, we break here to go refresh
                 return access_token_from_cache  # It is still good as new
         try:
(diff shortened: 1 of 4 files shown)


msal/managed_identity.py
```diff
@@ -274,6 +274,7 @@ def acquire_token_for_client(self, *, resource):  # We may support scope in the
                     "expires_in": int(expires_in),  # OAuth2 specs defines it as int
                 }
                 if "refresh_on" in entry and int(entry["refresh_on"]) < now:  # aging
+                    access_token_from_cache["refresh_on"] = int(entry["refresh_on"])
                     break  # With a fallback in hand, we break here to go refresh
```
pvaneck: Should break up the conditional like you did in `application.py`?
```suggestion
                if "refresh_on" in entry:
                    access_token_from_cache["refresh_on"] = int(entry["refresh_on"])
                    if int(entry["refresh_on"]) < now:  # aging
                        break  # With a fallback in hand, we break here to go refresh
```

PR #731, opened 2024-08-09
Change arc mi's detection algorithm

### LINKED ISSUE #727: [Feature Request] Add file based detection for Azure Arc
### MSAL client type

Confidential

### Problem Statement

In addition to env variable add file based detection for azure arc.

For more details see 
* [Guide to determine Azure arc env](https://msazure.visualstudio.com/One/_wiki/wikis/One.wiki/233012/VM-Extension-Authoring-for-Arc?anchor=determining-which-endpoint-to-use)
* [PR for MSAL.Net implementation](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4856)

### Proposed solution

_No response_

rayluo:
This PR will close #727

### DIFF
--- a/msal/managed_identity.py
+++ b/msal/managed_identity.py
@@ -316,6 +316,17 @@ def _scope_to_resource(scope):  # This is an experimental reasonable-effort appr
     return scope  # There is no much else we can do here
 
 
+def _get_arc_endpoint():
+    if "IDENTITY_ENDPOINT" in os.environ and "IMDS_ENDPOINT" in os.environ:
+        return os.environ["IDENTITY_ENDPOINT"]
+    if (  # Defined in https://msazure.visualstudio.com/One/_wiki/wikis/One.wiki/233012/VM-Extension-Authoring-for-Arc?anchor=determining-which-endpoint-to-use
+        sys.platform == "linux" and os.path.exists("/var/opt/azcmagent/bin/himds")
+        or sys.platform == "win32" and os.path.exists(os.path.expandvars(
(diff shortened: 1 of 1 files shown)


msal/managed_identity.py
```diff
@@ -316,6 +316,17 @@ def _scope_to_resource(scope):  # This is an experimental reasonable-effort appr
     return scope  # There is no much else we can do here
 
 
+def _get_arc_endpoint():
+    if "IDENTITY_ENDPOINT" in os.environ and "IMDS_ENDPOINT" in os.environ:
+        return os.environ["IDENTITY_ENDPOINT"]
```
jiasli, line 321: If detecting `/var/opt/azcmagent/bin/himds` and `%ProgramFiles%\AzureConnectedMachineAgent\himds.exe` is a reliable method, why bother checking `IDENTITY_ENDPOINT` and `IMDS_ENDPOINT`?

neha-bhargava (approved):

msal/managed_identity.py
rayluo: Indeed, that is an open topic. @neha-bhargava asked the Arc team specifically on this but received no response.

The current implementation here aligns with the implementation in MSAL .Net. In any case, this subtlety shall be considered as an implementation detail, which might be adjusted in the future again.

neha-bhargava: Agree with @rayluo. Added this in addition to env variable detection since env variables detection have been a success in case of windows. Might remove the env variable detection in future.

PR #747, opened 2024-09-19
client_assertion can accept a callback

### LINKED ISSUE #705: [Supportability Bug] Deprecate WithClientAssertion(string) as it leads to assertion expiration issues
### Scenario

ConfidentialClient - web site (AcquireTokenByAuthCode), ConfidentialClient - web api (AcquireTokenOnBehalfOf), ConfidentialClient - service to service (AcquireTokenForClient)

### Is this a new or an existing app?

None

### Issue description and reproduction steps

We keep getting feedback that `WithClientAssert(string)` causes expiration issues.

### Relevant code snippets

_No response_

### Expected behavior

_No response_

### Identity provider

Microsoft Entra ID (Work and School accounts and Personal Microsoft accounts)

### Regression

_No response_

### Solution and workarounds

_No response_

### LINKED ISSUE #746: [Feature Request] Client assertion should accept a callback function
### MSAL client type

Confidential

### Problem Statement

When client_credential is set to client_assertion, the only acceptable value is a token. However, tokens have a definite validity and as a result, the library as is it not suitable when using dynamic tokens - such as when using workload identities in AKS - because at some point the token becomes expired and a new instance of the Confidential client has to be initiated, which is not straightforward.

### Proposed solution

I propose that client_assertion should accept a callback function which is evaluated on demand by the MSAL library, allowing the user-supplied function to retrieve a custom token.

This feature is already available in the .NET and JS versions of this library.

rayluo:
* `client_assertion` can accept a callback.
* Also expose a `AutoRefresher` and document their usage. The new doc is [staged here](https://msal-python.readthedocs.io/en/docs-staging/#msal.ClientApplication.params.client_credential), just search the keyword "Supporting raw assertion obtained from elsewhere"
* Deprecate `client_assertion` as a string

[Internal work item is tracked here](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3289717).

This will close #705 and close #746


msal/application.py
```diff
@@ -757,6 +780,11 @@ def _build_client(self, client_credential, authority, skip_regional_client=False
             # Use client_credential.get("...") rather than "..." in client_credential
             # so that we can ignore an empty string came from an empty ENV VAR.
             if client_credential.get("client_assertion"):
+                if isinstance(client_credential["client_assertion"], str):
+                    warnings.warn(
+                        "Please use a callback instead of a string. See also "
```
bgavrilMS, line 785:
```suggestion
                        "Use a callback instead of a string. MSAL will invoke the callback every time it needs to fetch a token from the identity provider. The implementer is responsible for returning non-expired assertion and for caching assertions. See also "
```

msal/application.py
```diff
@@ -304,7 +304,30 @@ def __init__(
 
             .. admonition:: Supporting raw assertion obtained from elsewhere
 
+                *Added in version 1.32.0*:
+
+                MSAL Python accepts a callback which shall obtain an assertion::
+
+                    import msal
+
+                    def assertion_callback():
```
bgavrilMS, line 313: Can you add claims as input to this callback? This will help implementers of FIC perform token revocation, where the claims need to be take into account by the assertion provider as well.

May be separate PR, but want to make sure adding parametewrs to this is not a breaking change.

msal/application.py
```diff
@@ -304,7 +304,30 @@ def __init__(
...
+                        return '''A fresh JWT with claims:
+                            aud, exp, iss, jti, nbf, and sub'''
+
+                    smart_callback = msal.AutoRefresher(
+                        assertion_callback,
+                        expires_in=3600,  # Whatever value that you see fit
```
bgavrilMS, line 319: Not sure what this auto refresher is. We should mention FIC scenarios in the docs instead.

msal/application.py
rayluo: Will accept this suggestion in my next commit.

We might as well also append "... with the help of a helper `AutoRefresher`. See also ...".

I'm providing more explanation about the AutoRefresher in the [last comment below](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/747#discussion_r1768459705).

msal/application.py
rayluo: > Not sure what this auto refresher is.

Naming is hard, and new name suggestions are welcome (perhaps `CallbackCache` instead?). But the functionality behind the "auto refresher" was needed since day 1, because even when in the legit authenticated-by-a-cert scenario, MSAL Python internally has (1) a helper function to use that cert to create a *short-lived* assertion, and (2) a helper class `AutoRefresher` to cache the callback outcome and only call the callback again after a predefined expiry time elapsed. Such a cache behavior is desirable for the `#2` scenario described in [this comment](https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/746#issuecomment-2359581836).

In this PR, we repurpose the `#1` callback behavior to accept any callback; and expose the `#2` cache behavior so that app developers don't need to reinvent the wheel.

> We should mention FIC scenarios in the docs instead.

I suppose, the `#1` callback can be used in any scenario including FIC, but the `#2` cache concept shall always be applicable.

msal/application.py
rayluo: An adjustable callback can be created as a class, so that it can maintain its state. The crux is that the MSAL code path calling this callback does not know what proper values to feed in. So, the app developer - or a higher-level API - will need to expect an assertion error from eSTS, and then adjust that callback instance accordingly.

```python
class AdjustsableCallback:
    extra_claims = None

    def callback(self):
        return "A new assertion based on extra claims: " + self.extra_claims

ac = AdjustableCallback()
cca = ConfidentialClientApplication(..., client_credential={"client_assertion": ac.callback})

result = cca.acquire_token_for_client(...)
if result.get("error"):
    ac.extra_claims = get_new_claims_somehow()
    result = cca.acquire_token_for_client(...)
```

Do you want me to also document this advance scenario into the docs this time?

bgavrilMS: The claims have to flow from MSAL's acquire_token_for_client call into the callback. An example would be good. 

Here's the equivalent .NET code


```csharp

// singleton
var cca = ConfidentialCLientApp(...)
                .WithClientAssertion(
                        (input) => {
                              string claims = input.Claims; 
                              GetAssertionFromMsiOrWhatever(claims);
                  });


// later in the code
var authResult = cca.acquireTokenForClient();  // ok, callback invoked, claims is empty

// call downstream API
// 401 + WWW-AUthenticate returned
// parse header and extract claims

// get new token with claims challenge
authResult = cca.acquireTokenForClient(claims); // ok, callback is invoked, claims is not empty
```

msal/application.py
bgavrilMS: FIC example would be good. You can use example from here: https://review.learn.microsoft.com/en-us/identity/microsoft-identity-platform/federated-identity-credentials?branch=main

tests/certificate-with-password.pem
```diff
@@ -0,0 +1,51 @@
+-----BEGIN ENCRYPTED PRIVATE KEY-----
```
Robbie-Microsoft, line 1: Does msal Python utilize CI pipeline? Can you get these test certs from the lab instead of committing them here?

rayluo: Those certs are not certs for any real services. We just created some dummy certs for testing the cert loading code path. Thanks for your review, though. :)

rayluo:
Note: As an aftermath of all commits being rewritten, this PR currently contains `+16,009 −1,044` lines of changes from 250 commits happened in last 2 years, plus a merge conflict. This is un-review-able. Please do not review until I fix it.

bgavrilMS:
When are you merging this @rayluo ?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,32 @@
+---
```
bgavrilMS, line 1: There is a rebase problem with this PR, as it includes 867 commits?

PR #752, opened 2024-09-26
Warning when obsolete msal-extensions is detected

### LINKED ISSUE #745: [Bug] Updates to msal/application.py may now be out-of-sync with msal_extensions
**Describe the bug**

I just upgraded the MSAL Python package and all of our silent authentication routines started breaking - these authenticate manually and cache the access token for future calls. Interactive authentication works and the token cache appears to save the the account and token correctly, but when the repeat call is made with silent it does not find an account and fails.

I have diagnosed the error and traced it back to [this commit](https://github.com/AzureAD/microsoft-authentication-library-for-python/commit/41dbf29f4bcc23a54cd3de3141c1caa43455f0bb) that changes `self.token_cache.find` to `self.token_cache.search` in several places. 

It appears that upgraded to a MSAL version at or after this point requires an update of `msal_extensions`, upgrading from `msal_extensions-1.0 .0` to `msal_extensions-1.2.0` seems to work correctly, however, it was not obvious that this was the issue because both `find` and `search` are valid methods on the msal_extensions 1.0.0 token cache, but `find` finds the previously cached accounts and `search`  does not.

msal_extensions understandably is not a requirement for the MSAL package, but I would recommend creating some type of link to ensure the versions are in sync when there are breaking changes (especially since the commit history suggests you maintain/contribute to both).  

And if that's not doable, perhaps anyone else who runs into this and searches will find this bug report :)

**To Reproduce**
1. `pip install msal_extensions==1.0.0`
2. `pip install --upgrade msal`
3.  Run a script or test that authenticates interactively and caches the result with an `msal_extensions.PersistedTokenCache`
4. Run a script or test that authenticates silently and points to the same cache. 

**Expected behavior**
4 should succeed

**What you see instead**

This fails because `{msalapp}.get_accounts()` returns nothing due to the `search` call.

1.31.0

**Additional context**

Steps above are fixed if (1) is changed to `pip install msal_extensions>=1.2.0`

rayluo:
The lengthy discussion in ["Updates to msal/application.py may now be out-of-sync with msal_extensions"
#745](https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/745) considered a mitigation of introducing a new optional dependency `msal[extensions]` for MSAL. However, after offline discussion, we believe it has two drawbacks: (1) it is hardly discoverable for most `msal-extensions` users, and (2) it introduces a permanent "msal[extensions]" declaration only for trivial gain.

So, we ended up settling on an alternative. MSAL will now detect the `msal-extensions` from the current environment. If its version is lower than 1.2.0, MSAL will emit a warning like this:

```
/.../msal/application.py:203: UserWarning: Please upgrade msal-extensions. Only msal-extensions 1.2+ can work with msal 1.30+
```

This PR will close #745 .

CC @jiasli , @kcphila

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -194,6 +194,21 @@ class _ClientWithCcsRoutingInfo(Client):
             username, password, headers=headers, **kwargs)
 
 
+def _msal_extension_check():
+    # Can't run this in module or class level otherwise you'll get circular import error
+    try:
+        from msal_extensions import __version__ as v
+        major, minor, _ = v.split(".", maxsplit=3)
+        if not (int(major) >= 1 and int(minor) >= 2):
+            warnings.warn(
+                "Please upgrade msal-extensions. "
+                "Only msal-extensions 1.2+ can work with msal 1.30+")
+    except ImportError:
+        pass  # The optional msal_extensions is not installed. Business as usual.
+    except ValueError:
+        logger.exception(f"msal_extensions version {v} not in major.minor.patch format")
+
+
 class ClientApplication(object):
     """You do not usually directly use this class. Use its subclasses instead:
     :class:`PublicClientApplication` and :class:`ConfidentialClientApplication`.
@@ -635,6 +650,8 @@ class ClientApplication(object):
         self.authority_groups = None
         self._telemetry_buffer = {}
         self._telemetry_lock = Lock()
+        _msal_extension_check()
+
 
     def _decide_broker(self, allow_broker, enable_pii_log):
         is_confidential_app = self.client_credential or isinstance(


msal/application.py
```diff
@@ -194,6 +194,21 @@ def obtain_token_by_username_password(self, username, password, **kwargs):
             username, password, headers=headers, **kwargs)
 
 
+def _msal_extension_check():
+    # Can't run this in module or class level otherwise you'll get circular import error
+    try:
+        from msal_extensions import __version__ as v
+        major, minor, _ = v.split(".", maxsplit=3)
```
jiasli, line 201: Just as an FYI, Azure CLI uses [`packaging.version`](https://packaging.pypa.io/en/latest/version.html) for version parsing:

https://github.com/Azure/azure-cli/blob/b6ab754b68efba60b30806fa80dc3d0377be84ca/src/azure-cli-core/azure/cli/core/extension/operations.py#L18

but it may be too heavy for MSAL.

msal/application.py
```diff
@@ -194,6 +194,21 @@ def obtain_token_by_username_password(self, username, password, **kwargs):
             username, password, headers=headers, **kwargs)
 
 
+def _msal_extension_check():
+    # Can't run this in module or class level otherwise you'll get circular import error
```
jiasli, line 198: Nice comment! Makes things clear.

jiasli (approved):

msal/application.py
rayluo: Indeed. In fact, `msal-extensions` used to also use `packaging.version` but have since removed it in `https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/125`.

PR #756, opened 2024-10-11
MSAL will use env var MSAL_FORCE_REGION by default

### LINKED ISSUE #748: [Feature Request] Region auto enable on env variable
## Region auto-enable

1. On creation of a ConfidentialClientApplication, MSAL shall detect env variable `MSAL_FORCE_REGION`, which will be set to a specific region (e.g. `westus1`)
2. If this env variable is set, MSAL shall opt-in to ESTS-R with the value of this variable.

## Disabling the behavior

If region is explicitly set in MSAL to the string "DisableMsalForceRegion`, then the above no longer applies

## Conflicts with `WithAzureRegion(xyz)`

Use of the api `WithRegion(xyz)` takes precedence over the env variable. 

## Acceptance tests

For all of the tests, assume env variable MSAL_FORCE_REGION=eastus

1. No other config is used. ACTUAL region used: eastus
2. App developer configures region "westus" in MSAL. ACTUAL region used: westus
3. App developer configures region "DisableMsalForceRegion" in MSAL. ACTUAL region used: none

 ### Original issue: 
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/4930

rayluo:
This pull request updates the `msal/application.py` file to include a new feature that allows MSAL to use the `MSAL_FORCE_REGION` environment variable by default. This closes #748 

Docs are also updated and [staged here](https://msal-python.readthedocs.io/en/docs-staging/#msal.ClientApplication.params.azure_region).

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -5,6 +5,7 @@ import logging
 import sys
 import warnings
 from threading import Lock
+from typing import Optional  # Needed in Python 3.7 & 3.8
 import os
 
 from .oauth2cli import Client, JwtAssertionCreator
@@ -225,6 +226,7 @@ class ClientApplication(object):
     REMOVE_ACCOUNT_ID = "903"
 
     ATTEMPT_REGION_DISCOVERY = True  # "TryAutoDetect"
+    DISABLE_MSAL_FORCE_REGION = False  # Used in azure_region to disable MSAL_FORCE_REGION behavior
     _TOKEN_SOURCE = "token_source"
     _TOKEN_SOURCE_IDP = "identity_provider"
     _TOKEN_SOURCE_CACHE = "cache"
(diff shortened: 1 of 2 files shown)


msal/application.py
```diff
@@ -448,11 +449,14 @@ def __init__(
...
 
-              ``azure_region=None`` - meaning no region is used. This is the default value.
-              ``azure_region="some_region"`` - meaning the specified region is used.
-              ``azure_region=True`` - meaning MSAL will try to auto-detect the region. This is not recommended.
+            1. ``azure_region=None`` - This default value means no region is configured.
+               MSAL will use the region defined in env var ``MSAL_FORCE_REGION``.
+            2. ``azure_region="some_region"`` - meaning the specified region is used.
+            3. ``azure_region=True`` - meaning
+               MSAL will try to auto-detect the region. This is not recommended.
+            4. ``azure_region=False`` - meaning MSAL will use no region.
```
bgavrilMS, line 460: Not as per spec. Let's please keep it as per spec.

bgavrilMS (changes requested):
Looks good, but please let's use `DisableMsalForceRegion` string to be consistent with the other MSALs.

msal/application.py
rayluo: Per offline discussion, an alias `DISABLE_MSAL_FORCE_REFRESH` is now added [here](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/756/files#diff-4d32d284fa6acf43894719eff11cb366e5969e0b61593bfd00c68a4cf9c441caR229).

bgavrilMS (approved):

PR #759, opened 2024-10-21
Customizable token cache

rayluo:
Inspired by [its cousin in MSAL .Net](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/4922), this PR refactors MSAL Python's token cache to provide an INTERNAL helper which can store additional values from token request or response.

Implementation details (which are not exactly the same as MSAL .Net, but the overall purpose is the same):

* We have to differentiate between request and response because they are two different data sources in this code base. And the use case (CDT experiment) needs both of them.
* The same additional values are also used in token search.
* Currently, the changes above are only applicable to access token.
* The preexisting hard-coded `key_id` was also refactored this time to utilize the new generic declaration, and the existing test case was updated to reflect that.
* This PR does not bring other behavior changes, and all the existing test cases still pass. So, it should be safe to merge.
* Some other prototypes for recent projects will be built on top of this PR.

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -1,3 +1,4 @@
+from __future__ import annotations
 import functools
 import json
 import time
@@ -237,6 +238,10 @@ class ClientApplication(object):
         "auth_scheme is currently only available from broker. "
         "You can enable broker by following these instructions. "
         "https://msal-python.readthedocs.io/en/latest/#publicclientapplication")
+
+    _TOKEN_CACHE_DATA: dict[str, str] = {  # field_in_data: field_in_cache
+        "key_id": "key_id",  # Some token types (SSH-certs, POP) are bound to a key
+    }
 
     def __init__(
             self, client_id,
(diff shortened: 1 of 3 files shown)


msal/application.py
```diff
@@ -238,6 +239,10 @@ class ClientApplication(object):
         "You can enable broker by following these instructions. "
         "https://msal-python.readthedocs.io/en/latest/#publicclientapplication")
 
+    _TOKEN_CACHE_DATA: dict[str, str] = {  # field_in_data: field_in_cache
```
maliksahil, line 242: The type annotation dict[str, str] for _TOKEN_CACHE_DATA uses Python 3.9+ syntax. If you intend to support Python 3.7 or 3.8 (as implied by the import on line 9), consider using Dict[str, str] from typing for compatibility.

maliksahil: Consider adding a comment
     # Maps field names from input data to their corresponding field names in the token cache.
     # This is used to ensure that certain token types (e.g., SSH certificates, POP tokens)
     # are correctly associated with their identifying keys in the cache.

rayluo: Hi @maliksahil , it has been a while, how is it going? :-)

Thanks for catching that type annotation issue. I'll take a look at why that typo was not caught by our test automation, when we revive this PR.

Why did you notice this PR in the first place? Are you working on some cutting-edge scenario that needs the behavior of this PR?

maliksahil: Haha going well my friend. We are always looking at cutting edge stuff. Hope you and everyone else are well.

PR #766, opened 2024-11-07
Enable broker support on Linux for WSL

DharshanBJ:


setup.cfg
```diff
@@ -62,9 +62,11 @@ broker =
     # most existing MSAL Python apps do not have the redirect_uri needed by broker.
     #
     # We need pymsalruntime.CallbackData introduced in PyMsalRuntime 0.14
-    pymsalruntime>=0.14,<0.18; python_version>='3.6' and platform_system=='Windows'
+    pymsalruntime>=0.14,<0.19; python_version>='3.6' and platform_system=='Windows'
     # On Mac, PyMsalRuntime 0.17+ is expected to support SSH cert and ROPC
-    pymsalruntime>=0.17,<0.18; python_version>='3.8' and platform_system=='Darwin'
+    pymsalruntime>=0.17,<0.19; python_version>='3.8' and platform_system=='Darwin'
+    # PyMsalRuntime 0.18+ is expected to support broker on Linux
+    pymsalruntime>=0.18,<0.19; python_version>='3.8' and platform_system=='Linux'
```
fengga, line 69: Have you released or do you plan to release 0.18.0 for Linux? I saw right now it's still 0.17.1
https://pypi.org/project/pymsalruntime/#history

DharshanBJ: I have not yet released 0.18.0, but i'll be making that release soon, and this will be the first package with linux broker support

tests/broker-test.py
```diff
@@ -39,7 +39,8 @@
     _AZURE_CLI,
     authority="https://login.microsoftonline.com/organizations",
     enable_broker_on_mac=True,
-    enable_broker_on_windows=True)
```
fengga, line 42: Thanks, you find this test file, please make sure it is able to invoke broker and you need to manually finish the signin process in the broker prompt.

fengga (dismissed):
Thanks, this PR looks good to me. And please make sure get an approval from Ray.

tests/test_e2e.py
```diff
@@ -200,6 +200,7 @@ def _build_app(cls,
                 http_client=http_client or MinimalHttpClient(),
                 enable_broker_on_windows=broker_available,
                 enable_broker_on_mac=broker_available,
+                enable_broker_on_linux=broker_available,
```
rayluo: **praise:** Thanks for finding this test file!

**suggestion:** If you haven't already, please also do `python -m unittest tests.test_e2e` and make sure we won't run into any issue. Note that you will need to setup an `.env` first, whose content is described at the beginning of this test_e2e.py file.

tests/broker-test.py
```diff
@@ -39,7 +39,8 @@
...
+    enable_broker_on_windows=True,
+    enable_broker_on_linux=True)
```
rayluo: **nit:** Write it this way, so that future addition will result in a cleaner PR.
```suggestion
    enable_broker_on_linux=True,
    )
```

setup.cfg
rayluo: **suggestion:** This PR shall be changed to a draft. We won't merge it in until pymsalruntime 0.18 being released.

sample/interactive_sample.py
```diff
@@ -47,7 +47,8 @@
     oidc_authority=os.getenv('OIDC_AUTHORITY'),  # For External ID with custom domain
     #enable_broker_on_windows=True,  # Opted in. You will be guided to meet the prerequisites, if your app hasn't already
     #enable_broker_on_mac=True,  # Opted in. You will be guided to meet the prerequisites, if your app hasn't already
-
+    #enable_broker_on_linux=True,  # Opted in. You will be guided to meet the prerequisites, if your app hasn't already
+
```
rayluo:
```suggestion
    #enable_broker_on_linux=True,  # Opted in. You will be guided to meet the prerequisites, if your app hasn't already
```
**nit:** Remove unnecessary trailing space.
**suggestion:** Most modern editors shall support automatically trimming trailing spaces. Find that setting in your editor and change it.

msal/application.py
```diff
@@ -672,6 +672,7 @@ def _decide_broker(self, allow_broker, enable_pii_log):
                 "allow_broker is deprecated. "
                 "Please use PublicClientApplication(..., "
                 "enable_broker_on_windows=True, "
+                "enable_broker_on_linux=True, "
                 "enable_broker_on_mac=...)",
```
rayluo: **issue:** The `allow_broker=True` is only equivalent to - and can only be translated as - `enable_broker_on_windows=True`, not the other two. Sorry if the previous `enable_broker_on_mac=...` wording causes any confusion. Now I suggest to replace both the linux and mac lines with a `...)`.

```suggestion
                "...)",
```

msal/application.py
```diff
@@ -22,7 +22,7 @@
 
 
 # The __init__.py will import this. Not the other way around.
-__version__ = "1.31.0"  # When releasing, also check and bump our dependencies's versions if needed
+__version__ = "1.32.0"  # When releasing, also check and bump our dependencies's versions if needed
```
rayluo: **issue:** Please revert this change. We shall only bump version number during a release process. That way, a potential revert of a feature branch would not mess up the version number (because version numbers - once released - can never be reverted).

rayluo (changes requested):
Thanks for a clean PR! Implementation wise, it looks good. I added some suggestions above, mostly in terms of our workflow. Please make corresponding changes and then wait for the PyMsalRuntime release.

msal/application.py
DharshanBJ: done

setup.cfg
pvaneck: Can we ensure the next release of `pymsalruntime` have whls built for Python 3.13 as well? The latest version of `pymsalruntime` doesn't seem to have that.

rayluo (changes requested):
We will also need to change the [precise dependency version error message here](https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/dharshanb/brokerSupportLinux/msal/broker.py#L26-L29) and a [approximate version hint there](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/766/files#diff-4d32d284fa6acf43894719eff11cb366e5969e0b61593bfd00c68a4cf9c441caR1954)

msal/application.py
```diff
@@ -1921,7 +1921,7 @@ def __init__(self, client_id, client_credential=None, **kwargs):
 
         .. note::
 
-            You may set enable_broker_on_windows and/or enable_broker_on_mac to True.
+            You may set enable_broker_on_windows and/or enable_broker_on_mac and/or enable_broker_on_linux to True.
```
rayluo: **Question:** Does the Linux broker require yet another new redirect_uri being registered? If so, please also update the documentation [here](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/766/files#diff-4d32d284fa6acf43894719eff11cb366e5969e0b61593bfd00c68a4cf9c441caR1943-R1954).

DharshanBJ: No, we use the same redirect url as windows for linux.

tests/broker-test.py
DharshanBJ: Confirmed that we are able to invoke the broker on linux and manually finish the signin process.

setup.cfg
DharshanBJ: we have now released pymsalruntime 0.18.0

DharshanBJ: we have a separate work item to add support for python 3.13, we will prioritize that work soon.

DharshanBJ:
> We will also need to change the [precise dependency version error message here](https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/dharshanb/brokerSupportLinux/msal/broker.py#L26-L29) and a [approximate version hint there](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/766/files#diff-4d32d284fa6acf43894719eff11cb366e5969e0b61593bfd00c68a4cf9c441caR1954)

updated

tests/test_e2e.py
DharshanBJ: We don't support POP and ssh-cert flows for linux yet, tested the other e2e flows.

msal/application.py
```diff
@@ -1951,7 +1955,7 @@ def __init__(self, client_id, client_credential=None, **kwargs):
                  if your app is expected to run on Mac
```
rayluo, line 1971:
```suggestion
                 if your app is expected to run on Mac
               * ``ms-appx-web://Microsoft.AAD.BrokerPlugin/your_client_id``
                 if your app is expected to run on Linux, especially WSL
```

thomasaarholt:
@DharshanBJ, I would love to see this wrapped up.

I arrived here after trying to debug errors with using msal on WSL. This PR worked "as-advertised" and made my life a whole lot easier.

Before your PR, the following snippet kept returning a browser window with `The redirect URI 'http://localhost:<some port number>' specified in the request does not match the redirect URIs configured for the application`. 

Testing with `enable_broker_on_linux=True` "just worked". 🚀
```python
    app = PublicClientApplication(
        <client_id>,
        authority=<authority>,
        enable_broker_on_linux=True,
    )
    app.acquire_token_interactive(...)
```

thomasaarholt:
This work enables Data Scientists at Microsoft to develop and debug their pipelines better. Some of our models are deployed on linux environments. We believe that it is crucial to develop and test our work on the same environment that we deploy on, which is why I highly approve of this PR 😊 

@rayluo any chance you could take a look at the state of this PR when you have the time?

thomasaarholt:
I have one oddity. When using the broker on Linux, I get the following text printed `"{{\"ping\",\"success\"}}}"`
image

This does not happen on Windows:
image

DharshanBJ:
> I have one oddity. When using the broker on Linux, I get the following text printed `"{{\"ping\",\"success\"}}}"` image
> 
> This does not happen on Windows: image

The "{{\"ping\",\"success\"}}}" message is expected on WSL, it's from initializing the msal.wsl.proxy execuatble which is needed to talk to the windows broker on WSL.
On pure windows, msal.wsl.proxy is not needed and hence we dont see "{{\"ping\",\"success\"}}}"

msal/application.py
```diff
@@ -704,7 +704,7 @@ def _decide_broker(self, allow_broker, enable_pii_log):
 
     def is_pop_supported(self):
         """Returns True if this client supports Proof-of-Possession Access Token."""
-        return self._enable_broker
+        return self._enable_broker and sys.platform != "linux"
```
rayluo: Avoid using logic like "if var != one_thing", because you never know what the rest of things are. In this case, [`sys.platform` has much more values](https://docs.python.org/3/library/sys.html#sys.platform) than we should use.

Change this to `sys.platform in ("win32", "darwin")`. Same applies to other occurrences. Define a new internal helper when appropriate.

msal/application.py
```diff
@@ -1576,10 +1576,14 @@ def _acquire_token_silent_from_cache_and_possibly_refresh_it(
                     raise ValueError("auth_scheme is not supported in Cloud Shell")
                 return self._acquire_token_by_cloud_shell(scopes, data=data)
 
+            is_ssh_cert_or_pop_request = (
+                data.get("token_type") == "ssh-cert" or
+                data.get("token_type") == "pop" or
+                isinstance(auth_scheme, msal.auth_scheme.PopAuthScheme))
```
rayluo: There are two same snippets in this PR. How about refactoring it into an internal helper like this?

```python
def _is_ssh_cert_or_pop_request(token_type, auth_scheme) -> bool:
    ...

```

rayluo (changes requested):

rayluo:
> > I have one oddity. When using the broker on Linux, I get the following text printed `"{{\"ping\",\"success\"}}}"`
> 
> The "{{"ping","success"}}}" message is expected on WSL, it's from initializing the msal.wsl.proxy execuatble which is needed to talk to the windows broker on WSL. On pure windows, msal.wsl.proxy is not needed and hence we dont see "{{"ping","success"}}}"

@DharshanBJ , that feels like some sort of debug log. Could it be turned off or removed from that "msal.wsl.proxy"?

CC: @jiasli

msal/application.py
```diff
@@ -1992,15 +1998,23 @@ def __init__(self, client_id, client_credential=None, **kwargs):
...
+            This parameter defaults to None, which means MSAL will not utilize a broker.
+
+            New in MSAL Python 1.32.0. 
         """
         if client_credential is not None:
             raise ValueError("Public Client should not possess credentials")
         # Using kwargs notation for now. We will switch to keyword-only arguments.
         enable_broker_on_windows = kwargs.pop("enable_broker_on_windows", False)
         enable_broker_on_mac = kwargs.pop("enable_broker_on_mac", False)
+        enable_broker_on_linux = kwargs.pop("enable_broker_on_linux", False)
```
rayluo: The `dev` branch has received new changes. Please merge that into your branch, and then remove this line and add the new parameter into `__init__()`.

msal/application.py
```diff
@@ -2162,7 +2180,9 @@ def acquire_token_interactive(
                 )
```
rayluo, line 2197: One more thing. MSAL Python's caller (such as Azure CLI) can specify a parameter `app.acquire_token_interactive(..., prompt="select_account")` which will in turn call `pymsalruntime.signin_interactively()`. On Windows, that function will pop up the account selector, as requested. But on WSL, it does not pop any UI, and seemingly just return a token for the current user.

Is that expected, @DharshanBJ ? (Note that `az login` usually expects an account picker.) CC: @jiasli

rayluo:
> **it is crucial to develop and test our work on the same environment that we deploy on**, which is why I highly approve of this PR 😊

@thomasaarholt , we agree with **that part**. Is your work going to be deployed on WSL or on standalone Linux (which may or may not have Intune installed)?

Keep in mind that this PR is mainly targeting WSL. On standalone Linux, unless [Intune is installed](https://learn.microsoft.com/en-us/mem/intune/user-help/enroll-device-linux), MSAL Python will still fall back to browser-based UI. If your work after deployment will mainly run on standalone Linux without Intune, you might want to still focus testing the browser-based experience, for the **same reason** that you mentioned above.

FWIW, the default browser-based experience unfortunately deteriorates on WSL but technically still functions. In a freshly installed WSL ubuntu, it would end up with a message like this on the console "gio: `http://localhost:1234/...`: Operation not supported", but the app developer on WSL can still click on that url to bring up a browser page and completes the sign-in there. (Those are known issues [here](https://discuss.streamlit.io/t/gio-operation-not-supported/82863/3) and [there](https://github.com/AzureAD/microsoft-authentication-library-for-python/discussions/627).) If your end user is mainly on standalone Linux, the browser will typically automatically pop up without issue.

thomasaarholt:
Hi again!

> Is your work going to be deployed on WSL or on standalone Linux (which may or may not have Intune installed)?

Deployment happens on standalone linux, but we don't use the broker path for authentication there, so I'm not worried here. We're using this only to improve the dev-experience.

msal/application.py
DharshanBJ: updated

msal/application.py
DharshanBJ: updated

msal/application.py
DharshanBJ: updated

DharshanBJ:
> > > I have one oddity. When using the broker on Linux, I get the following text printed `"{{\"ping\",\"success\"}}}"`
> > 
> > 
> > The "{{"ping","success"}}}" message is expected on WSL, it's from initializing the msal.wsl.proxy execuatble which is needed to talk to the windows broker on WSL. On pure windows, msal.wsl.proxy is not needed and hence we dont see "{{"ping","success"}}}"
> 
> @DharshanBJ , that feels like some sort of debug log. Could it be turned off or removed from that "msal.wsl.proxy"?
> 
> CC: @jiasli

Yes, this can be done, we'll remove this as part of the next msal.wsl.proxy release

msal/application.py
DharshanBJ: No, an account picker will pop up even on WSL when prompt="select_account" is passed. I tested it on my machine and it works.
Please ensure your wsl2 instance is updated such that you have the latest msal.wsl.proxy.exe on your machine.
Its possible that the msal.wsl.proxy.exe might not be the latest, since the account picker logic was added to the proxy only in the latest release.

msal/__main__.py
```diff
@@ -300,6 +300,7 @@ def _main():
         instance_discovery=instance_discovery,
         enable_broker_on_windows=enable_broker,
         enable_broker_on_mac=enable_broker,
+        enable_broker_on_linux=enable_broker,
```
rayluo, line 303:
```suggestion
        enable_broker_on_linux=enable_broker,
        enable_broker_on_wsl=enable_broker,
```

tests/test_e2e.py
```diff
@@ -193,6 +193,7 @@ def _build_app(cls,
                 http_client=http_client or MinimalHttpClient(),
                 enable_broker_on_windows=_PYMSALRUNTIME_INSTALLED,
                 enable_broker_on_mac=_PYMSALRUNTIME_INSTALLED,
+                enable_broker_on_linux=_PYMSALRUNTIME_INSTALLED,
```
rayluo, line 196: Shall we also add this?
```suggestion
                enable_broker_on_linux=_PYMSALRUNTIME_INSTALLED,
                enable_broker_on_wsl=_PYMSALRUNTIME_INSTALLED,
```

msal/application.py
```diff
@@ -1998,12 +2006,27 @@ def __init__(
...
+            New in MSAL Python 1.32.0.
         """
         if client_credential is not None:
             raise ValueError("Public Client should not possess credentials")
+
         self._enable_broker = bool(
             enable_broker_on_windows and sys.platform == "win32"
-            or enable_broker_on_mac and sys.platform == "darwin")
+            or enable_broker_on_mac and sys.platform == "darwin"
+            or (enable_broker_on_linux or (enable_broker_on_wsl and is_wsl())) and sys.platform == "linux")
```
rayluo: **question (non-blocking)**
Is there any special case that necessitate the nested condition here? What would go wrong if we use the following easier-to-understand logic?
```suggestion
            or enable_broker_on_linux and sys.platform == "linux"
            or enable_broker_on_wsl and is_wsl()
            )
```

msal/application.py
```diff
@@ -2131,7 +2154,9 @@ def acquire_token_interactive(
                 # https://microsoft.sharepoint.com/:w:/t/Identity-DevEx/EatIUauX3c9Ctw1l7AQ6iM8B5CeBZxc58eoQCE0IuZ0VFw?e=tgc3jP&CID=39c853be-76ea-79d7-ee73-f1b2706ede05
             False
             ) and data.get("token_type") != "ssh-cert"  # Work around a known issue as of PyMsalRuntime 0.8
-        self._validate_ssh_cert_input_data(data)
+        self._validate_ssh_cert_input_data(data)
```
```suggestion
        self._validate_ssh_cert_input_data(data)
```

msal/application.py
DharshanBJ: Yes, this should work too, updated

msal/application.py
```diff
@@ -1958,9 +1964,11 @@ def __init__(
                  if your app is expected to run on Windows 10+
                * ``msauth.com.msauth.unsignedapp://auth``
                  if your app is expected to run on Mac
+               * ``ms-appx-web://Microsoft.AAD.BrokerPlugin/your_client_id``
+                 if your app is expected to run on Linux, especially WSL
 
             2. installed broker dependency,
-               e.g. ``pip install msal[broker]>=1.31,<2``.
+               e.g. ``pip install msal[broker]>=1.32,<2``.
```
rayluo: Since we shipped MSAL 1.32 just yesterday, all the mentioning of "1.32" shall be changed to "1.33" in this PR. Sorry for the inconvenience. :-)

msal/application.py
rayluo: I installed WSL2 with Ubuntu 24.04 right before I tested and left that comment, so, presumably I am already using an "updated" version of WSL.

A bit more detail is that when using the `python -m msal` and choosing prompt=select_account, I did notice that some sort of UI seemingly pops up, but it just disappeared within half a second, and MSAL ended up just returning a token for my default account. It is reproduceable on my WSL.

Can you lead the investigation to see what could be the issue here? I am not comfortable to merge this PR without solving this.

DharshanBJ: Discussed offline, we need to pick wsl release 2.5.1 for the account picker changes to work - https://github.com/microsoft/WSL/releases/tag/2.5.1

rayluo: UPDATE:
* It turns out that the account picker on WSL is currently only available from [WSL 2.5.1 Preview](https://github.com/microsoft/WSL/releases/tag/2.5.1), which needs to be installed by "wsl --update **--pre-release**".
* We are contacting WSL team to know their GA plan. We don't GA our feature when its dependency is still in preview.

DharshanBJ: The latest official release of wsl (2.4.13), has the account picker changes, tested it on my machine with wsl 2.4.13 and the account picker is popping up.

rayluo: Thanks. I have also tested it (by [downgrading the WSL](https://github.com/microsoft/WSL/issues/12711#issuecomment-2762381276) 2.5.1 preview to 2.4.13) and the account picker is popping up. Resolving this topic.

msal/application.py
```diff
@@ -1998,12 +2006,29 @@ def __init__(
...
+            New in MSAL Python 1.33.0.
         """
         if client_credential is not None:
             raise ValueError("Public Client should not possess credentials")
+
         self._enable_broker = bool(
             enable_broker_on_windows and sys.platform == "win32"
-            or enable_broker_on_mac and sys.platform == "darwin")
+            or enable_broker_on_mac and sys.platform == "darwin"
+            or enable_broker_on_linux and sys.platform == "linux"
```
rayluo, line 2033: **question**
Hi @DharshanBJ , I realize that the in the current implementation, the `enable_broker_on_wsl` below is for WSL-only, which matches its name well. But the `enable_broker_on_linux` logic here actually covers all linux including WSL, meaning, an `enable_broker_on_linux=True` on WSL will also enable broker. Is that what we want?

In other words, does `enable_broker_on_linux` meant to imply "all linux including WSL", or do we meant to have an `enable_broker_on_non_wsl_linux`?

(After we clarify this in this conversation, we shall fine tune the parameter name or doc accordingly.)

CC: @iulico-1 , @jiasli

DharshanBJ: Hi @rayluo, yes, enable_broker_on_linux will enable broker on wsl and baremetal linux (all linux) and  enable_broker_on_wsl enables broker only on wsl.

enable_broker_on_linux is meant to imply "all linux including WSL

msal/application.py
```diff
@@ -1998,12 +2006,29 @@ def __init__(
             This parameter defaults to None, which means MSAL will not utilize a broker.
 
             New in MSAL Python 1.31.0.
+            
+        :param boolean enable_broker_on_linux:
+            This setting is only effective if your app is running on Linux.
```
rayluo: **suggestion**
Adding this clarification based on [this conversation](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/766/files#r2019506415).
```suggestion
            This setting is only effective if your app is running on Linux, including WSL.
```

evelyn-ys (approved):

rayluo (approved):
Thanks for your effort all this time! Approving now, and I'll follow up with the merge and then the release shortly after.

jiasli:
Tested the code in https://github.com/Azure/azure-cli/pull/31169, but can't get it working.

The behavior is different from `enable_broker_on_windows` where an account selection window is opened. In WSL, no account selection window is opened. It fails to acquire access token even for the home tenant (Microsoft).

```
$ az login
...
Select the account you want to log in with. For more information on login with Azure CLI, see https://go.microsoft.com/fwlink/?linkid=2271136

Retrieving tenants and subscriptions for the selection...
Authentication failed against tenant 72f988bf-86f1-41af-91ab-2d7cd011db47 'Microsoft': Can't find token from MSAL cache.
```

jiasli (changes requested):
Doesn't launch account selection window.

rayluo:
> Doesn't launch account selection window.

Please [upgrade to WSL 2.4.13+](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/766/files#r2019328210)



> It fails to acquire access token even for the home tenant (Microsoft).

@DharshanBJ , can you investigate this one?

DharshanBJ:
> > Doesn't launch account selection window.
> 
> Please [upgrade to WSL 2.4.13+](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/766/files#r2019328210)
> 
> > It fails to acquire access token even for the home tenant (Microsoft).
> 
> @DharshanBJ , can you investigate this one?

Checked with jiasli, the issue is due to not having WSL 2.4.13+, since the user is not able to acquire the token interactively(due to not having the account picker changes), the error message "Authentication failed against tenant 72f988bf-86f1-41af-91ab-2d7cd011db47 'Microsoft': Can't find token from MSAL cache." is for that same reason.

jiasli:
I checked update in both Windows Store and Windows Update, but WSL is still on an old version

```
> wsl --version
WSL version: 2.1.5.0
Kernel version: 5.15.146.1-2
WSLg version: 1.0.60
MSRDC version: 1.2.5105
Direct3D version: 1.611.1-81528511
DXCore version: 10.0.25131.1002-220531-1700.rs-onecore-base2-hyp
Windows version: 10.0.26100.3476
```

`wsl --update` is only mentioned by 

- https://learn.microsoft.com/en-us/windows/wsl/basic-commands#update-wsl
- https://learn.microsoft.com/en-us/windows/wsl/install#want-to-try-the-latest-wsl-preview-features

https://learn.microsoft.com/en-us/windows/wsl/install should contain a section on "How to update WSL".

Also, we can't force or expect the users to update WSL to the latest version manually. It is better to have WSL updated by Windows Store or Windows Update automatically.

Besides this, the interface of this functionality looks good to me.

jiasli (approved):

fengga (approved):

gladjohn (approved):

setup.cfg
```diff
@@ -62,9 +62,11 @@ broker =
     # most existing MSAL Python apps do not have the redirect_uri needed by broker.
     #
     # We need pymsalruntime.CallbackData introduced in PyMsalRuntime 0.14
-    pymsalruntime>=0.14,<0.18; python_version>='3.6' and platform_system=='Windows'
+    pymsalruntime>=0.14,<0.19; python_version>='3.6' and platform_system=='Windows'
```
jiasli, line 65: I cannot use `pymsalruntime` 0.18.1 (https://github.com/Azure/azure-cli/pull/31563) right now as this line has not been released.

rayluo: True. Please subscribe/watch this repo to receive its release notification. We aim for each of the Azure CLI code freeze (minus 2 weeks) as checkpoint. Until then, you would have to manually install from MSAL's `dev` branch and test.

Issue #341, opened 2021-04-07
Inconsistent behavior between MSAL.PY and MSAL.NET while saving `organizations` tokens


jiasli:
## MSAL.PY

While saving ATs to cache, MSAL.PY gets `tenant_id` (`realm`) from `token_endpoint`:

https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/31b24afe3eb6edd3af58bab40f1387be02cb389d/msal/token_cache.py#L128

For `organizations`, `token_endpoint` is queried from

https://login.microsoftonline.com/organizations/v2.0/.well-known/openid-configuration

```
"token_endpoint":"https://login.microsoftonline.com/organizations/oauth2/v2.0/token"
```

MSAL.PY keeps the AT as `"realm": "organizations"`.

## MSAL.NET

On the other hand, MSAL.NET gets `tenantId` from `idToken` which will always be a GUID. 

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/3d9cb46d824820a580b7f826a71ecd5beb8131a8/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs#L46

```cs
            var tenantId = GetTenantId(idToken, requestParams);
```

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/3d9cb46d824820a580b7f826a71ecd5beb8131a8/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs#L65-L70

```cs
                msalAccessTokenCacheItem =
                    new MsalAccessTokenCacheItem(
                        ...
                        tenantId,
```

MSAL.NET keeps the AT as `"realm": "54826b22-38d6-4fb2-bad9-b7b93a3e9c5a"`.

bgavrilMS:
This aspect has been under-defined in MSAL libraries, but I think most MSALs will use the id token's `tid` claim to store tokens, because `organizations` (as well as `common`) is really a moniker for "I don't know the tenant, let AAD / the user figure it out". Once you get a token response from AAD, the tenant is known.

rayluo:
I intend to mark this issue as ByDesign/WontFix. And this is why.

[...]

PR #772, opened 2024-12-07
Supports dSTS by ClientApplication(..., authority="https://...example.com/dstsv2/...")

### LINKED ISSUE #767: [Bug] dSTS support
Please see https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/467 for details.

Configuration of dSTS authority should be the similar to configuration of AAD or ADFS authority, i.e. MSAL does some internal parsing. I think we should not do it CIAM style (WIthOidcAuthority) because Azure SDK will not work with that.

Note: dSTS is single tenanted for all intents and purposes.

rayluo:
Feature request #767 calls for support for dSTS authority.

The understanding was that:

* (a) The dSTS authority behaves similarly to an `oidc_authority` (i.e. no Instance Discovery, no "/v2.0" hardcoded string within its endpoint), so, it should just work if the caller would choose to put a dSTS url into the `oidc_authority` parameter, even if we might not advertise so.
* (b) Yet, we choose to allow using dSTS in the "traditional way", by accepting it via the `authority` parameter, so that MSAL's downstream ecosystem can support dSTS transparently without needing to pick up the `oidc_authority` parameter first.

Therefore, this PR attempts an implementation that simply converts the `authority=https://foo.bar.example.com/dstsv2/placeholder` into `oidc_authority=https://foo.bar.example.com/dstsv2/placeholder` under the hood, and then all the oidc authority behaviors will automatically kick in.

With regard to the tests:
* The test cases were refactored so that the new DstsAuthorityTestCase inherits all the previous OidcAuthorityTestCase (therefore testing point A above).
* A new test case was added to test the new behavior of `authority=https://foo.bar.example.com/dstsv2/placeholder` (therefore testing point B above)
* Currently, there is no new test cases mapping to the [Acceptance Tests described in this request](https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/467), because this PR builds up on top of the previously tested features (OIDC authority) and does not introduce new code in the token acquisition code path. In particular, `#2` doesn't apply because MSAL Python does not support `WithTenant(...)`. But I'm open to add more test cases if desirable.

This will resolve #767 

~P.S.: The test automation is currently failing due to other reason. They will be fixed soon outside of this PR. This situation does not prevent this PR from being reviewed.~ Test automation works again now.

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -6,6 +6,7 @@ import sys
 import warnings
 from threading import Lock
 from typing import Optional  # Needed in Python 3.7 & 3.8
+from urllib.parse import urlparse
 import os
 
 from .oauth2cli import Client, JwtAssertionCreator
@@ -622,6 +623,9 @@ class ClientApplication(object):
         # Here the self.authority will not be the same type as authority in input
         if oidc_authority and authority:
             raise ValueError("You can not provide both authority and oidc_authority")
+        if isinstance(authority, str) and urlparse(authority).path.startswith(
+            "/dstsv2"):  # dSTS authority's path always starts with "/dstsv2"
+            oidc_authority = authority  # So we treat it as if an oidc_authority
         try:
             authority_to_use = authority or "https://{}/common/".format(WORLD_WIDE)
             self.authority = Authority(
--- a/tests/test_authority.py
+++ b/tests/test_authority.py
@@ -104,31 +104,62 @@ class TestCiamAuthority(unittest.TestCase):
     "authorization_endpoint": "https://contoso.com/authorize",
     "token_endpoint": "https://contoso.com/token",
     })
-class TestOidcAuthority(unittest.TestCase):
+class OidcAuthorityTestCase(unittest.TestCase):
+    authority = "https://contoso.com/tenant"
+
+    def setUp(self):
+        # setUp() gives subclass a dynamic setup based on their authority
+        self.oidc_discovery_endpoint = (
+            # MSAL Python always does OIDC Discovery,
+            # not to be confused with Instance Discovery
+            # Here the test is to confirm the OIDC endpoint contains no "/v2.0"
+            self.authority + "/.well-known/openid-configuration")
+
     def test_authority_obj_should_do_oidc_discovery_and_skip_instance_discovery(
             self, oidc_discovery, instance_discovery):
         c = MinimalHttpClient()
-        a = Authority(None, c, oidc_authority_url="https://contoso.com/tenant")
+        a = Authority(None, c, oidc_authority_url=self.authority)
         instance_discovery.assert_not_called()
-        oidc_discovery.assert_called_once_with(
-            "https://contoso.com/tenant/.well-known/openid-configuration", c)
+        oidc_discovery.assert_called_once_with(self.oidc_discovery_endpoint, c)
         self.assertEqual(a.authorization_endpoint, 'https://contoso.com/authorize')
         self.assertEqual(a.token_endpoint, 'https://contoso.com/token')
 
     def test_application_obj_should_do_oidc_discovery_and_skip_instance_discovery(
             self, oidc_discovery, instance_discovery):
         app = msal.ClientApplication(
-            "id",
-            authority=None,
-            oidc_authority="https://contoso.com/tenant",
-            )
+            "id", authority=None, oidc_authority=self.authority)
         instance_discovery.assert_not_called()
         oidc_discovery.assert_called_once_with(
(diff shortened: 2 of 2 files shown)


msal/application.py
```diff
@@ -623,6 +624,9 @@ def __init__(
         # Here the self.authority will not be the same type as authority in input
         if oidc_authority and authority:
             raise ValueError("You can not provide both authority and oidc_authority")
+        if isinstance(authority, str) and urlparse(authority).path.startswith(
+            "/dstsv2"):  # dSTS authority's path always starts with "/dstsv2"
+            oidc_authority = authority  # So we treat it as if an oidc_authority
```
bgavrilMS, line 628: +1

I am wondering though how does MSAL Py handle changing the tenant at the request level? I am asking because Azure SDK does this and Azure SDK wants to remain agnostic of authority type (i.e. they want MSAL to just work with AAD, ADFS, dSTS)

bgavrilMS: The simplest is to have `request.WithTenantId(tid)` do nothing in case the authority is dSTS. 

Needs a unit test.

rayluo: [MSAL Python does not currently support `request.WithTenant(...)`](https://msal-python.readthedocs.io/en/latest/#msal.ConfidentialClientApplication.acquire_token_for_client) at all. Confirmed with a downstream partner that they are not currently using it either. It was tracked as a feature request, which we can work on.

Do we want to proceed with this PR as-is, or mark it as blocked while we pick up the `WithTenant(...)` feature now?

bgavrilMS (approved):

tests/test_authority.py
```diff
@@ -104,32 +104,63 @@ def test_authority_with_path_should_be_used_as_is(self, oidc_discovery):
...
+            self.oidc_discovery_endpoint, app.http_client)
         self.assertEqual(
             app.authority.authorization_endpoint, 'https://contoso.com/authorize')
         self.assertEqual(app.authority.token_endpoint, 'https://contoso.com/token')
 
+
+class DstsAuthorityTestCase(OidcAuthorityTestCase):
+    # Inherits OidcAuthority's test cases and run them with a dSTS authority
+    authority = (  # dSTS is single tenanted with a dummy tenant placeholder
+        'https://test-instance1-dsts.dsts.core.azure-test.net/dstsv2/tid')
```
neha-bhargava: Can we also add a test case with common? I think with dsts there can be a tenant id or common used in the authority.

neha-bhargava (approved):
Approve with comment to add another test for common as tenant

tests/test_authority.py
bgavrilMS: I'm not really sure how _common_ is "common" :) and user discovery in dsts. It may be worth asking Sai before.

rayluo: > I'm not really sure how _common_ is "common" :) and user discovery in dsts. It may be worth asking Sai before.

You meant OIDC discovery? This is the [OIDC discovery for common](https://test-instance1-dsts.dsts.core.azure-test.net/dstsv2/common/.well-known/openid-configuration) and this is for [the non-common](https://test-instance1-dsts.dsts.core.azure-test.net/dstsv2/contoso/.well-known/openid-configuration). I'll get back to you when I hear back from Sai.

bgavrilMS: Right, "common" is supported but doesn't have the same meaning and eSTS "common", since dSTS is single tenanted. So to me it doesn't make sense to inherit all of eSTS complex logic around multi-tenancy, such as replacing "common" with the id token's `tid` in the cache etc. 

A test is fine.

rayluo: > You meant OIDC discovery? This is the [OIDC discovery for common](https://test-instance1-dsts.dsts.core.azure-test.net/dstsv2/common/.well-known/openid-configuration) and this is for [the non-common](https://test-instance1-dsts.dsts.core.azure-test.net/dstsv2/contoso/.well-known/openid-configuration). I'll get back to you when I hear back from Sai.

Confirmed with Sai that the seemingly two sets of authorities, common and un-common, are just a side effect of reusing some eSTS code. Under the hood, that common endpoint will still convert the "common" into the tenant id and mint a token for that tenant id.

> So to me it doesn't make sense to inherit all of eSTS complex logic around multi-tenancy, such as replacing "common" with the id token's `tid` in the cache etc.

MSAL Python's [tenant agnostic approach](https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/341#issuecomment-1400936207) does not attempt converting "common" into tid.

PR #785, opened 2025-01-25
Scope to resource

### LINKED ISSUE #784: [Bug] `acquire_token_interactive` fails in Cloud Shell: Audience GUID/.default is not a supported MSI token audience.
### Describe the bug

`acquire_token_interactive` fails in Cloud Shell: Audience GUID/.default is not a supported MSI token audience.

As an example, we use ARM's app ID `797f4846-ba00-4fd7-ba43-dac1f8f63013` (can be retrieved by `az ad sp show --id "https://management.azure.com/"`). MSAL can't get an access token for it:

```sh
$ az account get-access-token --resource "797f4846-ba00-4fd7-ba43-dac1f8f63013"
A Cloud Shell credential problem occurred. When you report the issue with the error below, please mention the hostname 'SandboxHost-638732994281562992'
Audience 797f4846-ba00-4fd7-ba43-dac1f8f63013/.default is not a supported MSI token audience.
```

Azure CLI passes `797f4846-ba00-4fd7-ba43-dac1f8f63013/.default` as `scope` to MSAL, but due to the incorrect `msal.cloudshell._scope_to_resource` logic:

https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/3f3d1336751a19cb8a7c0575e68ffbea66581008/msal/cloudshell.py#L26-L37

the `/.default` suffix is not removed. **This is because only when `scope` is a URL, the `/.default` suffix is removed. For GUID `scope`, the `/.default` suffix is preserved.**

Here is a comparision of Azure CLI and MSAL's scope-to-resource conversion logic:

```py
from azure.cli.core.auth.util import scopes_to_resource as cli_scopes_to_resource
from msal.cloudshell import _scope_to_resource as msal_scope_to_resource

guid_scope = '797f4846-ba00-4fd7-ba43-dac1f8f63013/.default'
print(cli_scopes_to_resource([guid_scope]))
print(msal_scope_to_resource(guid_scope))

url_scope = 'https://management.azure.com//.default'
print(cli_scopes_to_resource([url_scope]))
print(msal_scope_to_resource(url_scope))
```

Output:

```
797f4846-ba00-4fd7-ba43-dac1f8f63013
797f4846-ba00-4fd7-ba43-dac1f8f63013/.default
https://management.azure.com/
https://management.azure.com
```

### Another issue: Incorrect handling of trailing slash

MSAL's scope-to-resource conversion logic has another issue: When `scope` is `https://management.azure.com//.default`, MSAL removes not only `/.default`, but also the trailing slash `/`, resulting in `https://management.azure.com`. This will also trigger failure for resources that require a trailing slash. See https://learn.microsoft.com/en-us/entra/identity-platform/scopes-oidc#trailing-slash-and-default

Azure CLI made a similar mistake before and fixed it in https://github.com/Azure/azure-cli/pull/15698.

### Azure CLI's implementation

The source code of `azure.cli.core.auth.util.scopes_to_resource`:

https://github.com/Azure/azure-cli/blob/464a79cb0f9c474da1f9d426b9aaf56afcffc47a/src/azure-cli-core/azure/cli/core/auth/util.py#L87-L106

```py
def scopes_to_resource(scopes):
    """Convert MSAL scopes to ADAL resource by stripping the /.default suffix and return a str.
    For example:
       ['https://management.core.windows.net//.default'] -> 'https://management.core.windows.net/'
       ['https://managedhsm.azure.com/.default'] -> 'https://managedhsm.azure.com'

    :param scopes: The MSAL scopes. It can be a list or tuple of string
    :return: The ADAL resource
    :rtype: str
    """
    if not scopes:
        return None

    scope = scopes[0]
    suffixes = ['/.default', '/user_impersonation']
    for s in suffixes:
        if scope.endswith(s):
            return scope[:-len(s)]

    return scope
```

**Instead of detecting whether `scope` is a URL, it detects if the `scope` ends with `/.default`. If so, remove the `/.default` suffix.**

rayluo:
This PR fixes the #784 

As an implementation detail, this PR does not find and remove a magic string `/.default`. Instead, it still attempts to semantically strip the "path(s)" and keep the "domain" or "guid". Regardless of such detail, please focus on reviewing the test cases.

CC @jiasli : Can you test from this PR? Please give me a reply and then I will merge it and ship it.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -25,10 +25,10 @@ jobs:
       LAB_OBO_PUBLIC_CLIENT_ID: ${{ secrets.LAB_OBO_PUBLIC_CLIENT_ID }}
 
     # Derived from https://docs.github.com/en/actions/guides/building-and-testing-python#starting-with-the-python-workflow-template
-    runs-on: ubuntu-latest  # It switched to 22.04 shortly after 2022-Nov-8
+    runs-on: ubuntu-22.04
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9, "3.10", "3.11", "3.12"]
+        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11', '3.12', '3.13']
 
     steps:
     - uses: actions/checkout@v4
--- a/msal/cloudshell.py
+++ b/msal/cloudshell.py
@@ -32,8 +32,12 @@ def _scope_to_resource(scope):  # This is an experimental reasonable-effort appr
         if scope.startswith(a):
             return a
     u = urlparse(scope)
+    if not u.scheme and not u.netloc:  # Typically the "GUID/scope" case
+        return u.path.split("/")[0]
     if u.scheme:
-        return "{}://{}".format(u.scheme, u.netloc)
+        trailer = (  # https://learn.microsoft.com/en-us/entra/identity-platform/scopes-oidc#trailing-slash-and-default
+            "/" if u.path.startswith("//") else "")
+        return "{}://{}{}".format(u.scheme, u.netloc, trailer)
     return scope  # There is no much else we can do here
 
 
--- a/tests/test_cloudshell.py
+++ b/tests/test_cloudshell.py
@@ -1,0 +1,21 @@
+import unittest
+from msal.cloudshell import _scope_to_resource
+
+class TestScopeToResource(unittest.TestCase):
+
+    def test_expected_behaviors(self):
+        for scope, expected_resource in {
+            "https://analysis.windows.net/powerbi/api/foo":
+                "https://analysis.windows.net/powerbi/api",  # A special case
+            "https://pas.windows.net/CheckMyAccess/Linux/.default":
+                "https://pas.windows.net/CheckMyAccess/Linux/.default",  # Special case
+            "https://double-slash.com//scope": "https://double-slash.com/",
+            "https://single-slash.com/scope": "https://single-slash.com",
+            "guid/some/scope": "guid",
+            "797f4846-ba00-4fd7-ba43-dac1f8f63013/.default":  # Realistic GUID
+                "797f4846-ba00-4fd7-ba43-dac1f8f63013"
+        }.items():
+            self.assertEqual(_scope_to_resource(scope), expected_resource)
+
+if __name__ == '__main__':
+    unittest.main()
--- a/tests/test_cloudshell.py
+++ b/tests/test_cloudshell.py
@@ -12,8 +12,10 @@ class TestScopeToResource(unittest.TestCase):
             "https://double-slash.com//scope": "https://double-slash.com/",
             "https://single-slash.com/scope": "https://single-slash.com",
             "guid/some/scope": "guid",
-            "797f4846-ba00-4fd7-ba43-dac1f8f63013/.default":  # Realistic GUID
-                "797f4846-ba00-4fd7-ba43-dac1f8f63013"
+            "6dae42f8-4368-4678-94ff-3960e28e3630/.default":
+                # The real guid of AKS resource
+                # https://learn.microsoft.com/en-us/azure/aks/kubelogin-authentication#how-to-use-kubelogin-with-aks
+                "6dae42f8-4368-4678-94ff-3960e28e3630",
         }.items():
             self.assertEqual(_scope_to_resource(scope), expected_resource)
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -52,7 +52,7 @@ install_requires =
     # And we will use the cryptography (X+3).0.0 as the upper bound,
     # based on their latest deprecation policy
     # https://cryptography.io/en/latest/api-stability/#deprecation
-    cryptography>=2.5,<46
+    cryptography>=2.5,<47
 
 
 [options.extras_require]


msal/cloudshell.py
```diff
@@ -32,8 +32,12 @@ def _scope_to_resource(scope):  # This is an experimental reasonable-effort appr
         if scope.startswith(a):
             return a
     u = urlparse(scope)
+    if not u.scheme and not u.netloc:  # Typically the "GUID/scope" case
+        return u.path.split("/")[0]
     if u.scheme:
-        return "{}://{}".format(u.scheme, u.netloc)
+        trailer = (  # https://learn.microsoft.com/en-us/entra/identity-platform/scopes-oidc#trailing-slash-and-default
+            "/" if u.path.startswith("//") else "")
+        return "{}://{}{}".format(u.scheme, u.netloc, trailer)
```
jiasli, line 40: If `resource` is provided to Azure CLI, Azure CLI just adds a trailing `/.default` to the `resource`:

https://github.com/Azure/azure-cli/blob/967983c749c70e5e7dec5f1f84dae9032dc1008d/src/azure-cli-core/azure/cli/core/_profile.py#L451

```py
            scopes = resource_to_scopes(resource)
```

https://github.com/Azure/azure-cli/blob/464a79cb0f9c474da1f9d426b9aaf56afcffc47a/src/azure-cli-core/azure/cli/core/auth/util.py#L83

```py
    scope = resource + '/.default'
```

So this scope-to-resource conversion is still not exactly the reverse of Azure CLI's process.

jiasli: There is an edge case where the `resource` contains more than 2 trailing slashes. 

Imagine `resource` is `https://myresource.com//`, CLI will convert it to `scope` as `https://myresource.com///.default`. The converted-back `resource` will be `https://myresource.com/`, losing the last slash.

jiasli: `https://myresource.com/mypath` will also be converted to `https://myresource.com`.

jiasli: This conversion logic will remove the path part.

As mentioned in https://github.com/Azure/azure-cli/pull/18312, valid identifier URI can also be

- `https://{verifiedCustomerDomain}/{string}`
- `https://{string}.{verifiedCustomerDomain}/{string}`

The trailing `/{string}` will be removed.

rayluo: > ...
> So this scope-to-resource conversion is still not exactly the reverse of Azure CLI's process.

Not that the current `msal.cloudshell._scope_to_resource()` conversion algorithm is perfect, but logically speaking, the goal is not reversing Azure CLI's process, but to match [Cloud Shell's quirk](https://msazure.visualstudio.com/One/_git/compute-CloudShell?path=/src/images/agent/env/envconfig.PROD.json).

> Imagine resource is `https://myresource.com//` ...

Theoretically, yes, but we never saw that before. And, most importantly, the "Cloud Shell's quirk" above contains several resources with one trailing slash, but never a double (or more) trailing slash.

> `https://myresource.com/mypath` will also be converted to `https://myresource.com`.

That is arguably an expected behavior. Again, we have never encountered a Cloud Shell resource containing path, except the [two known outliners which have already been handled properly](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/785/files#diff-f51764d044bb4e96883a8be8693858a026c79464a879fb3024a61dd8a509f09bR27-R33).

I will suggest we focus on [this test case](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/785/files#diff-c8075256c9d5927a76b8c2f7043c3d665de3e2693b3fc38dfa5ec731dc07288fR14) and verify its solving the "GUID/scope" use case.

tests/test_cloudshell.py
```diff
@@ -0,0 +1,19 @@
...
+
+    def test_expected_behaviors(self):
+        for scope, expected_resource in {
+            "https://analysis.windows.net/powerbi/api/foo":
+                "https://analysis.windows.net/powerbi/api",  # A special case
+            "https://pas.windows.net/CheckMyAccess/Linux/.default":
+                "https://pas.windows.net/CheckMyAccess/Linux/.default",  # Special case
+            "https://double-slash.com//scope": "https://double-slash.com/",
+            "https://single-slash.com/scope": "https://single-slash.com",
+            "guid/some/scope": "guid",
```
jiasli, line 14: Maybe using a real GUID to make it more reliable?

For example, add `797f4846-ba00-4fd7-ba43-dac1f8f63013/.default` -> `797f4846-ba00-4fd7-ba43-dac1f8f63013`.

```suggestion
            "guid/some/scope": "guid",
            "797f4846-ba00-4fd7-ba43-dac1f8f63013/.default": "797f4846-ba00-4fd7-ba43-dac1f8f63013"
```

.github/workflows/python-package.yml
```diff
@@ -25,10 +25,10 @@ jobs:
       LAB_OBO_PUBLIC_CLIENT_ID: ${{ secrets.LAB_OBO_PUBLIC_CLIENT_ID }}
 
     # Derived from https://docs.github.com/en/actions/guides/building-and-testing-python#starting-with-the-python-workflow-template
-    runs-on: ubuntu-latest  # It switched to 22.04 shortly after 2022-Nov-8
+    runs-on: ubuntu-22.04  # ubuntu-latest switched to 22.04 shortly after 2022-Nov-8
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9, "3.10", "3.11", "3.12"]
+        python-version: [3.7, 3.8, 3.9, "3.10", "3.11", "3.12", 3.13]
```
jiasli: Any reason why `"3.10"`, `"3.11"` and `"3.12"` are quoted?

tests/test_cloudshell.py
```diff
@@ -0,0 +1,19 @@
...
+        }.items():
+            self.assertEqual(_scope_to_resource(scope), expected_resource)
+
+if __name__ == '__main__':
+    unittest.main()
```
jiasli: Missing an ending new line.

.github/workflows/python-package.yml
```diff
@@ -25,10 +25,10 @@ jobs:
       LAB_OBO_PUBLIC_CLIENT_ID: ${{ secrets.LAB_OBO_PUBLIC_CLIENT_ID }}
 
     # Derived from https://docs.github.com/en/actions/guides/building-and-testing-python#starting-with-the-python-workflow-template
-    runs-on: ubuntu-latest  # It switched to 22.04 shortly after 2022-Nov-8
+    runs-on: ubuntu-22.04  # ubuntu-latest switched to 22.04 shortly after 2022-Nov-8
```
jiasli: https://github.com/actions/runner-images
https://github.blog/changelog/2024-09-25-actions-new-images-and-ubuntu-latest-changes/

```suggestion
    runs-on: ubuntu-22.04  # ubuntu-latest switched to 24.04 shortly after 2024-09-23
```

.github/workflows/python-package.yml
rayluo: If I remember correctly, "3.10" needs to be quoted otherwise it would be interpreted as `3.1`. The 3.11 and 3.12 did not have to be quoted, but it was somehow committed that way long ago.

.github/workflows/python-package.yml
rayluo: Good catch. That was an outdated comment that can actually be removed. Now updated.

.github/workflows/python-package.yml
jiasli: The example of `setup-python` actually quotes all `python-version`s: https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#matrix-testing

```
      matrix:
        python-version: ['3.x', 'pypy2.7', 'pypy3.8', 'pypy3.9' ]
```

jiasli: Did a quick experiment:

```py
from yaml import load, CLoader
r = load('python-version: [3.7, 3.8, 3.9, "3.10", "3.11", "3.12", 3.13]', Loader=CLoader)
print(r)

# output
{'python-version': [3.7, 3.8, 3.9, '3.10', '3.11', '3.12', 3.13]}
```

Without quoting, the version number is treated as a float. That's why `3.10` becomes `3.1`. All version should be quoted to become a string.

.github/workflows/python-package.yml
```diff
@@ -25,10 +25,10 @@ jobs:
       LAB_OBO_PUBLIC_CLIENT_ID: ${{ secrets.LAB_OBO_PUBLIC_CLIENT_ID }}
 
     # Derived from https://docs.github.com/en/actions/guides/building-and-testing-python#starting-with-the-python-workflow-template
-    runs-on: ubuntu-latest  # It switched to 22.04 shortly after 2022-Nov-8
+    runs-on: ubuntu-22.04
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9, "3.10", "3.11", "3.12"]
+        python-version: [3.7, 3.8, 3.9, "3.10", 3.11, 3.12, 3.13]
```
jiasli:
```suggestion
        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11', '3.12', '3.13']
```

tests/test_cloudshell.py
```diff
@@ -0,0 +1,21 @@
...
+        for scope, expected_resource in {
+            "https://analysis.windows.net/powerbi/api/foo":
+                "https://analysis.windows.net/powerbi/api",  # A special case
+            "https://pas.windows.net/CheckMyAccess/Linux/.default":
+                "https://pas.windows.net/CheckMyAccess/Linux/.default",  # Special case
+            "https://double-slash.com//scope": "https://double-slash.com/",
+            "https://single-slash.com/scope": "https://single-slash.com",
+            "guid/some/scope": "guid",
+            "797f4846-ba00-4fd7-ba43-dac1f8f63013/.default":  # Realistic GUID
+                "797f4846-ba00-4fd7-ba43-dac1f8f63013"
```
jiasli, line 16: https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/785/files#r1929651794 is my fault. Even though ARM's app ID can be used to get an access token locally, it fails in Cloud Shell:

> Audience 797f4846-ba00-4fd7-ba43-dac1f8f63013 is not a supported MSI token audience.

jiasli: https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/785#discussion_r1929651794 is my fault. Even though ARM's app ID can be used to get an access token locally, it fails in Cloud Shell:

> Audience 797f4846-ba00-4fd7-ba43-dac1f8f63013 is not a supported MSI token audience.

`6dae42f8-4368-4678-94ff-3960e28e3630` (`https://aks-aad-server.azure.com`) works, but `e406a681-f3d4-42a8-90b6-c2b029497af1` (`https://storage.azure.com`) results in error

> Timeout waiting for token from portal. Audience: e406a681-f3d4-42a8-90b6-c2b029497af1

jiasli: https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/785#discussion_r1929651794 is my fault. Even though ARM's app ID can be used to get an access token locally, it fails in Cloud Shell:

> Audience 797f4846-ba00-4fd7-ba43-dac1f8f63013 is not a supported MSI token audience.

`6dae42f8-4368-4678-94ff-3960e28e3630` (`https://aks-aad-server.azure.com`) works, but `e406a681-f3d4-42a8-90b6-c2b029497af1` (`https://storage.azure.com`) results in error

> Timeout waiting for token from portal. Audience: e406a681-f3d4-42a8-90b6-c2b029497af1

jiasli: `6dae42f8-4368-4678-94ff-3960e28e3630` is officially documented: https://learn.microsoft.com/en-us/azure/aks/kubelogin-authentication#how-to-use-kubelogin-with-aks

```suggestion
            "6dae42f8-4368-4678-94ff-3960e28e3630/.default":  # Realistic GUID
                "6dae42f8-4368-4678-94ff-3960e28e3630"
```

rayluo: The current implementation actually works for this kind of pattern regardless of its guid. But, OK, the PR has now been changed to use that real GUID.

tests/test_cloudshell.py
```diff
@@ -0,0 +1,23 @@
+import unittest
+from msal.cloudshell import _scope_to_resource
+
+class TestScopeToResource(unittest.TestCase):
+
+    def test_expected_behaviors(self):
+        for scope, expected_resource in {
+            "https://analysis.windows.net/powerbi/api/foo":
+                "https://analysis.windows.net/powerbi/api",  # A special case
```
bgavrilMS, line 9: Hmm, this isn't correct, you should only be stripped /.default ?

msal/cloudshell.py
bgavrilMS: There seem to be a lot of special cases here. Will there be other customers for this code path?

tests/test_cloudshell.py
```diff
@@ -0,0 +1,23 @@
...
+            "https://pas.windows.net/CheckMyAccess/Linux/.default":
+                "https://pas.windows.net/CheckMyAccess/Linux/.default",  # Special case
+            "https://double-slash.com//scope": "https://double-slash.com/",
+            "https://single-slash.com/scope": "https://single-slash.com",
+            "guid/some/scope": "guid",
```
bgavrilMS, line 14: Is `guid` really the answer here? I think you should only strip away `.\default`

bgavrilMS (approved):

msal/cloudshell.py
```diff
@@ -32,8 +32,12 @@ def _scope_to_resource(scope):  # This is an experimental reasonable-effort appr
         if scope.startswith(a):
             return a
     u = urlparse(scope)
+    if not u.scheme and not u.netloc:  # Typically the "GUID/scope" case
```
bgavrilMS, line 35: I don't agree with this logic. It can potentially strip important scope information and be seen as a security issue. 

Since this is managed identity, we only deal with app scopes and there is 1:1 mapping between app scope and resource. i.e. scope = resource + "./default"

As such this logic should take it into account - i.e. strip "./default" if it exists. It should not attempt to remove anything else.

bgavrilMS: This is the logic that MSAL.NET uses https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main//src/client/Microsoft.Identity.Client/Utils/ScopeHelper.cs#L106

bgavrilMS (changes requested):
Restrict the logic to strings ending in "./default"

tests/test_cloudshell.py
rayluo: Semantically, since the Cloud Shell's token endpoint is expecting a resource, so we consider the resource of a potential "GUID/path/subpath" input to be only the leading part i.e. the "GUID".

Ideally, Cloud Shell will support scope directly in the future and then we will be able to remove all these workarounds.

msal/cloudshell.py
rayluo: We discussed offline. The code path is for Azure CLI running inside Cloud Shell to get a token for the signed-in user. It is never about a managed machine. (That is also why it is misleading to expose Cloud Shell token acquisition as part of the managed identity.)

Currently, Cloud Shell's token endpoint accepts a resource parameter which shall not contain any scope; Cloud Shell also has special case that accepts "https://pas.windows.net/CheckMyAccess/Linux/<b>.default</b>" as-is. That .Net logic will not work here.

msal/cloudshell.py
rayluo: No other customer uses this code path. It is for Cloud Shell only, and not for any other managed identity.

rayluo:
Per the agreement in a meeting today, this PR is now shipped as [MSAL Python 1.31.2b1](https://github.com/AzureAD/microsoft-authentication-library-for-python/releases/tag/1.31.2b1) to meet a release schedule. Another round of discussion with broader audience will be scheduled in the near future; further changes can be made accordingly.

bgavrilMS (approved):

PR #786, opened 2025-01-30
Pass Sku and Ver to MsalRuntime

Ugonnaak1:
Passes Msal.py sku and version to Msalcpp during runtime flows to be used in x-client-xtra-sku header in order to support ESTS telemetry

[AB#3138694](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3138694)

### DIFF
--- a/msal/broker.py
+++ b/msal/broker.py
@@ -9,10 +9,7 @@ import uuid
 
 from .sku import __version__, SKU
 
-
 logger = logging.getLogger(__name__)
-
-
 try:
     import pymsalruntime  # Its API description is available in site-packages/pymsalruntime/PyMsalRuntime.pyi
     pymsalruntime.register_logging_callback(lambda message, level: {  # New in pymsalruntime 0.7
--- a/msal/broker.py
+++ b/msal/broker.py
@@ -9,6 +9,10 @@ import uuid
 
 
 logger = logging.getLogger(__name__)
+
+from .application import (
+    __version__)
+
 try:
     import pymsalruntime  # Its API description is available in site-packages/pymsalruntime/PyMsalRuntime.pyi
     pymsalruntime.register_logging_callback(lambda message, level: {  # New in pymsalruntime 0.7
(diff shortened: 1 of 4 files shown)


msal/broker.py
```diff
@@ -9,6 +9,10 @@
 
 
 logger = logging.getLogger(__name__)
+
+from .application import (
+    __version__)
+
```
rayluo: [In Python, the convention](https://peps.python.org/pep-0008/#imports) is to group all imports at the beginning of the file.

And, now that you ended up duplicating the sku magic string, so you should have go ahead and define a const in application.py for this change.

```suggestion

from .application import __version__, SKU


logger = logging.getLogger(__name__)


```

msal/broker.py
```diff
@@ -135,6 +139,9 @@ def _get_new_correlation_id():
 def _enable_msa_pt(params):
     params.set_additional_parameter("msal_request_type", "consumer_passthrough")  # PyMsalRuntime 0.8+
 
+def _pass_client_sku(params):
+    params.set_additional_parameter("msal_client_sku", "MSAL.Python")  
+    params.set_additional_parameter("msal_client_ver", __version__)
```
rayluo:
```suggestion
def _build_msal_runtime_auth_params(client, authority):
    params = pymsalruntime.MSALRuntimeAuthParameters(client_id, authority)
    params.set_additional_parameter("msal_client_sku", SKU)
    params.set_additional_parameter("msal_client_ver", __version__)
    return params
```

And then change all the `params = pymsalruntime.MSALRuntimeAuthParameters(client_id, authority)` below into `params = _build_msal_runtime_auth_params(client, authority)`, and remove all the `_pass_client_sku(params)`.

This way, it will be less likely to forget adding sku, should we need to add more Runtime flows in the future.

msal/broker.py
```diff
@@ -7,8 +7,12 @@
 import time
 import uuid
 
+from .application import __version__
```
rayluo: Better add a new sku.py which contains nothing but `SKU="..."` and `__version__ = "..."`, and then we have both application.py and broker.py (plus `__init__.py`) do `from .sku import ...`. This way, the application.py and broker.py will not have a circular reference to each other.

rayluo (approved):

PR #790, opened 2025-02-10
Enable Issue-Sentinel to scan for similar issues

DharshanBJ:
Enable Issue-Sentinel to scan for similar issues
Onboarding reference document - https://github.com/Azure/issue-Sentinel?tab=readme-ov-file


rayluo (approved):
The change is identical to the [canonical source](https://github.com/Azure/issue-Sentinel?tab=readme-ov-file), so, approving in advance. However, don't we need to also onboard and setup an env var here?

PR #791, opened 2025-02-12
ManagedIdentityClient(..., client_capabilities=["cp1"]).acquire_token_for_client(..., claims_challenge="...")

rayluo:
~This pull request adds the `access_token_sha256_to_refresh` parameter to the `acquire_token_for_client` method in the `msal.ConfidentialClientApplication` class.~

~This PR meets [feature requirements and acceptance criteria](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/5111).~

This PR supports `ManagedIdentityClient(..., client_capabilities=["cp1"]).acquire_token_for_client(..., claims_challenge="...")`.

Currently, the new behavior will be available for Service Fabric. Other Managed Identity v1 providers may pick up those new behaviors soon. *After* that happens, we will enable this behavior for more Managed Identity code paths.

### DIFF
--- a/msal/managed_identity.py
+++ b/msal/managed_identity.py
@@ -2,6 +2,7 @@
 # All rights reserved.
 #
 # This code is licensed under the MIT License.
+import hashlib
 import json
 import logging
 import os
@@ -10,7 +11,7 @@ import sys
 import time
 from urllib.parse import urlparse  # Python 3+
 from collections import UserDict  # Python 3+
-from typing import Optional, Union  # Needed in Python 3.7 & 3.8
+from typing import List, Optional, Union  # Needed in Python 3.7 & 3.8
 from .token_cache import TokenCache
 from .individual_cache import _IndividualCache as IndividualCache
 from .throttled_http_client import ThrottledHttpClientBase, RetryAfterParser
@@ -162,6 +163,7 @@ class ManagedIdentityClient(object):
         http_client,
         token_cache=None,
         http_cache=None,
+        client_capabilities: Optional[List[str]] = None,
     ):
         """Create a managed identity client.
 
@@ -191,6 +193,17 @@ class ManagedIdentityClient(object):
         :param http_cache:
             Optional. It has the same characteristics as the
             :paramref:`msal.ClientApplication.http_cache`.
+
+        :param list[str] client_capabilities: (optional)
+            Allows configuration of one or more client capabilities, e.g. ["CP1"].
+
+            Client capability is meant to inform the Microsoft identity platform
+            (STS) what this client is capable for,
+            so STS can decide to turn on certain features.
+
+            Implementation details:
+            Client capability in Managed Identity is relayed as-is
+            via ``xms_cc`` parameter on the wire.
 
         Recipe 1: Hard code a managed identity for your app::
 
@@ -238,6 +251,7 @@ class ManagedIdentityClient(object):
             http_cache=http_cache,
         )
         self._token_cache = token_cache or TokenCache()
+        self._client_capabilities = client_capabilities
 
     def _get_instance(self):
         if self.__instance is None:
@@ -266,8 +280,7 @@ class ManagedIdentityClient(object):
             and then a *claims challenge* will be returned by the target resource,
             as a `claims_challenge` directive in the `www-authenticate` header,
             even if the app developer did not opt in for the "CP1" client capability.
-            Upon receiving a `claims_challenge`, MSAL will skip a token cache read,
-            and will attempt to acquire a new token.
+            Upon receiving a `claims_challenge`, MSAL will attempt to acquire a new token.
 
         .. note::
 
@@ -278,11 +291,13 @@ class ManagedIdentityClient(object):
             This is a service-side behavior that cannot be changed by this library.
             `Azure VM docs <https://learn.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#get-a-token-using-http>`_
         """
+        access_token_to_refresh = None  # This could become a public parameter in the future
         access_token_from_cache = None
         client_id_in_cache = self._managed_identity.get(
             ManagedIdentity.ID, "SYSTEM_ASSIGNED_MANAGED_IDENTITY")
         now = time.time()
-        if not claims_challenge:  # Then attempt token cache search
+        if True:  # Attempt cache search even if receiving claims_challenge,
+                  # because we want to locate the existing token (if any) and refresh it
             matches = self._token_cache.find(
                 self._token_cache.CredentialType.ACCESS_TOKEN,
                 target=[resource],
@@ -297,6 +312,11 @@ class ManagedIdentityClient(object):
                 expires_in = int(entry["expires_on"]) - now
                 if expires_in < 5*60:  # Then consider it expired
                     continue  # Removal is not necessary, it will be overwritten
+                if claims_challenge and not access_token_to_refresh:
+                    # Since caller did not pinpoint the token causing claims challenge,
+                    # we have to assume it is the first token we found in cache.
+                    access_token_to_refresh = entry["secret"]
+                    break
                 logger.debug("Cache hit an AT")
                 access_token_from_cache = {  # Mimic a real response
                     "access_token": entry["secret"],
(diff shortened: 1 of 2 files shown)


msal/managed_identity.py
```diff
@@ -402,6 +432,8 @@ def _obtain_token(http_client, managed_identity, resource):
             os.environ["IDENTITY_HEADER"],
             os.environ["IDENTITY_SERVER_THUMBPRINT"],
             resource,
+            access_token_sha256_to_refresh=access_token_sha256_to_refresh,
```
gladjohn, line 435:
```suggestion
            token_sha256_to_refresh=access_token_sha256_to_refresh,
```

@rayluo according to the [spec](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/docs/msiv1_token_revocation.md) it is `token_sha256_to_refresh`

msal/managed_identity.py
```diff
@@ -563,7 +598,12 @@ def _obtain_token_on_service_fabric(
     logger.debug("Obtaining token via managed identity on Azure Service Fabric")
     resp = http_client.get(
         endpoint,
-        params={"api-version": "2019-07-01-preview", "resource": resource},
+        params={k: v for k, v in {
+            "api-version": "2019-07-01-preview",
+            "resource": resource,
+            "token_sha256_to_refresh": access_token_sha256_to_refresh,
+            "xms_cc": ",".join(client_capabilities) if client_capabilities else None,
```
gladjohn, line 605: Yes, this will be passed irrelevant of claims exist or not. thanks.

msal/managed_identity.py
```diff
@@ -563,7 +598,12 @@ def _obtain_token_on_service_fabric(
     logger.debug("Obtaining token via managed identity on Azure Service Fabric")
     resp = http_client.get(
         endpoint,
-        params={"api-version": "2019-07-01-preview", "resource": resource},
+        params={k: v for k, v in {
+            "api-version": "2019-07-01-preview",
```
gladjohn, line 602:
```suggestion
            "api-version": "2019-07-01-preview",
```
We just need a final confirmation on this one @rayluo. App Service is setting a new api-version, we want a sign off from SF

tests/test_mi.py
```diff
@@ -245,18 +265,40 @@ def test_machine_learning_error_should_be_normalized(self):
...
+
+    def test_happy_path_with_claim_challenge_should_send_sha256_to_provider(self):
+        claims_challenge='{"access_token": {"nbf": {"essential": true, "value": "1563308371"}}}'
+        self._test_happy_path(self.app, claims_challenge=claims_challenge).assert_called_with(
+            'http://localhost',
+            params={
+                'api-version': '2019-07-01-preview',
+                'resource': 'R',
+                'token_sha256_to_refresh': hashlib.sha256(b"AT").hexdigest(),
+            },
```
gladjohn, line 304: do we need a test of claims + forceRefresh?

tests/test_mi.py
```diff
@@ -245,18 +265,40 @@ def test_machine_learning_error_should_be_normalized(self):
...
+            headers={'Secret': 'foo'},
+        )
```
gladjohn, line 306: can we also add a test case specifically injecting a mismatched hash to verify how the code handles that scenario?

tests/test_mi.py
```diff
@@ -245,18 +265,40 @@ def test_machine_learning_error_should_be_normalized(self):
...
                 int(time.time()) + expires_in),
         )) as mocked_method:
             super(ServiceFabricTestCase, self)._test_happy_path(
-                app, mocked_method, expires_in)
+                app, mocked_method, expires_in, claims_challenge=claims_challenge)
+            return mocked_method
 
-    def test_happy_path(self):
-        self._test_happy_path(self.app)
+    def test_happy_path_with_client_capabilities_should_relay_capabilities(self):
```
gladjohn, line 281: can you also pass an empty list client_capabilities=[], and confirm that xms_cc is omitted? did not see one.

msal/managed_identity.py
rayluo: What kind of sign-off do we need? A confirmation written in email? Or as a PR? (This is my genuine question, because I want to be more effective on this kind of things.)

In this case, I chatted with SF via Teams channel. I can add you into that group chat upon request.

gladjohn: I do it [this way](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3184155), so we have tracking items closed and partner's agreement captured

msal/managed_identity.py
rayluo: TL;DR: This is not really a problem here, but still thanks for pointing this out.

Long explanation. There are (inevitably?) several internal layers of helpers. Currently, most of the helpers use a name "access_token_sha256_to_refresh", which "access" prefix was influenced by MSAL C++'s official parameter name "AccessTokenToRefresh". But we did use the right name "token_sha256_to_refresh" on the wire when sending out the request. You can confirm that by both the implementation inside `_obtain_token_on_service_fabric()` and by the test cases.

tests/test_mi.py
rayluo: Not applicable, unless we would also want to implement CCA.AcquireTokenForClient(...).WithAccessTokenSha256ToRefresh(...) in non-dotnet MSALs.

tests/test_mi.py
rayluo: Not currently applicable, as MSAL Python does not support forceRefresh. We may revisit this when the circumstances changes in the future.

msal/managed_identity.py
rayluo: > I do it [this way](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3184155), so we have tracking items closed and partner's agreement captured

That work item is now in Done status, but where can we find the partner's agreement captured? The api version topic seems not mentioned, neither in that work item, nor in [the spec that work item referenced](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5137/files).

tests/test_mi.py
rayluo: Fair point. Added it into the next test case.

msal/managed_identity.py
rayluo: @gladjohn , now that we have that [final confirmation](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3184155), can we resolve this and re-review this PR?

msal/managed_identity.py
gladjohn: agreed

gladjohn (approved):

neha-bhargava (approved):

msal/managed_identity.py
```diff
@@ -266,8 +280,7 @@ def acquire_token_for_client(
             and then a *claims challenge* will be returned by the target resource,
             as a `claims_challenge` directive in the `www-authenticate` header,
             even if the app developer did not opt in for the "CP1" client capability.
-            Upon receiving a `claims_challenge`, MSAL will skip a token cache read,
-            and will attempt to acquire a new token.
+            Upon receiving a `claims_challenge`, MSAL will attempt to acquire a new token.
```
Robbie-Microsoft, line 283: I think you should specifically say "skip the cache"

msal/managed_identity.py
Robbie-Microsoft, line 602: Can you define the API value as a constant?

msal/managed_identity.py
rayluo: Actually, I specifically changed that "skip the cache *read*" into its current new form. Because "skip the cache" is an imprecise term (although we sometimes use it colloquially). Our implementation does need to look into the cache, in order to find a cached token (if any) and then calculate its hash.

msal/managed_identity.py
rayluo: This code base defines constant typically only when it's being used multiple times. If a value is used only once, we tend to use the literal value on the spot. This way, the implementation has better locality; the new reader/maintainer does not need to jump through multiple files (or rely on intellisense in their IDE) to find the real value.

It is a matter of style preference. We can talk more offline.

msal/managed_identity.py
Robbie-Microsoft: ok, resolving this

msal/managed_identity.py
Robbie-Microsoft: I personally think it would be better to define a constant for something like an API version, but you are the implementer in Python so it's up to you. I'll resolve this.

PR #820, opened 2025-05-08
Properly throw MsalServiceError exception

### LINKED ISSUE #818: [Bug] KeyError: 'error' when receiving a 400+ status code
Hello!

**Describe the bug**
Library raises this stack trace when a > 400 status is received

```
../.venv/lib/python3.12/site-packages/msal/application.py:632: in __init__
    self.authority = Authority(
../.venv/lib/python3.12/site-packages/msal/authority.py:79: in __init__
    openid_config = tenant_discovery(
../.venv/lib/python3.12/site-packages/msal/authority.py:222: in tenant_discovery
    resp.raise_for_status()
../.venv/lib/python3.12/site-packages/msal/throttled_http_client.py:69: in raise_for_status
    raise MsalServiceError("HTTP Error: {}".format(self.status_code))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = MsalServiceError('HTTP Error: 500'), args = ('HTTP Error: 500',)
kwargs = {}

    def __init__(self, *args, **kwargs):
>       super(MsalError, self).__init__(self.msg.format(**kwargs), *args)
E       KeyError: 'error'

../.venv/lib/python3.12/site-packages/msal/exceptions.py:33: KeyError
```

**To Reproduce**
The closest is: https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/dev/sample/confidential_client_sample.py

And was reproduced in a test using requests_mock

**Expected behavior**
Should successfully raise a MsalServiceError i suppose

**What you see instead**
KeyError: 'error'

Paste the output of this
`python -c "import msal; print(msal.__version__)"`
```
1.32.3
```

**Additional context**
Seems to have been introduced here:
https://github.com/AzureAD/microsoft-authentication-library-for-python/commit/29d1ac1558d6f87754fef71ed950bd806919d803#diff-3ebc80b207eb5a4e4eae904208ad8d17cc8f7eaf23f5a7d82a185578f79898bfR61

This instantiation can't work

```
>>> from msal.exceptions import MsalServiceError
>>> MsalServiceError("HTTP Error: {}".format(401))
```

rayluo:
MSAL Python defined a `MsalServiceError` exception for years but did not actually use/throw it, until recently. This PR refactors the usage of this exception so that it is used properly. Comes with test cases.

### DIFF
--- a/msal/exceptions.py
+++ b/msal/exceptions.py
@@ -27,12 +27,26 @@
 
 class MsalError(Exception):
     # Define the template in Unicode to accommodate possible Unicode variables
-    msg = u'An unspecified error'
+    msg = u'An unspecified error'  # Keeping for backward compatibility
 
-    def __init__(self, *args, **kwargs):
-        super(MsalError, self).__init__(self.msg.format(**kwargs), *args)
-        self.kwargs = kwargs
 
 class MsalServiceError(MsalError):
-    msg = u"{error}: {error_description}"
+    msg = u"{error}: {error_description}"  # Keeping for backward compatibility
+    def __init__(
+        self,
+        *args,
+        error: str, error_description: str,  # Historically required, keeping them for now
+            # 1. We can't simply remove them, or else it will be a breaking change
+            # 2. We may change them to optional without breaking anyone. However,
+            #    such a change will be a one-way change, because once being optional,
+            #    we will never be able to change them (back) to be required.
+            # 3. Since they were required and already exist anyway,
+            #    now we just keep them required "for now",
+            #    just in case that we would use them again.
+            # There is no plan to do #1; and we keep option #2 open; we go with #3.
+        **kwargs,
+    ):
+        super().__init__(*args, **kwargs)
+        self._error = error
+        self._error_description = error_description
 
--- a/msal/throttled_http_client.py
+++ b/msal/throttled_http_client.py
@@ -66,7 +66,10 @@ class NormalizedResponse(Response):
     # self.raise_for_status = raw_response.raise_for_status
     def raise_for_status(self):
         if self.status_code >= 400:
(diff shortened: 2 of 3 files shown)


msal/exceptions.py
```diff
@@ -27,12 +27,18 @@
...
-        super(MsalError, self).__init__(self.msg.format(**kwargs), *args)
-        self.kwargs = kwargs
 
 class MsalServiceError(MsalError):
-    msg = u"{error}: {error_description}"
+    msg = u"{error}: {error_description}"  # Keeping for backward compatibility
+    def __init__(
+        self,
+        *args,
+        error: str, error_description: str,  # Historically required, keeping them for now
```
Avery-Dunn, line 38: Since it's only being kept "for now", is there any plan/backlog item to remove this at some point? Or is it something that might be needed again in the future?

Avery-Dunn (approved):

gladjohn (approved):

msal/exceptions.py
rayluo: Good question. Perhaps I shall add more inline comments to elaborate:

1. We can't simply remove it, or else it will be a breaking change
2. We may, however, change required parameters into optional, at any time, without breaking anyone. However, such a change will be a one-way change, because once being optional, we will never be able to change it (back) to be required.
3. Since it was required and already exists anyway, now we just keep it required "for now", just in case that we would use it again.

There is no plan to do `#1`; and we will keep option `#2` open; currently, we are doing `#3`.

neha-bhargava (approved):

PR #822, opened 2025-05-15
Improve test cases to test header-less response

### LINKED ISSUE #817: [Feature Request] Add MSI E2E tests
### MSAL client type

Confidential

### Problem Statement

MSI E2E tests, similar to MSAL.NET and MSAL.GO

These should cover: IMDS, Arc and AppService

### Proposed solution

_No response_

rayluo:
MSAL has long been allowing callers to specify their own http client. Early versions of MSAL Python did not require that http client's response to contain headers. This PR changes the test helper to mimic a header-less response by default, so that our existing test cases will automatically cover header-less response scenarios. This can help preventing regression.

P.S.
Issue #817 was created in an aftermath of such a regression. This PR precisely covers those scenarios; therefore, it can resolve #817.
(End-to-end tests, while being useful in their own right, would not be enough to guard against this kind of "header-less response regression", because real e2e requests are typically sent with real http clients that always contain headers.)

Lastly, note that this PR changes only a couple of files in the "./tests" directory. It does not alter any existing MSAL behavior. It is a low-risk PR.

### DIFF
--- a/tests/test_throttled_http_client.py
+++ b/tests/test_throttled_http_client.py
@@ -50,18 +50,6 @@ class DummyHttpClient(object):
         raise CloseMethodCalled("Not used by MSAL, but our customers may use it")
 
 
-class DummyHttpClientWithoutResponseHeaders(DummyHttpClient):
-    def post(self, url, params=None, data=None, headers=None, **kwargs):
-        response = super().post(url, params, data, headers, **kwargs)
-        del response.headers  # Early versions of MSAL did not require http client to return headers
-        return response
-
-    def get(self, url, params=None, headers=None, **kwargs):
-        response = super().get(url, params, headers, **kwargs)
-        del response.headers  # Early versions of MSAL did not require http client to return headers
-        return response
-
-
 class CloseMethodCalled(Exception):
     pass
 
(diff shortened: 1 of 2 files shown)


tests/test_throttled_http_client.py
```diff
@@ -82,9 +70,10 @@ def test_pickled_minimal_response_should_contain_signature(self):
             status_code=200, headers={}, text="foo")))
 
     def test_throttled_http_client_base_response_should_tolerate_headerless_response(self):
-        http_client = ThrottledHttpClientBase(DummyHttpClientWithoutResponseHeaders(
-            status_code=200, response_text="foo"))
-        response = http_client.post("https://example.com")
+        http_client = DummyHttpClient(status_code=200, response_text="foo")
```
gladjohn, line 94: should we add the regression bug info here?

gladjohn (approved):

gladjohn:
looks good @rayluo. just one minor clarification comment.

tests/test_throttled_http_client.py
rayluo: OK, I'll probably add some inline comment in my next commit.

neha-bhargava (approved):

PR #825, opened 2025-05-30
PoC: A runner to load specs-based test cases from a central service and run it as part of the test automation

rayluo:
This PR contains a runner that can load specs-based test cases from a central service. The schema of those test cases will be defined elsewhere.

The test runner is then integrated as ONE test case, into the existing test automation. Its effect will be one of the following three outcomes.
* If all specs-based test cases, you will see a line like this in the pipeline log:
  [`tests/test_smile.py .                                                    [ xx%]`](https://github.com/AzureAD/microsoft-authentication-library-for-python/actions/runs/15343449263/job/43174451643#step:6:45)
  That small dot after the file name means the multiple specs-based test cases as a whole passed.
  We don't normally need to find that log because, if all other traditional existing test cases also pass, the build will simply end with a green status, as it normally would be.
* If some of the specs-based test cases failed, you will see a line like this in the pipeline log:
  `tests/test_smile.py F                                                    [ xx%]`
  That capital F after the file name means some of the multiple specs failed and, as a result, the entire test automation run will be marked as failed.
* If for whatever reason the centralized specs-based test cases service is down, this PR implements it to automatically skip the specs-based test cases. You will see a line like this in the pipeline log:
  [`tests/test_smile.py s                                                    [ xx%]`](https://github.com/AzureAD/microsoft-authentication-library-for-python/actions/runs/15342520558/job/43171802813#step:6:44)
  That lowercase s after the file name means the specs-based test cases were skipped.
  This way, the occasional central test service maintenance will not block existing test automation.

The test runner currently supports test cases for managed identity v1. Going forward, this repo will receive more changes to the test runner to support more scenarios such as CCA. Actual specs-based test cases will continue to be hosted elsewhere centrally.

### DIFF
--- a/tests/smile.py
+++ b/tests/smile.py
@@ -1,0 +1,241 @@
+#!/usr/bin/env python3
+"""
+MSAL Feature Test Runner
+Interprets testcase file(s) to create and execute test cases using MSAL.
+
+Initially created by the following prompt:
+Write a python implementation that can read content from feature.yml, create variables whose names are defined in the "arrange" mapping's keys, and the variables' value are derived from the "arrange" mapping's value; interpret those value as if they are python snippet using MSAL library.
+"""
+import os
+import sys
+import logging
+from contextlib import contextmanager
+from typing import Dict, Any, List, Optional
+
+import yaml
+import msal
+import requests
+
+
+logging.basicConfig(level=logging.INFO, format='%(asctime)s - %(name)s - %(levelname)s - %(message)s')
+logger = logging.getLogger(__name__)
+
+class SmileTestRunner:
+
+    def __init__(self, testcase_url: str):
+        self.testcase_url = testcase_url
+        self.test_spec = None
+        self.variables = {}
+
+    def load_feature(self) -> Dict[str, Any]:
+        """Load and validate the feature file."""
+        try:
+            with requests.get(self.testcase_url) as response:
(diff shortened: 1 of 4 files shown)


tests/smile.py
```diff
@@ -0,0 +1,241 @@
...
+        self.testcase_url = testcase_url
+        self.test_spec = None
+        self.variables = {}
+
+    def load_feature(self) -> Dict[str, Any]:
+        """Load and validate the feature file."""
+        try:
+            with requests.get(self.testcase_url) as response:
+                response.raise_for_status()
+                self.test_spec = yaml.safe_load(response.text)
```
bgavrilMS, line 35: We discussed that the tests would be written in plain language (English)?

tests/smile.py
```diff
@@ -0,0 +1,241 @@
+#!/usr/bin/env python3
+"""
+MSAL Feature Test Runner
+Interprets testcase file(s) to create and execute test cases using MSAL.
+
+Initially created by the following prompt:
+Write a python implementation that can read content from feature.yml, create variables whose names are defined in the "arrange" mapping's keys, and the variables' value are derived from the "arrange" mapping's value; interpret those value as if they are python snippet using MSAL library.
```
bgavrilMS, line 7: Where is feature.yml ?

PR #831, opened 2025-06-13
Use lowercase environment value during searching

rayluo:
Context: The "environment" field in the token cache contains an authority's domain name, which is stored as lowercase in cache. We did that right. But we noticed that when performing a search in the token cache, we were not converting the "environment" field to lowercase. This is usually not a problem, because the domain name is typically in lowercase. And no customer reported any issue. However, recently    when we are running Managed Identity test cases on Windows, we found out that the domain name placeholder that we use, turns out to be mixed case on Windows, such as "CPC-raylu-UB85C". This would cause cache miss when running the Managed Identity client on a Windows machine.

This PR improves the token cache search/match logic to always search environment by lowercase.

### DIFF
--- a/msal/token_cache.py
+++ b/msal/token_cache.py
@@ -126,7 +126,13 @@ class TokenCache(object):
 
     @staticmethod
     def _is_matching(entry: dict, query: dict, target_set: set = None) -> bool:
-        return is_subdict_of(query or {}, entry) and (
+        query_with_lowercase_environment = {
+            # __add() canonicalized entry's environment value to lower case,
+            # so we do the same here.
+            k: v.lower() if k == "environment" and isinstance(v, str) else v
+            for k, v in query.items()
+        } if query else {}
+        return is_subdict_of(query_with_lowercase_environment, entry) and (
             target_set <= set(entry.get("target", "").split())
             if target_set else True)
 
--- a/tests/test_mi.py
+++ b/tests/test_mi.py
@@ -190,6 +190,10 @@ class VmTestCase(ClientTestCase):
             headers={'Metadata': 'true'},
             )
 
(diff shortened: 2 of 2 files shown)


msal/token_cache.py
```diff
@@ -126,7 +126,13 @@ def _get(self, credential_type, key, default=None):  # O(1)
 
     @staticmethod
     def _is_matching(entry: dict, query: dict, target_set: set = None) -> bool:
-        return is_subdict_of(query or {}, entry) and (
+        query_with_lowercase_environment = {
+            # __add() canonicalized entry's environment value to lower case,
+            # so we do the same here.
+            k: v.lower() if k == "environment" else v
```
gladjohn: can this (v) be null?

msal/token_cache.py
```diff
@@ -126,7 +126,13 @@ def _get(self, credential_type, key, default=None):  # O(1)
...
+            for k, v in query.items()
```
gladjohn, line 133: would be good to have some unit tests to assert retrieval with mixed-case query succeeds

rayluo: Thanks for the reminding. A new test case has been added below.

msal/token_cache.py
rayluo: Improbable, because the values are not derived from arbitrary input; they are with actual data generated by our own MSAL implementation. That being said, since you brought this up, a new change has been made to add some defensive programming. Thanks!

msal/token_cache.py
rayluo: @gladjohn , can you review this again to see whether your comments were resolved?

gladjohn (approved):

PR #833, opened 2025-06-19
Document how to enable sha256 for client credential

### LINKED ISSUE #832: [Bug] The docs and the API enable ppl to use SHA256 for credentialas
Related to - https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/639 

The docs still reference the use of SHA1 https://msal-python.readthedocs.io/en/latest/#msal.ClientApplication.params.client_credential

This needs to be fixed as per original issue. All docs need to showcase the use of SHA256.

This is not just a documentation issue, those dictionaries need to take in SHA256, or the dictionaries need to be deprecated.

rayluo:
This PR updates the documentation for how to enable SHA-256 thumbprint for client credentials.

Note that the current interface does not take an explicit SHA-256 thumbprint. It will automatically use the SHA-256 thumbprint of the certificate when reading client certificates from PFX files.

The new doc can be read from this [doc staging site](https://msal-python.readthedocs.io/en/docs-staging/#msal.ClientApplication.__init__). The new sentences are scattered in 4 different purple boxes.

This PR will resolve #832 .

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -280,15 +280,17 @@ class ClientApplication(object):
 
             .. admonition:: Support using a certificate in X.509 (.pem) format
 
-                Deprecated because it uses SHA-1 thumbprint.
+                Deprecated because it uses SHA-1 thumbprint,
+                unless you are still using ADFS which supports SHA-1 thumbprint only.
                 Please use the .pfx option documented later in this page.
 
                 Feed in a dict in this form::
 
                     {
                         "private_key": "...-----BEGIN PRIVATE KEY-----... in PEM format",
-                        "thumbprint": "A1B2C3D4E5F6...",
-                        "passphrase": "Passphrase if the private_key is encrypted (Optional. Added in version 1.6.0)",
+                        "thumbprint": "An SHA-1 thumbprint such as A1B2C3D4E5F6...",
+                        "passphrase": "Needed if the private_key is encrypted (Added in version 1.6.0)",
+                        "public_certificate": "...-----BEGIN CERTIFICATE-----...",  # Needed if you use Subject Name/Issuer auth. Added in version 0.5.0.
                     }
 
                 MSAL Python requires a "private_key" in PEM format.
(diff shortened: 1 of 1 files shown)


msal/application.py
```diff
@@ -280,6 +280,9 @@ def __init__(
 
             .. admonition:: Support using a certificate in X.509 (.pem) format
 
+                Deprecated because it uses SHA-1 thumbprint.
```
Robbie-Microsoft: It's deprecated, but does it still work for backwards compatibility? Should you note that?

Robbie-Microsoft:
I'm confused as to how this all works... the changes in this PR are reflected on the document staging site? Is that correct?

There are 6 purple boxes there btw, not 4 like you've indicted above. And the changes in this PR don't seem to line up with the content of the purple boxes.

msal/application.py
rayluo: My understanding of the term Deprecation is aligned with [this answer](https://stackoverflow.com/a/9208164/728675):

> Deprecated means that it is still in use, but only for historical purposes and it will be removed probably in the next big release. It is recommended that you do not use deprecated functions or features - even if they are present in the current library for example.

So, yes, it still works for backwards compatibility. If that is OK with you, I intend to not add a "it still works for backwards compatibility" statement to the docs here, otherwise we might need to retrofit the same sentence into maybe a dozen deprecated parameters here and there in our docs.

rayluo:
> > The new doc can be read from this [doc staging site](https://msal-python.readthedocs.io/en/docs-staging/#msal.ClientApplication.__init__). The new sentences are scattered in 4 different purple boxes.
>
> I'm confused as to how this all works... the changes in this PR are reflected on the document staging site? Is that correct?
> 
> There are 6 purple boxes there btw, not 4 like you've indicted above. And the changes in this PR don't seem to line up with the content of the purple boxes.

Sorry about the confusion. There are 6 purple boxes for 6 scenarios. This PR deprecates 2 of them and added SHA-2 content into another 2, so the changes are "scattered in FOUR (~not~ among six) different purple boxes".

Yes, the changes in this PR are reflected on the document staging site. If necessary, you can compare it with the current status-quo doc.
Before (i.e. the current latest): https://msal-python.readthedocs.io/en/latest/#msal.ClientApplication.params.client_credential
After (i.e. the staging): https://msal-python.readthedocs.io/en/docs-staging/#msal.ClientApplication.params.client_credential

bgavrilMS (changes requested):
The x509 from disk is not a good solution, as per our advisors.

msal/application.py
bgavrilMS: Robbie is right though. ADFS still uses this. So it's not fully deprecated.

The pfx solution is not yet fit for purpose, since it requires the cert to be on disk. 

How about adding a `thumbprintSha256` to the dictionary and adding a note that `thumbprint` is deprecated except for ADFS instead?

Internally, MSAL will choose `thumbpritnSha256` if it is set, or fallback to `thumbprint` if it is not set.

bgavrilMS:
> > > The new doc can be read from this [doc staging site](https://msal-python.readthedocs.io/en/docs-staging/#msal.ClientApplication.__init__). The new sentences are scattered in 4 different purple boxes.
> > 
> > 
> > I'm confused as to how this all works... the changes in this PR are reflected on the document staging site? Is that correct?
> > There are 6 purple boxes there btw, not 4 like you've indicted above. And the changes in this PR don't seem to line up with the content of the purple boxes.
> 
> Sorry about the confusion. There are 6 purple boxes for 6 scenarios. This PR deprecates 2 of them and added SHA-2 content into another 2, so the changes are "scattered in FOUR (not six) different purple boxes".
> 
> Yes, the changes in this PR are reflected on the document staging site. If necessary, you can compare it with the current status-quo doc. Before (i.e. the current latest): https://msal-python.readthedocs.io/en/latest/#msal.ClientApplication.params.client_credential After (i.e. the staging): https://msal-python.readthedocs.io/en/docs-staging/#msal.ClientApplication.params.client_credential

suggestion: I agree with @Robbie-Microsoft that there are too many boxes. Maybe you should combine the "normal cert" with the "sn/i cert" boxes?

rayluo:
* Good point on the deprecation is only for non-ADFS. With ADFS, customer still needs to use SHA-1. Docs has been modified accordingly.
* Those 6 boxes were historically added in some 5 or 6 different versions, so they became that way. But you are both right that 6 sounds too much. The latest commit consolidated them into 4.
* I have updated the PR accordingly, and the doc staging site is also updated.
* Yes, we shall also support loading a cert without a file. Shall we create a dedicate github issue to track that enhancement? Meanwhile, this PR correctly represents the status quo of the current certificate usage of the MSAL Python versions already shipped. So, this PR in itself does not make situation any worse. Perhaps we can still review and merge this PR first. Shall we?

rayluo:
> * Good point on the deprecation is only for non-ADFS. With ADFS, customer still needs to use SHA-1. Docs has been modified accordingly.
> * Those 6 boxes were historically added in some 5 or 6 different versions, so they became that way. But you are both right that 6 sounds too much. The latest commit consolidated them into 4.
> * I have updated the PR accordingly, and the doc staging site is also updated.
> * Yes, we shall also support loading a cert without a file. Shall we create a dedicate github issue to track that enhancement? Meanwhile, this PR correctly represents the status quo of the current certificate usage of the MSAL Python versions already shipped. So, this PR in itself does not make situation any worse. Perhaps we can still review and merge this PR first. Shall we?

Hi @bgavrilMS , this PR improves the docs for the existing MSAL Python behaviors. May I get your approval to merge this PR as-is and get back to other improvement ideas in new PRs?

bgavrilMS (approved):

PR #835, opened 2025-06-24
Thumbprint for certificate made optional

vi7us:
As per request not to use SHA-1 algorithm for calculation of  SHA-1 certificate thumbprint, this parameter is now optional and if not passed, is calculated inside the library itself.

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -66,16 +66,21 @@ def _str2bytes(raw):
     except:
         return raw
 
-def _extract_cert_and_thumbprints(private_key, cert):
+def _extract_cert_and_thumbprints(cert):
     # Cert concepts https://security.stackexchange.com/a/226758/125264
     from cryptography.hazmat.primitives import hashes, serialization
-    cert_pem = cert.public_bytes(encoding=serialization.Encoding.PEM).decode()
+    cert_pem = cert.public_bytes(  # Requires cryptography 1.0+
+        encoding=serialization.Encoding.PEM).decode()
     x5c = [
-        '\n'.join(cert_pem.splitlines()[1:-1])
+        '\n'.join(
+            cert_pem.splitlines()
+            [1:-1]  # Strip the "--- header ---" and "--- footer ---"
+        )
     ]
-    sha256_thumbprint = cert.fingerprint(hashes.SHA256()).hex()
-    sha1_thumbprint = cert.fingerprint(hashes.SHA1()).hex()
-    return private_key, sha256_thumbprint, sha1_thumbprint, x5c
+    # https://cryptography.io/en/latest/x509/reference/#x-509-certificate-object
+    sha256_thumbprint = cert.fingerprint(hashes.SHA256()).hex()  # Requires cryptography 0.7+
+    sha1_thumbprint = cert.fingerprint(hashes.SHA1()).hex()  # Requires cryptography 0.7+
+    return sha256_thumbprint, sha1_thumbprint, x5c
 
 def _parse_pfx(pfx_path, passphrase_bytes):
     # Cert concepts https://security.stackexchange.com/a/226758/125264
@@ -86,7 +91,8 @@ def _parse_pfx(pfx_path, passphrase_bytes):
             f.read(), passphrase_bytes)
     if not (private_key and cert):
         raise ValueError("Your PFX file shall contain both private key and cert")
-    return _extract_cert_and_thumbprints(private_key, cert)
+    sha256_thumbprint, sha1_thumbprint, x5c = _extract_cert_and_thumbprints(cert)
+    return private_key, sha256_thumbprint, sha1_thumbprint, x5c
 
 
 def _load_private_key_from_pem_str(private_key_pem_str, passphrase_bytes):
@@ -290,7 +296,10 @@ class ClientApplication(object):
 
                     {
                         "private_key": "...-----BEGIN PRIVATE KEY-----... in PEM format",
-                        "thumbprint": "An SHA-1 thumbprint such as A1B2C3D4E5F6...",  # Optional since version 1.35.0
+                        "thumbprint": "An SHA-1 thumbprint such as A1B2C3D4E5F6..."
+                            "Changed in version 1.35.0, if thumbprint is absent"
+                            "and a public_certificate is present, MSAL will"
+                            "automatically calculate an SHA-256 thumbprint instead.",
                         "passphrase": "Needed if the private_key is encrypted (Added in version 1.6.0)",
                         "public_certificate": "...-----BEGIN CERTIFICATE-----...",  # Needed if you use Subject Name/Issuer auth. Added in version 0.5.0.
                     }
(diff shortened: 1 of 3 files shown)


msal/application.py
```diff
@@ -306,7 +308,7 @@ def __init__(
 
                     {
                         "private_key": "...-----BEGIN PRIVATE KEY-----... in PEM format",
-                        "thumbprint": "A1B2C3D4E5F6...",
+                        "thumbprint": "A1B2C3D4E5F6...", (Optinal, if not provided, MSAL will calculate it. Added in version 1.34.0)
```
bgavrilMS:
```suggestion
                        "thumbprint": "A1B2C3D4E5F6...", (Deprecated. Provide the public certificate instead.)
```

bgavrilMS: Can you update the doc at line 288 as well?

bgavrilMS (approved):
Approved with comments. @rayluo for final sign-off

msal/application.py
vi7us: @bgavrilMS i'm not sure if it will work without public cert passed as we use it to calculate x5c, maybe method _extract_cert_and_thumbprints can be updated to check if public cert is passed and if not x5c will be set to None?

rayluo: I thought you [said (in PR 833)](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/833#discussion_r2162622909) that it is not fully deprecated. We will need to settle on the wording, one way or another.

msal/application.py
```diff
@@ -815,6 +817,15 @@ def _build_client(self, client_credential, authority, skip_regional_client=False
                         passphrase_bytes)
                     if client_credential.get("public_certificate") is True and x5c:
                         headers["x5c"] = x5c
+                elif (client_credential.get("private_key") 
+                      and client_credential.get("public_certificate") 
+                      and not client_credential.get("thumbprint")): # in case user does not pass thumbprint but only certificate and private key
```
rayluo: @vi7us , so this implementation only works for the Subject Name/Issuer (SNI) scenario. Why not also make it work for non-SNI?

Besides, this implementation does not seem to handle the passphrase in the `{"private_key": "...", "public_certificate": "...", "passphrase": "..."}` situation. That feels like a regression. Would you consider refactoring this into the next `elif` block starting at line 830?

Also, in case you are not aware of, [currently MSAL Python's `.pfx` code path already supports the absence of thumbprint](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/833#issuecomment-2998955868). We will probably consider adding a new "private_key_pfx_bytes" in the near future.

rayluo (changes requested):

msal/application.py
vi7us: Updated as per request

msal/application.py
vi7us: @rayluo 
- updated to handle also passphrase option. 
- For the Non-SNI do you mean to include new parameter that will be used as a flag to add x5c to the headers? 
```
       "private_key": "...-----BEGIN PRIVATE KEY-----... in PEM format",
       "thumbprint": "A1B2C3D4E5F6...",
       "public_certificate": "...-----BEGIN CERTIFICATE-----...",
       "passphrase": "Passphrase if the private_key is encrypted (Optional. Added in version 1.6.0)",

       "sni_auth": [True|False], # new param
```

I know it is possible to use .pfx file but that means it must be saved locally which is not what we were suggested (as per email thread). 

For the `private_key_pfx_bytes` do you mean doing something like this?
```
        pkcs12Decoded = base64.b64decode(private_key_pfx_bytes)
        private_key, certificate, _ = pkcs12.load_key_and_certificates(pkcs12Decoded, B"")
```

rayluo: @vi7us , thanks for continuing with this PR!

* For non-SNI, we are already working on [another PR to have non-SNI and SNI share the same doc "box"](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/833/files) as we speak, the [staged doc is already rendered here](https://msal-python.readthedocs.io/en/docs-staging/#msal.ClientApplication.params.client_credential). Building on top of that concept and docs, what I have in mind in terms of the usage is this (CC @bgavrilMS ):

  ```python
  {
    "private_key": "...-----BEGIN PRIVATE KEY-----... in PEM format",

    # If present, thumbprint shall be SHA-1 which is still needed when using ADFS.
    # If absent, MSAL Python will automatically use SHA256 thumbprint.
    "thumbprint": "An SHA-1 thumbprint such as A1B2C3D4E5F6...",  # Optional since version 1.34.0

    "passphrase": "Needed if the private_key is encrypted (Added in version 1.6.0)",

    "public_certificate": "...-----BEGIN CERTIFICATE-----...",  # Needed if you use Subject Name/Issuer auth. Added in version 0.5.0.
  }
  ```
  Implementation wise, I wonder whether the new `elif ...` block can be refactored into its next `else` block, to reduce code duplication on the passphrase handling.

* Forget about my `private_key_pfx_bytes` idea, for now. Representing binary data in `client_credential` parameter has other implications (so far the `client_credential` value could be `json.load(...)` from a text string, via env var; introducing binary data would hinder that usage). Luckily, we don't immediately need it if we get this thumbprint-less PEM code path working.

msal/sku.py
```diff
@@ -2,5 +2,5 @@
 """
 
 # The __init__.py will import this. Not the other way around.
-__version__ = "1.33.0b1"
+__version__ = "1.34.0b1"
```
rayluo: Please also revert this change.

We will only bump version number when we cut a release, and decide the actual number at that time.

Besides, bumping version number in each feature PR brings complications in case we would end up reverting the feature PR.

msal/application.py
bgavrilMS: @rayluo - if the CX provides the public certificate, then thumbprint is not needed? MSAL can always compute it? In that case the property becomes indeed deprecated. Internally, MSAL will still continue to use SHA1 for ADFS.

rayluo: If I am not mistaken, the thumbprint is a hash of the private key, not the public certificate. So, whether the customer provides a public certificate (to use SNI feature), is irrelevant.

But you are right that MSAL could treat the absence of the thumbprint parameter as "compute the hash automatically". We may still want to keep the old `thumbprint="an sha-1 hash"` usage for ADFS, because MSAL historically just uses a **magic string** "https://contoso.com **/adfs**" to detect ADFS mode, and that might not always be reliable. Better keep the old `thumbprint="an sha-1 hash"` usage, so that ADFS customer may still opt in to use SHA-1 when needed.

msal/application.py
rayluo, line 849: @bgavrilMS , @vi7us ,

I was under the impression that this PR makes the usage of SHA-1 `thumbprint` parameter optional, by automatically calculating an SHA-1 thumbprint. But it ends up implicitly change the thumbprint-less code path to calculate an SHA256 instead.

* Before this PR, the Subject Name / Issuer (SNI) code path sets only `sha1_thrumbprint` therefore @vi7us was NOT hitting this `sha256_thumbprint` code path here.

* After @vi7us 's first commit with the new `_extract_cert_and_thumbprints()` helper calculating both sha1 and sha256, I believe this sha256_thumbprint code path becomes activated. This is a subtle behavior change, but not necessarily a bad thing. Previously, I wasn't sure whether/when Entra ID fully supports sha256, now @vi7us made the change and started using the sha256 thumbprint code path successfully, we can infer that Entra ID indeed fully supports `{"algorithm": "PS256", "sha256_thumbprint": sha256_thumbprint}` now.

Therefore, I refactored this PR so that it becomes clear that (1) when `thumbprint` parameter is used, it expects an SHA1 thumbprint (for backward compatibility and for ADFS scenario); (2) when the `thumbprint` parameter is absent, MSAL Python will use an SHA**256** (rather than SHA**1**) thumbprint.

@vi7us , if at all possible, please test this now-rebased feature branch (you can back up your original feature branch if desirable). If there is no issue reported from your side, I'll merge in this refactored PR and ship it in our next release.

rayluo (approved):

msal/application.py
Avery-Dunn: > Previously, I wasn't sure whether/when Entra ID fully supports sha256... (1) when thumbprint parameter is used, it expects an SHA1 thumbprint (for backward compatibility and for ADFS scenario); (2) when the thumbprint parameter is absent, MSAL Python will use an SHA256 (rather than SHA1) thumbprint.

In case there were still any doubts, this is how Java and .NET currently work as well (SHA1 for ADFS, SHA256 for everything else):
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4616
https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/840

msal/application.py
```diff
@@ -803,15 +814,30 @@ def _build_client(self, client_credential, authority, skip_regional_client=False
...
+                        _load_private_key_from_pem_str(
                             client_credential['private_key'], passphrase_bytes)
-                    else:  # PEM without passphrase
-                        private_key = client_credential['private_key']
+                        if passphrase_bytes
+                        else client_credential['private_key']
+                    )
+
+                    # Determine thumbprints based on what's provided
+                    if client_credential.get("thumbprint"):
```
Avery-Dunn, line 826: Are there any tests covering this behavior?

rayluo: Good point! A new test file has been added in the modified commit. Please re-review.

Avery-Dunn (approved):

PR #837, opened 2025-07-10
Expose broker_error_status to public

fengga:
https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3313444

Azure CLI asking for exposing the broker error status, other than finding a substring inside the error description to determine if it is InteractionRequired


jiasli:
This PR is no longer necessary. See the explanation at https://github.com/Azure/azure-cli/pull/31699#issuecomment-3068103598.

PR #844, opened 2025-08-06
Change the value of "x-client-os" header

tongyuze:
eSTS expects we send x-client-os: Windows instead of win32 as there are no OS called win 32. win32 is not a recognized OS string in the current server-side platform detection logic.

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -2,6 +2,7 @@ import functools
 import json
 import time
 import logging
+import platform
 import sys
 import warnings
 from threading import Lock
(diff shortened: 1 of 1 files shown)


rayluo (changes requested):
Stakeholders discussed offline and are evaluating alternative options. So, I am leaving a Request Changes review here for now, so that my other teammates will not merge this PR as-is.

rayluo:
Summary of the offline discussion.

* We consider have service-side consume the client-side "win32" value which has existed since day one.
* Regardless, the user-agent value and x-client-os value, which is what CA relies on for platform detection in this case, is easily spoofable even without a proxy, which is why our [public guidance](https://learn.microsoft.com/en-us/entra/identity/conditional-access/concept-conditional-access-conditions#device-platforms) advises against using platform information for Conditional Access controls.

PR #845, opened 2025-08-12
ADFS labs were decommissioned since late July 2025

rayluo:
Disables the relevant e2e test cases

### DIFF
--- a/tests/test_e2e.py
+++ b/tests/test_e2e.py
@@ -841,12 +841,14 @@ class AtPopWithExternalKeyTestCase(PopWithExternalKeyTestCase):
 
 
 class WorldWideTestCase(LabBasedTestCase):
+    _ADFS_LABS_UNAVAILABLE = "ADFS labs were temporarily down since July 2025 until further notice"
 
     def test_aad_managed_user(self):  # Pure cloud
         config = self.get_lab_user(usertype="cloud")
         config["password"] = self.get_lab_user_secret(config["lab_name"])
         self._test_username_password(**config)
 
+    @unittest.skip(_ADFS_LABS_UNAVAILABLE)
     def test_adfs4_fed_user(self):
         config = self.get_lab_user(usertype="federated", federationProvider="ADFSv4")
         config["password"] = self.get_lab_user_secret(config["lab_name"])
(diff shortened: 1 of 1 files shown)


tests/test_e2e.py
```diff
@@ -841,12 +841,14 @@ def test_user_account(self):
...
 
 class WorldWideTestCase(LabBasedTestCase):
+    _ADFS_LABS_DECOMMISSIONED = "ADFS labs were decommissioned since July 2025 until further notice"
 
     def test_aad_managed_user(self):  # Pure cloud
         config = self.get_lab_user(usertype="cloud")
         config["password"] = self.get_lab_user_secret(config["lab_name"])
         self._test_username_password(**config)
 
+    @unittest.skip(_ADFS_LABS_DECOMMISSIONED)
```
bgavrilMS: ADFSv2 was not decommissioned

tests/test_e2e.py
```diff
@@ -864,6 +866,7 @@ def test_adfs2_fed_user(self):
...
     def test_adfs2019_fed_user(self):
```
bgavrilMS, line 870: not decommisioned

bgavrilMS (changes requested):
advs 2019 and v4 were not decommissioned.

tests/test_e2e.py
rayluo: When running this test case, it will visit the MEX endpoint of the ADFS https://fs.msidlab8.com/adfs/services/trust/mex

That endpoint gives an http 504 error right now. If it is not decommissioned, then we will need to investigate whether it is not configured properly.

tests/test_e2e.py
rayluo: When running this test case, it will visit the MEX endpoint of the ADFS https://fs.msidlab4.com/adfs/services/trust/mex

That endpoint gives an http 504 error right now. If it is not decommissioned, then we will need to investigate whether it is not configured properly.

tests/test_e2e.py
```diff
@@ -864,6 +866,7 @@ def test_adfs2_fed_user(self):
         config["password"] = self.get_lab_user_secret(config["lab_name"])
         self._test_username_password(**config)
 
+    @unittest.skip(_ADFS_LABS_DECOMMISSIONED)
```
gladjohn:
```suggestion
    @unittest.skip(_ADFS_TESTS)
```

rayluo: @bgavrilMS and @gladjohn ,
Just to be clear. All the modified test cases are currently failing, so I have to disable them to make the build pipeline passing (or else I wouldn't be able to ship MSAL Python 1.34.0b1 last week).

@gladjohn , if your concern is only about the wording, I/we could change the [message in line 844](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/845/files#diff-ed3b8af48cd21dc1ac735b839d89b172a6f45569c324af70a2bdb4f60e5fbcddR844).

bgavrilMS (approved):

PR #846, opened 2025-08-27
Set Redirect Uri for broker silent flow on Linux platform

xinyuxu1026:
Customer got 'redirectUri is NULL or EMPTY|illegal_argument_exception' on Linux platform due to the authParameter check in Linux Java broker. Set the redirect uri to default in silent flow on Linux platform to resolve the issue.

### DIFF
--- a/msal/broker.py
+++ b/msal/broker.py
@@ -145,17 +145,20 @@ def _build_msal_runtime_auth_params(client_id, authority):
     params.set_additional_parameter("msal_client_ver", __version__)
     return params
 
+def _set_redirect_uri_for_linux(params):
+    if sys.platform == "linux":
+        # This is required by Linux Java Broker to set a non-empty valid redirect_uri
+        params.set_redirect_uri(
+            "https://login.microsoftonline.com/common/oauth2/nativeclient"
+        )
+
 def _signin_silently(
         authority, client_id, scopes, correlation_id=None, claims=None,
         enable_msa_pt=False,
         auth_scheme=None,
         **kwargs):
     params = _build_msal_runtime_auth_params(client_id, authority)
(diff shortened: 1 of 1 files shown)


msal/broker.py
```diff
@@ -151,6 +151,11 @@ def _signin_silently(
         auth_scheme=None,
         **kwargs):
     params = _build_msal_runtime_auth_params(client_id, authority)
+    if sys.platform == "linux":
+        # This is required by Linux Java Broker to set a non-empty valid redirect_uri
+        params.set_redirect_uri(
+            "https://login.microsoftonline.com/common/oauth2/nativeclient"
+        )
```
rayluo: Thanks. The change in this PR looks straightforward. In order to reduce code duplication, though, I would suggest we move the logic to a global helper function, like this:

```python
def _set_redirect_uri_for_linux(params):
    if sys.platform == "linux":
        # This is required by Linux Java Broker to set a non-empty valid redirect_uri
        params.set_redirect_uri(
            "https://login.microsoftonline.com/common/oauth2/nativeclient"
        )
```

And then the two code paths can just do:
```suggestion
    _set_redirect_uri_for_linux(params)
```

rayluo (changes requested):
Overall looks good. A minor change is suggested.

Afterwards, you can have customer install the patch right from your feature branch by `pip install https://github.com/AzureAD/microsoft-authentication-library-for-python/archive/refs/heads/xinyu/redirecturi.zip`

msal/broker.py
xinyuxu1026: Thanks! I just updated.

rayluo (approved):

DharshanBJ (approved):

PR #855, opened 2025-09-25
ROPC deprecation

Ugonnaak1:
[AB#3385891](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3385891)

### DIFF
--- a/msal/__main__.py
+++ b/msal/__main__.py
@@ -10,7 +10,7 @@ Usage 1: Run it on the fly.
 Usage 2: Build an all-in-one executable file for bug bash.
     shiv -e msal.__main__._main -o msaltest-on-os-name.pyz .
 """
-import base64, getpass, json, logging, sys, os, atexit, msal, warnings
+import base64, getpass, json, logging, sys, os, atexit, msal
 
 _token_cache_filename = "msal_cache.bin"
 global_cache = msal.SerializableTokenCache()
--- a/msal/__main__.py
+++ b/msal/__main__.py
@@ -10,7 +10,7 @@ Usage 1: Run it on the fly.
 Usage 2: Build an all-in-one executable file for bug bash.
     shiv -e msal.__main__._main -o msaltest-on-os-name.pyz .
 """
-import base64, getpass, json, logging, sys, os, atexit, msal
+import base64, getpass, json, logging, sys, os, atexit, msal, warnings
 
 _token_cache_filename = "msal_cache.bin"
 global_cache = msal.SerializableTokenCache()
@@ -149,7 +149,13 @@ def _acquire_token_interactive(app, scopes=None, data=None):
     return result
 
 def _acquire_token_by_username_password(app):
-    """acquire_token_by_username_password() - See constraints here: https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-authentication-flows#constraints-for-ropc"""
+    """
+    [Deprecated] This API is deprecated and will be removed in a future release. Use a more secure flow instead. 
+    Migration guide: https://aka.ms/msal-ropc-migration
+
+    acquire_token_by_username_password() - See constraints here: https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-authentication-flows#constraints-for-ropc
+    """
+    warnings.warn("This API has been deprecated, please use a more secure flow. See https://aka.ms/msal-ropc-migration for migration guidance", DeprecationWarning)
     print_json(app.acquire_token_by_username_password(
         _input("username: "), getpass.getpass("password: "), scopes=_input_scopes()))
 
@@ -322,7 +328,6 @@ def _main():
             _acquire_pop_token_interactive,
             ] if isinstance(app, msal.PublicClientApplication) else []
         ) + [
-            _acquire_token_by_username_password,
             _remove_account,
         ] + ([
             _acquire_token_for_client,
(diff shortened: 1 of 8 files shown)


msal/application.py
```diff
@@ -1841,6 +1845,7 @@ def acquire_token_by_username_password(
             - A successful response would contain "access_token" key,
             - an error response would contain "error" and usually "error_description".
         """
+        warnings.warn("This API has been deprecated, please use a more secure flow. See https://aka.ms/msal-ropc-migration for migration guidance", DeprecationWarning)
```
rayluo: 1. First of all, the ROPC flow was available for both public client application (PCA) and confidential client application (CCA). The work item in PR description does not provide any detail on its scope, either. Assuming this PR means for PCA only (because the [ROPC deprecation in CCA is still TBD](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/5366)), we must add this warning conditionally.
2. [Nitpick] The [line length is recommended to be less than 79 characters](https://peps.python.org/pep-0008/#maximum-line-length). You may use the multi-line string technique here.

```suggestion
        warnings.warn("""This API has been deprecated, please use a more secure flow.
See https://aka.ms/msal-ropc-migration for migration guidance""", DeprecationWarning)
```
3. In fact, one `warnings.warn(...)` in this code path is the only needed treatment in this PR. We shall revert all other changes in this PR because, until we actually remove ropc in the next major version bump, the feature needs to exist, and all the tests are therefore still useful.

msal/__main__.py
```diff
@@ -149,7 +149,13 @@ def _acquire_token_interactive(app, scopes=None, data=None):
...
 
 def _acquire_token_by_username_password(app):
-    """acquire_token_by_username_password() - See constraints here: https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-authentication-flows#constraints-for-ropc"""
+    """
+    [Deprecated] This API is deprecated and will be removed in a future release. Use a more secure flow instead. 
+    Migration guide: https://aka.ms/msal-ropc-migration
+
+    acquire_token_by_username_password() - See constraints here: https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-authentication-flows#constraints-for-ropc
+    """
+    warnings.warn("This API has been deprecated, please use a more secure flow. See https://aka.ms/msal-ropc-migration for migration guidance", DeprecationWarning)
```
rayluo: Is this an untested PR? You shall test your PR manually to see ALL of your change(s) taking effect.

Given that this PR currently removes line 325, all these inline docs and the warning become dead code, and nobody would see those doc and warning in runtime. So, why do we still make this change here?

rayluo (changes requested):

msal/__main__.py
Ugonnaak1: I added back line 325 and have ran tests locally to make sure the change is taking effect

msal/application.py
Ugonnaak1: Understood! I've made these changes

msal/application.py
```diff
@@ -1841,6 +1845,11 @@ def acquire_token_by_username_password(
             - A successful response would contain "access_token" key,
             - an error response would contain "error" and usually "error_description".
         """
+        is_confidential_app = self.client_credential or isinstance(
+            self, ConfidentialClientApplication)
+        if not is_confidential_app:
+            warnings.warn("This API has been deprecated for PCA flows, please use a more secure flow. " \
```
bgavrilMS:
```suggestion
            warnings.warn("This API has been deprecated for public client flows, please use a more secure flow. " \
```

bgavrilMS (approved):

msal/application.py
```diff
@@ -1815,7 +1815,11 @@ def acquire_token_by_username_password(
             # because this ROPC won't work with MSA account anyway.
             auth_scheme=None,
             **kwargs):
-        """Gets a token for a given resource via user credentials.
+        """
+        [Deprecated] This API is deprecated for PublicClientApplication(PCA) flows and will be removed in a future release. Use a more secure flow instead.
```
rayluo: Suggest reverting this docstring. The warning below is sufficient.

msal/__main__.py
```diff
@@ -149,7 +149,12 @@ def _acquire_token_interactive(app, scopes=None, data=None):
     return result
 
 def _acquire_token_by_username_password(app):
-    """acquire_token_by_username_password() - See constraints here: https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-authentication-flows#constraints-for-ropc"""
+    """
+    [Deprecated] This API is deprecated for PublicClientApplication(PCA) flows and will be removed in a future release. Use a more secure flow instead.
```
rayluo: Can we not see the warning already? This docstring change seems unnecessary therefore shall be reverted.

msal/application.py
```diff
@@ -1841,6 +1845,11 @@ def acquire_token_by_username_password(
             - A successful response would contain "access_token" key,
             - an error response would contain "error" and usually "error_description".
         """
+        is_confidential_app = self.client_credential or isinstance(
+            self, ConfidentialClientApplication)
+        if not is_confidential_app:
+            warnings.warn("This API has been deprecated for public client flows, please use a more secure flow. " \
```
rayluo: It is nice and clean to use the `is_confidential_app`. Thanks!

Regarding the warning message, weren't you about to adopt [this earlier suggestion](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/855#discussion_r2380865357) verbatim? Your current implementation uses `\` which is also discouraged by the same pep8 link that I shared above.

rayluo (approved):
Thanks for your effort addressing all the comments!

PR #860, opened 2025-11-05
Deprecate the params field of _obtain_token

Avery-Dunn:
Deprecates the ability to send arbitrary query parameters to a token request according to KR [3310905](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3310905), similar to what was done in other MSALs: 
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5536

https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/1001

This behavior was intended for niche scenarios which the library did not explicitly cover, and not really meant for production. It can affect the contents and validity of tokens but were not used in our caching scheme which led to bad caching behavior: if the extra query parameters changed between requests we could return tokens that were not valid for the latest set of parameters.

MSAL Python did not have an explicit `WithExtraQueryParameters` API like other MSALs, however the internal `BaseClient._obtain_token` and `Client._obtain_token` functions had a `params` argument which set the query parameters just like other MSALs. A customer could define these params by setting them in the `kwargs` argument of some public APIs.

This PR removes the explicit `params` argument from those functions. For backwards compatibility we still check the `kwargs` argument for `params`, and if it is used a deprecation warning in `BaseClient._obtain_token` will inform the customer.

### DIFF
--- a/msal/oauth2cli/oauth2.py
+++ b/msal/oauth2cli/oauth2.py
@@ -184,14 +184,23 @@ class BaseClient(object):
 
     def _obtain_token(  # The verb "obtain" is influenced by OAUTH2 RFC 6749
             self, grant_type,
-            params=None,  # a dict to be sent as query string to the endpoint
             data=None,  # All relevant data, which will go into the http body
             headers=None,  # a dict to be sent as request headers
             post=None,  # A callable to replace requests.post(), for testing.
-                        # Such as: lambda url, **kwargs:
-                        #   Mock(status_code=200, text='{}')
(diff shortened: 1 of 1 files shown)


rayluo:
[In Python, `_name_with_leading_underscore` is considered internal helpers](https://peps.python.org/pep-0008/#descriptive-naming-styles). So, there is no need to alter this internal helper. (An analogy can be that neither of those counterpart PRs removes the query parameter from their underlying http helper functions, because they don't need to.) I would suggest we close this PR without any change.

PR #840, opened 2025-07-18
Add OIDC issuer validation

Avery-Dunn:
Adds new validation according to: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3268768

The issuer returned by the call to the OIDC endpoint (.well-known/openid-configuration) is valid if one of the following is true:
- It exactly matches the authority URL
- It has a known Microsoft host (e.g., login.microsoftonline.com)
- It has the same scheme and host as the authority (path can be different)
- For CIAM, the issuer follows the pattern of {tenant}.ciamlogin.com, even when using a custom domain

The validation behavior was added to `authority.py`, and new tests were added to `test_authority.py` to cover it

In addition, a few other tests in `test_authority.py` and  `test_application.py` were given valid issuers in their mocked OIDC discovery responses in order to pass validation, but were otherwise unchanged.

(this is the full implementation of the work started in this draft PR: https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/830)

PR #865, opened 2025-12-18
Add OIDC Issuer Validation

4gust:
### Problem

When using OIDC authority URLs, MSAL Python was not validating the `issuer` claim from the OIDC discovery response. Per the [OpenID Connect Discovery specification](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata), the issuer returned in the discovery document should be validated. Without this validation, applications could be vulnerable to accepting tokens from unauthorized identity providers.

### Solution

Added issuer validation when initializing an authority with `oidc_authority_url`:

1. **New `TRUSTED_ISSUER_HOSTS` constant** - A frozenset of well-known Microsoft identity provider hosts:
   - `login.microsoftonline.com`, `login.microsoft.com`, `login.windows.net`, `sts.windows.net`
   - National clouds: `login.chinacloudapi.cn`, `login.partner.microsoftonline.cn`, `login.microsoftonline.de`, `login.microsoftonline.us`, `login.usgovcloudapi.net`, `login-us.microsoftonline.com`

2. **New `_validate_issuer()` function** - Validates the issuer claim with O(1) lookups:
   - **Exact match**: Issuer matches the authority URL (normalized for trailing slashes)
   - **Trusted host**: Issuer is from a well-known Microsoft host
   - **Regional pattern**: Supports regional variants like `westus2.login.microsoft.com`

3. **Validation is called only for OIDC authorities** - Standard Entra authorities continue to work without additional validation

### Files Changed

- `msal/authority.py` - Added `TRUSTED_ISSUER_HOSTS`, `_validate_issuer()` function, and validation call

### Breaking Changes

Applications using `oidc_authority` parameter will now require the OIDC discovery response to include a valid `issuer` claim that either:
- Matches the authority URL exactly, OR
- Is from a trusted Microsoft identity provider host

### DIFF
--- a/msal/authority.py
+++ b/msal/authority.py
@@ -30,6 +30,20 @@ WELL_KNOWN_B2C_HOSTS = [
     ]
 _CIAM_DOMAIN_SUFFIX = ".ciamlogin.com"
 
+# Trusted issuer hosts for OIDC issuer validation.
+# These are Microsoft's well-known identity provider hosts.
+TRUSTED_ISSUER_HOSTS = frozenset([
+    AZURE_PUBLIC,                       # Microsoft Azure Worldwide
+    "login.microsoft.com",              # Microsoft Azure Worldwide
+    "login.windows.net",                # Microsoft Azure Worldwide (validation scenarios)
+    "sts.windows.net",                  # Microsoft STS
+    "login.partner.microsoftonline.cn", # Microsoft Azure China
+    AZURE_CHINA,                        # Microsoft Azure China (legacy)
+    "login.microsoftonline.de",         # Microsoft Azure Germany
+    "login-us.microsoftonline.com",     # Microsoft Azure US Government (legacy)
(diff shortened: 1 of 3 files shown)


bgavrilMS (approved):

Avery-Dunn:
What is the difference between this issuer validation and the one in this older PR? https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/840

Also, back when we did this validation for the other MSALs we had to consider the special case CIAM custom domains: a customer could set up `xyz.something.com` as their authority, but behind the scenes it seemed like Entra would still have `{tenant}.ciamlogin.com` as the issuer.

4gust:
@Avery-Dunn I hadn’t noticed that before—thank you for pointing it out.
The difference is only that caused the regional check to fail in this case.
I’ll close this issue and update #840 to also validate region-based checks.

PR #868, opened 2026-01-09
Add form_post response mode support for system browser authentication

ashok672:
Implements support for OAuth 2.0 response_mode=form_post in the system browser authentication flow, aligning with modern OAuth 2.0 best practices (RFC 6749).


msal/oauth2cli/authcode.py
```diff
@@ -111,27 +111,76 @@ class _AuthCodeHandler(BaseHTTPRequestHandler):
     def do_GET(self):
         # For flexibility, we choose to not check self.path matching redirect_uri
         #assert self.path.startswith('/THE_PATH_REGISTERED_BY_THE_APP')
-        qs = parse_qs(urlparse(self.path).query)
-        if qs.get('code') or qs.get("error"):  # So, it is an auth response
+        parsed_url = urlparse(self.path)
+        qs = parse_qs(parsed_url.query)
+        
+        if qs.get('code'):  # Auth code via GET is a security risk - reject it
```
isra-fel: How about having a grace period that both GET and POST are supported? Otherwise if an app can't just switch to form post it'll be stuck with an older version of MSAL

ashok672: I am not sure if anyone really uses this. I also don't see a scenario where the developer would be specifically preferring one method over another. Overall, I am asking for an approval for this approach for all public MSALs with Bogdan's team. Let me conclude that conversation and update here.

rayluo: Actually, we shall complete remove the `code` branch and `error` branch from `do_GET()`, based on the reasons described in the context above.

msal/oauth2cli/authcode.py
```diff
@@ -111,27 +111,76 @@ class _AuthCodeHandler(BaseHTTPRequestHandler):
...
-                template = (self.server.success_template
-                    if "code" in qs else self.server.error_template)
-                if _is_html(template.template):
-                    safe_data = _escape(auth_response)  # Foiling an XSS attack
-                else:
-                    safe_data = auth_response
-                self._send_full_response(template.safe_substitute(**safe_data))
-                self.server.auth_response = auth_response  # Set it now, after the response is likely sent
+            self._process_auth_response(auth_response)
+        elif not qs and parsed_url.path == '/':
```
rayluo: Do we really need this branch? Wouldn't the following `else:` in line 140 catch it properly?

msal/oauth2cli/authcode.py
```diff
@@ -111,27 +111,76 @@ class _AuthCodeHandler(BaseHTTPRequestHandler):
...
+                "error": "authentication_failed", 
+                "error_description": "Please close this window and try again."
+            }
+            # Include the original state so state validation doesn't fail
+            if self.server.auth_state:
+                auth_response["state"] = self.server.auth_state
+            self._process_auth_response(auth_response)
         else:
             self._send_full_response(self.server.welcome_page)
         # NOTE: Don't do self.server.shutdown() here. It'll halt the server.
```
rayluo, line 131: This line was valuable. In your new implementation, I think we better also move this line to the end of the new `do_POST()`.

msal/oauth2cli/authcode.py
```diff
@@ -111,27 +111,76 @@ class _AuthCodeHandler(BaseHTTPRequestHandler):
...
 
+    def do_POST(self):
+        # Handle form_post response mode where auth code is sent via POST body
+        content_length = int(self.headers.get('Content-Length', 0))
+        post_data = self.rfile.read(content_length).decode('utf-8')
+        try:
+            from urllib.parse import parse_qs as parse_qs_post
+        except ImportError:
+            from urlparse import parse_qs as parse_qs_post
```
rayluo: Why is this necessary? Didn't we already import it at the very beginning of the file?

msal/oauth2cli/authcode.py
```diff
@@ -111,27 +111,76 @@ class _AuthCodeHandler(BaseHTTPRequestHandler):
...
+        else:
+            self._send_full_response("Invalid POST request", is_ok=False)
+
+    def _process_auth_response(self, auth_response):
+        """Process the auth response from either GET or POST request."""
+        if self.server.auth_state and self.server.auth_state != auth_response.get("state"):
+            # OAuth2 successful and error responses contain state when it was used
+            # https://www.rfc-editor.org/rfc/rfc6749#section-4.2.2.1
+            self._send_full_response("State mismatch")  # Possibly an attack
+            # Don't set auth_response for security, but mark as done to avoid hanging
```
rayluo: The hanging effect was kind of a feature. The rationale was that normally the state shall match and there would be no hanging at all.

When the state mismatches, it usually happens when I/we were developing this, so keeping the http server running and waiting for the next response (thus "hanging") was helpful during development.

I think keeping that typically intangible "hanging" is better than introducing a new state `done` in the code path.

msal/oauth2cli/authcode.py
```diff
@@ -111,27 +111,76 @@ class _AuthCodeHandler(BaseHTTPRequestHandler):
...
+            template = (self.server.success_template
+                if "code" in auth_response else self.server.error_template)
+            if _is_html(template.template):
+                safe_data = _escape(auth_response)  # Foiling an XSS attack
+            else:
+                safe_data = dict(auth_response)  # Make a copy to avoid mutating original
+            # Provide default values for common OAuth2 response fields
+            # to avoid showing literal placeholder text like "$error_description"
+            safe_data.setdefault("error", "")
+            safe_data.setdefault("error_description", "")
```
rayluo: Did we actually observe those placeholder text in error code path during manual testing? The `error` shall have already be available in the error scenario, according to the specs. `error_description` is typically also available, but yes, you can use this treatment for `error_description` and `error_uri`.

msal/oauth2cli/authcode.py
```diff
@@ -317,19 +366,22 @@ def _get_auth_response(self, result, auth_uri=None, timeout=None, state=None,
                     auth_uri_callback(_uri)
 
         self._server.success_template = Template(success_template or
-            "Authentication completed. You can close this window now.")
+            "Authentication complete. You can return to the application. Please close this browser tab.\n\n"
+            "For your security: Do not share the contents of this page, the address bar, or take screenshots.")
         self._server.error_template = Template(error_template or
-            "Authentication failed. $error: $error_description. ($error_uri)")
+            "Authentication failed. $error: $error_description.\n\n"
```
rayluo: If you already add the `setdefault` treatment, whey do we still need to remove the `error_uri` from the error output?

rayluo (changes requested):
Thanks for the PR which communicated the intention clearly. The implementation needs some change.

MSAL Python code base serves for both public clients and confidential clients. The current architecture is:

1. oauth2.py handles all OAuth2 interactions which does NOT include hosting a web server to receive the auth code response. The web server part is controlled by whatever web framework the app developer chooses for their confidential application. We cannot forcefully override our response_mode to form_post, otherwise their existing query-based app would be broken (sev 1). What we can do (in oauth2.py) is to issue a warning when the response_mode value is not "form_post".
2. application.py defines the higher-level API that utilizes oauth2.py. So it will automatically trigger that warning in `#1`. The only thing needs to be done is to modify its documentation. (More on this below.)
3. authcode.py is the mini web server implementation that is used by Public Clients (when not using broker). We fully control how this web server is implemented, so, we can simply remove the "auth code via GET" mode, and refactor it to use form_post mode only.

With that context, now my following inline comments shall be easy to understand.

msal/oauth2cli/authcode.py
ashok672: This is simplified lot more with your previous comment about getting rid of code and error branch,

msal/oauth2cli/authcode.py
ashok672: We have decided to go ahead with deprecation and automative override. Applications should migrate immediately if at all they used query mode response_mode explicitly which is unlikely.

msal/oauth2cli/authcode.py
ashok672: yes I did see those placeholder text in error code path. Now with form_post the erroe detail is not returned and most of the times, the error is shown in eSTS UX.

msal/oauth2cli/authcode.py
ashok672: good catch. Fixed it.

msal/oauth2cli/authcode.py
ashok672: The experience was baffling to me initially until I understood what was going on. I think in this case, I will choose to keep the fix and not hang. Potentially state mismatches could happen in production as well (CSRF attacks and browser navigation issues)

msal/oauth2cli/authcode.py
```diff
@@ -370,8 +407,6 @@ def __exit__(self, exc_type, exc_val, exc_tb):
             )
         print(json.dumps(receiver.get_auth_response(
             auth_uri=flow["auth_uri"],
-            welcome_template=
```
ashok672, line 373: Clicking on abort now throws error about query mode not supported.
Also, Sign In or abort option is not consistent with other MSALs. Lets just do success or failure message.

msal/oauth2cli/authcode.py
```diff
@@ -370,8 +407,6 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
-                "<a href='$auth_uri'>Sign In</a>, or <a href='$abort_uri'>Abort</a",
             error_template="<html>Oh no. $error</html>",
```
ashok672, line 419: I  prefer more professional message that we had in the past.

msal/oauth2cli/authcode.py
ashok672: Let's be consistent with other public MSALs. None of the other MSALs show sign in or abort option.

msal/oauth2cli/authcode.py
ashok672: Reverted this change.

msal/oauth2cli/authcode.py
rayluo: Fixed in latest commit and discussed offline.

msal/oauth2cli/authcode.py
rayluo: This entire section is inside the `if __name__ == '__main__':` branch, it is just a built-in test. It will not affect the real code path. Besides, here we just fix a typo.

ashok672: That makes sense. I should have expanded the code little more to see the context.

rayluo (approved):
Worked closely with @ashok672 offline to agree on the latest changes. This is good to go!

Avery-Dunn (approved):

PR #876, opened 2026-02-20
Added withFmi method for cca app

4gust:
## Add `fmi_path` parameter to `acquire_token_for_client`

Adds Federated Managed Identity (FMI) support to `ConfidentialClientApplication.acquire_token_for_client()`, consistent with `WithFMIPath()` in MSAL Go and `.WithFmiPath()` in MSAL .NET.

### Usage

```python
import msal

app = msal.ConfidentialClientApplication(
    "your-client-id",
    client_credential={"private_key_pfx_path": "/path/to/cert.pfx", "public_certificate": True},
    authority="https://login.microsoftonline.com/your-tenant-id",
)

result = cca.acquire_token_for_client(
    scopes=["api://resource/.default"],
    fmi_path="SomeFmiPath/FmiCredentialPath",
)
```
### What is included 
New optional FMI path parameter on the client credential method — sends the path in the token request body and validates it is a string
Cache key extensibility — tokens acquired with different FMI paths are cached separately using a SHA-256 hash, with the "atext" credential type to match Go and .NET cache key format
Cache isolation — FMI-cached tokens do not interfere with regular client credential calls or with each other
Standard field exclusion — all standard OAuth2 body parameters are excluded from the cache key hash so only non-standard fields like the FMI path differentiate cache entries
Cross-MSAL compatibility — the hash algorithm produces identical output to MSAL Go and MSAL .NET, verified against their test vectors
### E2E test
Added E2E test to msid client

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -2491,6 +2491,32 @@ class ConfidentialClientApplication(ClientApplication):  # server-side web app
                 self.token_cache.remove_at(at)
         # acquire_token_for_client() obtains no RTs, so we have no RT to remove
 
+    def acquire_token_for_client_with_fmi_path(self, scopes, fmi_path, claims_challenge=None, **kwargs):
+        """Acquires token for the current confidential client with a Federated Managed Identity (FMI) path.
+
+        This is a convenience wrapper around :func:`~acquire_token_for_client`
+        that attaches the ``fmi_path`` parameter to the token request body.
+
+        :param list[str] scopes: (Required)
+            Scopes requested to access a protected API (a resource).
+        :param str fmi_path: (Required)
+            The Federated Managed Identity path to attach to the request.
+        :param claims_challenge:
+            The claims_challenge parameter requests specific claims requested by the resource provider
+            in the form of a claims_challenge directive in the www-authenticate header to be
+            returned from the UserInfo Endpoint and/or in the ID Token and/or Access Token.
+            It is a string of a JSON object which contains lists of claims being requested from these locations.
+
+        :return: A dict representing the json response from Microsoft Entra:
+
+            - A successful response would contain "access_token" key,
+            - an error response would contain "error" and usually "error_description".
+        """
+        data = kwargs.pop("data", {})
+        data["fmi_path"] = fmi_path
+        return self.acquire_token_for_client(
+            scopes, claims_challenge=claims_challenge, data=data, **kwargs)
+
     def acquire_token_on_behalf_of(self, user_assertion, scopes, claims_challenge=None, **kwargs):
         """Acquires token using on-behalf-of (OBO) flow.
 
--- a/msal/application.py
+++ b/msal/application.py
@@ -2427,7 +2427,7 @@ class ConfidentialClientApplication(ClientApplication):  # server-side web app
     except that ``allow_broker`` parameter shall remain ``None``.
     """
 
-    def acquire_token_for_client(self, scopes, claims_challenge=None, **kwargs):
+    def acquire_token_for_client(self, scopes, claims_challenge=None, fmi_path=None, **kwargs):
         """Acquires token for the current confidential client, not for an end user.
 
         Since MSAL Python 1.23, it will automatically look for token from cache,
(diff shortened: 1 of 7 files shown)


msal/application.py
```diff
@@ -2491,6 +2491,32 @@ def remove_tokens_for_client(self):
...
+            returned from the UserInfo Endpoint and/or in the ID Token and/or Access Token.
+            It is a string of a JSON object which contains lists of claims being requested from these locations.
+
+        :return: A dict representing the json response from Microsoft Entra:
+
+            - A successful response would contain "access_token" key,
+            - an error response would contain "error" and usually "error_description".
+        """
+        data = kwargs.pop("data", {})
+        data["fmi_path"] = fmi_path
```
bgavrilMS: Are the tokens cached like in MSAL .NET and MSAL GO? I don't think they are. 

I see a PR here that look at proper caching https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/759 but I am not sure it is compliant with the rest.

msal/application.py
```diff
@@ -2491,6 +2491,32 @@ def remove_tokens_for_client(self):
                 self.token_cache.remove_at(at)
         # acquire_token_for_client() obtains no RTs, so we have no RT to remove
 
+    def acquire_token_for_client_with_fmi_path(self, scopes, fmi_path, claims_challenge=None, **kwargs):
```
bgavrilMS: Should we have FMI path as one of the kwargs instead? Creating an entire new API seems pretty heavy.

msal/application.py
4gust: Added a cache enhancement so now we create a hash key for some data keyvalues

msal/token_cache.py
```diff
@@ -12,6 +14,63 @@
...
+#   - "claims"      : Handled separately; its presence forces a token refresh
+#   - "scope"       : Already represented as "target" in the AT cache key;
+#                     also added to data only at wire-time, not at cache-lookup time
+#   - "username"    : Standard ROPC grant parameter, not an extra body parameter
+#   - "password"    : Standard ROPC grant parameter, not an extra body parameter
+#
+# Included fields (examples — anything NOT in this set is included):
+#   - "fmi_path"    : Federated Managed Identity credential path
+#   - any future extra body parameter that should isolate cache entries
+_EXT_CACHE_KEY_EXCLUDED_FIELDS = frozenset({
```
bgavrilMS, line 45: Pls add a comment that these fields are already taken care of separately.

msal/application.py
bgavrilMS: This comment still stands. Should we "an overload" of AcquireTokenForClient existing API?

msal/token_cache.py
```diff
@@ -12,6 +14,63 @@
...
+    "key_id",
+    "token_type",
+    "req_cnf",
+    "claims",
```
bgavrilMS: I think you are missing a few things: 

- client_assertion and client_assertion_type and client_secret
- grant_type
- force_refresh

We probably need to enhance existing tests to assert that no "unknown" key is used.

msal/token_cache.py
```diff
@@ -12,6 +14,63 @@
 logger = logging.getLogger(__name__)
 _GRANT_TYPE_BROKER = "broker"
 
+# Fields in the request data dict that should NOT be included in the extended
```
bgavrilMS, line 17: The .NET spec mentioned using "atext" in the cache key. Maybe we are missing a test that attempts to load a token from a file.

msal/token_cache.py
```diff
@@ -12,6 +14,89 @@
...
+# cache key hash. Everything else in data IS included, because those are extra
+# body parameters going on the wire and must differentiate cached tokens.
+#
+# Excluded fields and reasons:
+#   - "client_id"              : Standard OAuth2 client identifier, same for every request
```
bgavrilMS:
```suggestion
#   - "client_id"              : Standard OAuth2 client identifier, already included in cache key
```

msal/token_cache.py
```diff
@@ -12,6 +14,89 @@
...
+#   - "grant_type"             : Standard OAuth2 grant type (e.g. jwt-bearer, refresh_token)
```
bgavrilMS:
```suggestion
#   - "grant_type"             : It is possible to combine grants to get tokens, e.g. obo + refresh_token, auth_code + refresh_token etc.
```

msal/token_cache.py
```diff
@@ -12,6 +14,89 @@
...
+#   - "scope"                  : Already represented as "target" in the AT cache key
+#   - "claims"                 : Handled separately; its presence forces a token refresh
+#   - "username"               : Standard ROPC grant parameter
```
bgavrilMS:
```suggestion
#   - "username"               : Standard ROPC grant parameter. Tokens are cached by user ID (subject or oid+tid) instead
```

msal/token_cache.py
```diff
@@ -12,6 +14,89 @@
...
+#   - "password"               : Standard ROPC grant parameter
```
bgavrilMS:
```suggestion
#   - "password"               : Standard ROPC grant parameter. Tokens are tied to credentials.
```

msal/application.py
4gust: updated the original api

msal/token_cache.py
4gust: updated

msal/token_cache.py
4gust: Updated

Avery-Dunn (approved):

bgavrilMS (approved):

gladjohn (approved):

PR #878, opened 2026-02-23
Removed logs that were causing Alerts

4gust:
Removed logs from oidc
Updated the version for msal python sku for release

### DIFF
--- a/msal/authority.py
+++ b/msal/authority.py
@@ -3,10 +3,6 @@ try:
     from urllib.parse import urlparse
 except ImportError:  # Fall back to Python 2
     from urlparse import urlparse
-import logging
-
-
-logger = logging.getLogger(__name__)
 
 # Endpoints were copied from here
 # https://docs.microsoft.com/en-us/azure/active-directory/develop/authentication-national-cloud#azure-ad-authentication-endpoints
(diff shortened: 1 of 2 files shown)


msal/sku.py
```diff
@@ -2,5 +2,5 @@
 """
 
 # The __init__.py will import this. Not the other way around.
-__version__ = "1.34.0"
+__version__ = "1.35.0"
```
bgavrilMS, line 5: Why this change? Did we forget to update the version number last time?

bgavrilMS (approved):

msal/sku.py
4gust: It was not release, it was a pre-release

https://pypi.org/project/msal/#history

msal/authority.py
```diff
@@ -92,11 +88,9 @@ def __init__(
         self._http_client = http_client
         self._oidc_authority_url = oidc_authority_url
         if oidc_authority_url:
-            logger.debug("Initializing with OIDC authority: %s", oidc_authority_url)
```
Avery-Dunn, line 95: I think the flag was only on logging "openid_config" since it was a raw HTTP response.

But why remove these authority URL logs? Seems like useful info to help identify config issues.

Avery-Dunn (approved):

gladjohn (approved):

msal/authority.py
4gust: I can add it back, do we need this ?

Avery-Dunn: Yeah, I think it's good to keep. I've definitely had trouble helping customers in the past because the logs didn't say where token requests were actually going.

4gust: added it back

bgavrilMS (approved):

Issue #891, opened 2026-03-24
[Bug] Do not rely on "endswith domain" for issuer validation


bgavrilMS:
https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/dev/msal/authority.py#L155 shows a pattern of validating the OIDC issuer (a defense in depth initiative) which uses `endswith(domain)`. This is incorrect, as a  fake domain can be "fakeb2clogin.com" and would pass the filter. 

The solution is to replace with a stronger `endswith("."+domain)`, e.g. `endswith(".b2clogin.com")`

Please audit the other MSALs as part of this.

gladjohn:
> Please audit the other MSALs as part of this.

Other MSALs have been audited and looks good

PR #896, opened 2026-04-07
Fix OIDC issuer domain spoofing in B2C host validation

4gust:
## Problem
The OIDC issuer validation in `has_valid_issuer()` used `endswith(domain)` to check B2C hosts, which allows malicious domains to pass validation #891:

```python
# Before: fakeb2clogin.com passes because it endswith("b2clogin.com")
issuer_host.endswith(h)  # h = "b2clogin.com"
```
### Fix
Require a dot separator to ensure only true subdomains match:
```python
issuer_host.endswith("." + h)  # Only matches *.b2clogin.com
```
### Tests
Added 5 tests covering spoofing scenarios (fakeb2clogin.com, evilb2clogin.com, b2clogin.com.evil.com, fakeciamlogin.com) and valid subdomain acceptance.

### DIFF
--- a/msal/authority.py
+++ b/msal/authority.py
@@ -250,7 +250,9 @@ class Authority(object):
             authority_parsed.netloc == issuer_parsed.netloc):
             return True
         
-        # Case 5: Check if issuer host ends with any well-known B2C host (e.g., tenant.b2clogin.com)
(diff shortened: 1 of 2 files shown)


gladjohn (approved):
🚢

PR #909, opened 2026-05-08
Use tenant-specific authority instead of organizations in certain tests

Avery-Dunn:
Two OBO integration tests (`test_acquire_token_obo` and 
`test_cca_obo_should_bypass_regional_endpoint_therefore_still_work`) intermittently fail with: ``AADSTS53003: Access has been blocked by Conditional Access policies. The access policy does not allow token issuance.`

These were the only automated ROPC tests using the multi-tenant `/organizations` authority. With `/organizations`, AAD must perform home-realm discovery to route the request to the correct tenant, and this additional routing context can cause risk-based Conditional Access policies to intermittently block the sign-in — particularly in CI environments with shared IPs and high sign-in frequency.

This PR switches both tests to use the tenant-specific `user.authority`, matching the pattern used by every other passing ROPC test. The OBO flow itself is unaffected — only the authority used to obtain the initial user assertion changes.

### DIFF
--- a/tests/test_e2e.py
+++ b/tests/test_e2e.py
@@ -910,9 +910,8 @@ class WorldWideTestCase(LabBasedTestCase):
         web_api_app = get_app_config(AppSecrets.WEB_API_CLIENT)
 
         # Step 1: PCA gets token for user to access the WebAPI
-        # Note: Java test uses "organizations" authority for PCA
         config_pca = {
-            "authority": "https://login.microsoftonline.com/organizations",
+            "authority": user.authority,
             "client_id": web_api_app.app_id,
             "username": user.upn,
             "password": password,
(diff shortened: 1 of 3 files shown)


4gust (approved):

bgavrilMS (approved):

tests/test_e2e.py
```diff
@@ -902,31 +902,28 @@ def test_acquire_token_obo(self):
...
         
         Flow:
-        1. PCA acquires token for user to access the WebAPI (scope: api://<app_id>/access_as_user)
-        2. WebAPI (CCA) uses that token as assertion to get token for downstream service (Graph)
+        1. S2S app (PCA) acquires token for user, targeting the WebAPI
+        2. WebAPI (CCA) uses that token as assertion to get token for downstream (Graph)
         """
         user = get_user_config(UserSecrets.PUBLIC_CLOUD)
         password = get_user_password(user)
+        s2s_app = get_app_config(AppSecrets.S2S_CLIENT)
```
4gust, line 910: Could we add a short comment explaining why s2s_app is used here instead of web_api_app? Something like "S2S client supports ROPC with tenant-specific authority; web_api_app is a resource app" would save the next developer a round-trip to Key Vault to figure out the relationship.

4gust (approved):
Small comment

4gust:
@Avery-Dunn  Do we still needs these changes to test?

PR #933, opened 2026-06-21
docs: Guidance for higher-level SDKs to consume MSAL mTLS PoP

gladjohn:
## Summary

Adds a single documentation file explaining how higher-level SDKs (e.g., azure-identity, azure-sdk-for-python) can integrate with the MSI v2 mTLS Proof-of-Possession API.

**This PR contains only documentation — no code changes.**

## Contents

- Public API surface and return value contract
- `WindowsCertificate` object properties and methods (accurate to implementation)
- Step-by-step integration pattern (credential → transport)
- `SchannelSession` usage (certificate in constructor, not per-request)
- No-fallback behavior (matches MSAL .NET `MtlsPopTokenNotSupportedinImdsV1`)
- End-user experience goal (zero mTLS awareness)
- .NET comparison table
- Minimum integration example
- Future OpenSSL 3 CNG Provider path

## Related

- PR #931 (MSI v2 mTLS PoP implementation)
- MSAL .NET PR #6049 (`GetManagedIdentityCapabilitiesAsync`)

### DIFF
--- a/docs/Guidance-for-Higher-Level-SDKs-to-Consume-MSAL.md
+++ b/docs/Guidance-for-Higher-Level-SDKs-to-Consume-MSAL.md
@@ -1,0 +1,379 @@
+# Guidance for Higher-Level SDKs to Consume MSAL mTLS PoP
+
+## Purpose
+
+This document explains how higher-level SDKs (e.g., `azure-identity`, `azure-sdk-for-python`)
+can consume MSAL Python's MSI v2 mTLS Proof-of-Possession API to provide seamless mTLS
+token-bound authentication to end users.
+
+## What MSAL Python Exposes
+
+### Public API Surface
+
+```python
+import msal
+import requests
+
+# 1. Create ManagedIdentityClient (existing API)
+client = msal.ManagedIdentityClient(
+    msal.SystemAssignedManagedIdentity(),
+    http_client=requests.Session(),
+)
+
+# 2. Acquire mTLS-bound token (new parameters)
+result = client.acquire_token_for_client(
+    resource="https://vault.azure.net",
+    mtls_proof_of_possession=True,       # Always routes to MSI v2 (hard fail if unsupported)
+    with_attestation_support=True,       # KeyGuard tier (attested); False = Software tier
+)
+```
+
+### Return Value Contract
+
+```python
+{
+    "access_token": "eyJ0eXAi...",                    # mTLS-bound PoP token
+    "token_type": "mtls_pop",                         # Token type (use in auth header)
+    "expires_in": 86399,                              # Seconds until expiry
+    "binding_certificate": WindowsCertificate(...),   # Opaque cert+key handle
+    "cert_thumbprint_sha256": "buc7x...",             # Base64url SHA-256 thumbprint
+    "cert_pem": "-----BEGIN CERTIFICATE-----\n...",   # Public cert (PEM)
+    "cert_der_b64": "MIIC...",                        # Public cert (Base64 DER)
+}
+```
+
+### Key Objects
+
+#### `WindowsCertificate`
+
+Python equivalent of .NET's `X509Certificate2`. Wraps a non-exportable private key
+(NCRYPT_KEY_HANDLE) and the associated X.509 certificate.
+
+```python
+from msal import WindowsCertificate
+
+cert: WindowsCertificate = result["binding_certificate"]
+
+# Properties
+cert.thumbprint_sha256       # str: hex SHA-256 thumbprint (uppercase, 64 chars)
+cert.x5t_s256               # str: base64url SHA-256 (matches cnf.x5t#S256 in JWT)
+cert.public_certificate_pem  # str: PEM-encoded public certificate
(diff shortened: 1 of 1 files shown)


g2vinay:
From Azure SDK POV:

1. It won't be a clean/smooth integration into python azure-core via the SChannel transport (involves native bindings), as it will involve building an adapter around it for windows only AKV scenario here. Whether the current python http stack works with SChannel Transport or not, is something to be confirmed by the Identity Python SDK owner.

2. The distribution of native Attestation Lib (5.3 MB), will add additional bloat for 3P.

3. Majority Python customers are linux focused. Windows is minority. There won't be major demand for the windows flow.

docs/Guidance-for-Higher-Level-SDKs-to-Consume-MSAL.md
```diff
@@ -0,0 +1,391 @@
...
+| `msal` | Token acquisition + `WindowsCertificate` | Yes |
+| `msal-key-attestation` | MAA attestation (KeyGuard proof) | Yes for `with_attestation_support=True` |
+| `msal-schannel-transport` | WinHTTP-based downstream mTLS | Yes (or build your own) |
+| `requests` | HTTP client for MSAL's IMDS calls | Yes |
+
+---
+
+## Future: OpenSSL 3 CNG Provider (Strategic Path)
+
+When a Microsoft-supported OpenSSL 3 CNG Provider becomes available, the
```
kashifkhan, line 379: I want to flag that this "strategic path," as written, isn't reachable today  and likely not for a long while. I checked with the Python core devs  on whether hardware-key mTLS would become reachable from the stdlib `ssl`  module /  `SSLContext` , and the answer was "not coming any time soon."


As y'all know, `ssl.SSLContext` has no way to reference a provider/PKCS#11 key, `load_cert_chain`  only accepts key material as files, and there's no signing-callback or provider-key hook. So even if a Microsoft-supported OpenSSL 3 CNG provider existed, `requests/aiohttp`  still couldn't use it, because both sit on stdlib `ssl`, which can't point at a provider-backed key by thumbprint. 

The snippet in this section ( `create_mtls_context(thumbprint=...`)  → plain  requests.get(...) ) would require exactly the CPython feature that doesn't exist.

The proper channel for this is a new TLS module (the PEP 543 /  tlslib  effort), and its native-backend work (incl. SChannel) is currently stalled for lack of contributors https://discuss.python.org/t/pre-pep-discussion-revival-of-pep-543-a-unified-tls-api-for-python/51263

They are looking for people who can drive this however :)

docs/Guidance-for-Higher-Level-SDKs-to-Consume-MSAL.md
```diff
@@ -0,0 +1,391 @@
...
+response = session.get(url, headers=headers)
+response = session.post(url, headers=headers, body=data)
+```
+
+For azure-core integration, build a custom `HttpTransport`:
+
+```python
+from azure.core.pipeline.transport import HttpTransport
+
+class SchannelTransport(HttpTransport):
```
kashifkhan, line 206: This particular Schannel Transport will basically have to mirror the functionality provided by requests and aiohttp ( for async ) . You will have to create a full WinHTTP based HTTP Client making sure we get the following features out from it :

* request/response serialization and a conformant  azure.core.rest.HttpResponse  (streaming/ iter_bytes / read /encoding) -> for folks who are using storage, cosmos, etc etc that have different needs from a http client
* connection pooling / keep-alive (the sketch opens a fresh mTLS handshake per request)
* redirects, proxies, timeouts, retries, TLS config, and error→exception mapping — all of which WinHTTP does differently than requests/aiohttp, so they'd have to be reconciled to keep behavior consistent across platforms
* a separate async transport entirely, since  aiohttp  can't present the key and WinHTTP here is sync

 A few thoughts on differences - 
* does the transport need to worry about differences between Windows versions for eg 
* does winhttp support reading certs in the same manner as requests/aiohttp ?
* Will standard proxy env vars ( HTTP_PROXY / HTTPS_PROXY / NO_PROXY ) be honored
* Diagnostics diverge too:  `urllib3 / aiohttp`  debug logging (and standard support steps) won't work on WinHTTP — its tracing lives in Windows ETW — and policy-level logs may not reflect WinHTTP's internal redirects/headers.

Just this itself greatly increases the effort and maintenance cost and the security concerns around this too

docs/Guidance-for-Higher-Level-SDKs-to-Consume-MSAL.md
```diff
@@ -0,0 +1,391 @@
...
+
+```python
+from azure.identity import ManagedIdentityCredential
+from azure.keyvault.secrets import SecretClient
+
+# One line to enable mTLS PoP
+credential = ManagedIdentityCredential(mtls_pop=True)
+
+# Standard SDK usage — no mTLS awareness needed by the developer
+client = SecretClient(
```
kashifkhan, line 236: credential and transport are separate in azure-core. So doesn't SecretClient still need updating to hand the credential to a WinHTTP transport and use it

PR #951, opened 2026-08-07
Harden regional authority validation

4gust:
## Summary
- Fail closed for unknown authority and OIDC validation failures when regional mode is enabled.
- Restrict regional resilience to transport failures from well-known Microsoft authority hosts.
- Enforce safe DNS-label validation for explicit and automatically detected regions.

## Tests
- python -m unittest tests.test_application.TestAuthorityValidationWithRegionalMode tests.test_region
- git diff --check

The lab-backed E2E selector was updated for strict raw region handling but could not run locally because its optional zure.identity dependency is not installed.

### DIFF
--- a/msal/application.py
+++ b/msal/application.py
@@ -14,6 +14,7 @@ from .oauth2cli.oidc import decode_part
 from .authority import (
     Authority,
     WORLD_WIDE,
+    WELL_KNOWN_AUTHORITY_HOSTS,
     _get_instance_discovery_endpoint,
     _get_instance_discovery_host,
 )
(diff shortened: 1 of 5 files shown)


bgavrilMS (approved):

gladjohn (approved):

msal/application.py
```diff
@@ -717,31 +735,32 @@ def get_client_assertion():
...
+                        azure_region
+                        or force_region is not None
+                    )):
                 raise
+            self.authority = Authority(
+                authority_to_use,
+                self.http_client,
+                validate_authority=True,
+                instance_discovery=False,
+                )
```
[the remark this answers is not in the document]
4gust: Thanks. `requests.exceptions.RequestException` directly inherits `OSError` (as do `ConnectionError` and `Timeout`), so this guard already handles Requests transport failures. I added a regression test that raises `requests.exceptions.ConnectionError` during known-authority discovery and verifies the regional retry succeeds (`a98631d`).

msal/region.py
```diff
@@ -25,7 +25,7 @@ def _validate_region(region, source="unknown"):
 
 
 def _detect_region(http_client=None):
-    region = os.environ.get("REGION_NAME", "").replace(" ", "").lower()  # e.g. westus2
+    region = os.environ.get("REGION_NAME", "")  # e.g. westus2
```
Avery-Dunn, line 28: I brought up a similar question in the Go changes (https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/646#discussion_r3751232422), but why are we removing the normalization?

I think MSAL Java has never normalized it, but I don't know for sure if that means it's not needed in any MSAL or if Java has a bug that simply hasn't been reported.

One the one hand it means MSAL Java has worked fine for years without normalization, but on the other hand surely there was some reason for .NET and most other MSALs were trying to normalize it.

Avery-Dunn: Resolved in a separate discussion, normalization should not be needed.

Avery-Dunn (approved):

